PR #5, opened 2016-10-14
Avoid code repetition in resumable_io::start and resumable_io::start_pending

AlexBAV:
Separate `resumable_io::start_pending` method is no longer required; `resumable_io::start` is now capable of working with both callbacks returning `void` and `bool`. Also avoids code repetition.

`start_pending` is left for compatibility though (implemented via a call to `resumable_io::start`).

### DIFF
--- a/10.0.14393.0/winrt/base.h
+++ b/10.0.14393.0/winrt/base.h
@@ -7395,104 +7395,15 @@ struct resumable_io
     }
 
     template <typename F>
-    auto start(F callback)
+    auto start(F &&callback)
     {
-        struct awaitable : awaitable_base, F
-        {
-            awaitable(PTP_IO io, F callback) noexcept :
-                m_io(io),
-                F(callback)
-            {}
-
-            bool await_ready() const noexcept
-            {
-                return false;
-            }
-
-            void await_suspend(std::experimental::coroutine_handle<> resume_handle)
-            {
-                m_resume = resume_handle;
-                StartThreadpoolIo(m_io);
-
-                try
-                {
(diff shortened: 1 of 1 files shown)


kennykerr:
We cannot directly accept pull requests at the moment since the headers are generated by the C++/WinRT compiler. Even the base.h header is an amalgamation produced by this compiler. Still, we’ll take a look at the pull request and see if we can include it in a future update. Thanks for the feedback!

kennykerr:
I had a look at your suggested changes. I like the idea of combining both methods into a single start method that can handle both scenarios. We will make this available in a future update. Thanks for the feedback!

AlexBAV:
Thanks, Kenny.

BTW, I wonder if it is legal at all to cancel a callback if Windows IO function completes synchronously. It appears that if the handle notification mode is not set to `FILE_SKIP_COMPLETION_PORT_ON_SUCCESS` (see <https://msdn.microsoft.com/en-us/library/windows/desktop/aa365538(v=vs.85).aspx>), it would be an error to do so.

It looks like the start_pending is only for those handles for which this notification mode is set. What do you think of it?

kennykerr:
Yes, the thread pool callback must be canceled if the I/O completes synchronously *and* the completion packet will not be queued. This is only the case if the FILE_SKIP_COMPLETION_PORT_ON_SUCCESS mode is set and why I originally had both start and start_pending. The implementation needs to handle this carefully to avoid leaks.

AlexBAV:
Thnk you for clarification.

PR #18, opened 2016-10-19
Use / for include paths, not \

LegalizeAdulthood:
I was an MVP for Direct3D for 10 years, you might already have a CLA on file for me...

### DIFF
--- a/10.0.14393.0/Samples/AsyncReader/pch.h
+++ b/10.0.14393.0/Samples/AsyncReader/pch.h
@@ -2,4 +2,4 @@
 
 #pragma comment(lib, "windowsapp")
 
-#include "winrt\base.h"
+#include "winrt/base.h"
--- a/10.0.14393.0/Samples/Blocks/pch.h
+++ b/10.0.14393.0/Samples/Blocks/pch.h
@@ -2,7 +2,7 @@
 
 #pragma comment(lib, "windowsapp") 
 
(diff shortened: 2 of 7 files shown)


raffaeler:
I believe it does not make sense to make pull requests on the libraries as it is all generated code.
The patch should be always applied to the generator.

BrentRe:
Actually this particular request is to update the Sample projects, which are not generated code.

raffaeler:
oops, sorry

LegalizeAdulthood:
Yes, as @BrentRe pointed out, the libraries are fine, it's just the sample code that is using \ instead of /.

They also use `#pragma lib`, which isn't portable C++, but noone else besides Windows has a WinRT runtime anyway, so I doubt that part really matters.

BrentRe:
We could move the #pragma lib into the VC++ project settings to remove it. But as all these samples are Windows-specific, I'm not sure there's much benefit. The reasoning for putting in the source code was to alert developers that Windows Store applications must link with that library. Something that's easily missed when it buried in a linker command line setting.

For the record, what are you suggesting to change from back slashes to forward slashes? I'm guessing because both work and forward is more platform agnostic... I'm not pushing back but these samples are very Windows specific (as is C++/WinRT currently).

kennykerr:
Thanks for the contribution Richard!

PR #20, opened 2016-10-21
Add nuget package spec and targets file to configure project correctly

tjp9889:
Nuget package points to latest version

### DIFF
--- a/nuget/microsoft.cppwinrt.nuspec
+++ b/nuget/microsoft.cppwinrt.nuspec
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>microsoft.cppwinrt</id>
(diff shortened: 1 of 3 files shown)


nuget/microsoft.cppwinrt.nuspec
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>microsoft.cppwinrt</id>
+    <version>0.0.0</version>
```
tjp9889: http://docs.nuget.org/ndocs/create-packages/prerelease-packages#semantic-versioning
Nothing in this delta is dependent on the version number you end up choosing though.

tjp9889: Probably what you want is more than one package (one per targeted SDK), whose version tracks the C++/WinRT compiler version

PR #86, opened 2017-01-21
More typo fixes for "Migrating C++ CX source code to C++ WinRT.md"

jonitis:
More typo fixes.

One thing that was not typo, but I was not sure whether it is right or not:

This C++/CX code:
  void LogWrapLine(Platform::String^ str)

converts to the standard C++ code:

  void LogWrapLine(std::wstring str)

Do you really recommend to pass c++ wstring by value and not as "const std::wstring& str"? Compiler will figure out to create hstring_ref, doesn't it?

### DIFF
--- a/Docs/Migrating C++ CX source code to C++ WinRT.md
+++ b/Docs/Migrating C++ CX source code to C++ WinRT.md
@@ -78,7 +78,7 @@ First let's look at how to get property values. Here's the C++/CX code:
 
     auto id = record->XboxUserId;
 
-    auto state = record->UserState,
+    auto state = record->UserState;
 
     auto size = record->PresenceDeviceRecords->Size;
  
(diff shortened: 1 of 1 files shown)


BrentRector:
Regarding:This C++/CX code:
void LogWrapLine(Platform::String^ str)
converts to the standard C++ code:
void LogWrapLine(std::wstring str)

Yes, passing the string as a const wstring & is better. However, this particular code has nothing to do with WinRT, though the snippet didn't emphasize it. LogWrapLine is simply a standard C++ method. So the original code really shouldn't have used a Platform::String in the first place. Which is why I changed it to use wstring, not hstring.

As an aside, as guidance for using C++/WinRT, all projected WinRT APIs use hstring_ref as the string parameter type. Generally, you use 'hstring' should be used for WinRT string variables or fields. 'hstring_ref' should be used for WinRT API string parameters.

PR #87, opened 2017-01-21
Change array_ref m_size from uint32_t to size_type

jonitis:
Just cosmetic change, since array_ref defines size_type.

Other places that also may be changed are:
- make_array_iterator(T * data, uint32_t size, uint32_t index = 0)
- com_array impl_put_size(com_array & value, const uint32_t size) an impl_detach() result pair
- array_size_proxy m_size and operator uint32_t * () noexcept
- com_array_proxy constructor and m_size

### DIFF
--- a/10.0.14393.0/winrt/base.h
+++ b/10.0.14393.0/winrt/base.h
@@ -1997,13 +1997,13 @@ struct array_ref
 
 protected:
 
-    array_ref(pointer data, uint32_t size) :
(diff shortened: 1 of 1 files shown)


kennykerr:
We don't yet have the ability to accept pull requests for code changes since the bulk of the code is generated by the C++/WinRT compiler. Thanks for the suggestions! We're certainly taking note of them.

Issue #124, opened 2017-03-03
cmake interoperability


regnirpsj:
is there a (minimal) example using cppwinrt in a cmake-based project?

bigdimboom:
Request cmake for regular winrt projects and for UWP projects.

johanlindfors:
CMake already supports UWP projects, and it can easily be configured for generating cppwinrt based solutions as well, the only thing I'm struggling with currently is how to disable /ZW in that project generation.

johanlindfors:
Fixed a simple solution that leverages CMake to generate a project using cppwinrt. Please comment!

kennykerr:
I don't know much about cmake so I'll let others comment on that, but here are the steps you need to convert a C++/CX store project into a standard C++ project.

1. Remove /ZW
2. Remove C++/CX code
3. Add standard wWinMain function
4. Add empty app.winmd file for app package

It's that last step that most people forget. The C++/CX compiler automatically creates one and if you simply convert the project it will work for a while, until you happen to delete your output folder and rebuild.

FeBrex:
Instead of adding an empty app.winmd one can also add a minimal idl and configure the MIDL compiler to generate the winmd

1. add <projectName>.idl, insert a namespace definition
2. Check that /winrt MIDL compiler switch is active
3. Check value of /winmd filename MIDL compiler switch

Zingam:
@FeBrex @BrentRe 

Can we have an official example of this setup?

Sunbreak:
I'm developing Flutter for Windows. Hope the snippet help

https://github.com/woodemi/quick_blue/pull/10/files

```cmake
set_target_properties(${PLUGIN_NAME} PROPERTIES
  VS_PROJECT_IMPORT ${CMAKE_BINARY_DIR}/packages/Microsoft.Windows.CppWinRT/build/native/Microsoft.Windows.CppWinRT.props
)

target_link_libraries(${PLUGIN_NAME} PRIVATE ${CMAKE_BINARY_DIR}/packages/Microsoft.Windows.CppWinRT/build/native/Microsoft.Windows.CppWinRT.targets)
```

dirkarnez:
> I'm developing Flutter for Windows. Hope the snippet help
> 
> https://github.com/woodemi/quick_blue/pull/10/files
> 
> ```cmake
> set_target_properties(${PLUGIN_NAME} PROPERTIES
[...]

PR #126, opened 2017-03-06
Leverage CMake to generate project using cppwinrt

johanlindfors:
This sample was requested in #124 

It demonstrates a simple setup in which CMake is leveraged to generate the project structure for the IBuffer sample.

Change directory to the included output-directory and use the generate.bat script to generate solution.

Works fine on a Surface Pro 4 with Visual Studio 2017 RC and CMake 3.8.0-rc2


johanlindfors:
Updated this PR to work with Creators Update.

MathiasMagnus:
@johanlindfors @kennykerr This is a nearly 3 year old story, but why exactly was this not taken in? Honestly I'm amazed at how little CMake support core Windows-related technologies have. Large projects that would benefit the most from being able to compile on Windows with any compiler are generally committed to some build system, usually other than MSBuild.

_(Longwinded, destructive ranting on frustrations on using cross-platform tooling for Windows platform development swallowed.)_

Sunbreak:
@MathiasMagnus 

I'm developing Flutter for Windows. Hope the snippet help

https://github.com/woodemi/quick_blue/pull/10/files

```cmake
set_target_properties(${PLUGIN_NAME} PROPERTIES
  VS_PROJECT_IMPORT ${CMAKE_BINARY_DIR}/packages/Microsoft.Windows.CppWinRT/build/native/Microsoft.Windows.CppWinRT.props
)

target_link_libraries(${PLUGIN_NAME} PRIVATE ${CMAKE_BINARY_DIR}/packages/Microsoft.Windows.CppWinRT/build/native/Microsoft.Windows.CppWinRT.targets)
```

LBJ-the-GOAT:
> @MathiasMagnus
> 
> I'm developing Flutter for Windows. Hope the snippet help
> 
> https://github.com/woodemi/quick_blue/pull/10/files
> 
> ```cmake
> set_target_properties(${PLUGIN_NAME} PROPERTIES
>   VS_PROJECT_IMPORT ${CMAKE_BINARY_DIR}/packages/Microsoft.Windows.CppWinRT/build/native/Microsoft.Windows.CppWinRT.props
> )
> 
> target_link_libraries(${PLUGIN_NAME} PRIVATE ${CMAKE_BINARY_DIR}/packages/Microsoft.Windows.CppWinRT/build/native/Microsoft.Windows.CppWinRT.targets)
> ```

thinks man, now my project is running smoothly and very nicely because of your greate TIPS! THANK YOU!

walbourn:
The [vcpkg](https://vcpkg.io/) manager has excellent support for CMake, so having a vcpkg port for cppwinrt might be a good solution here.

sylveon:
There is a cppwinrt port, but the port is a no-op, so installing it and trying to use `<winrt/base.h>` makes the compiler use the Windows SDK's copy, so we just circle back to the initial problem: https://github.com/microsoft/vcpkg/blob/master/ports/cppwinrt/portfile.cmake

Ideally cppwinrt itself would have CMake stuff that allows the user to invoke the tool to generate consuming headers from the SDK and WinMD files, create WinMD files from IDL, generate source to produce a runtime component, etc. And then we'd just have the vcpkg port fetch the cppwinrt.exe tool and those CMake tasks, and CMake/vcpkg users could use those tasks in their project.

Unfortunately, just fetching the tool, generating the headers and embedding them in the vcpkg port wouldn't work too good, as not only are you limited to consuming system APIs only, but the version of the Windows SDK used by the port might not match the one used by the project (and there is nothing that can be done to adjust this from the project's side).

But it would need somebody with enough CMake knowledge to be able to invest time and resources into developing this (and my knowledge of CMake is certainly not good enough for such a task)

walbourn:
@sylveon That's a pretty lame port, but it's better than nothing I suppose. I've authored a few ports that fetch NuGets and could be a model for the basics of how to set this up you could try starting with...

https://github.com/microsoft/vcpkg/blob/master/ports/dxsdk-d3dx

https://github.com/microsoft/vcpkg/tree/master/ports/xaudio2redist

I need to revisit the ``find_package`` implementation soon for those. If it's not addressed by then, I can take a stab at what a cppwinrt vcpkg port might look like...

PR #216, opened 2017-07-08
NDEBUG fix

leeter:
Fix `NDEBUG` randomly breaking parts of the library because `assert` goes to `((void)0)` when the macro is defined.

### DIFF
--- a/10.0.14393.0/winrt/base.h
+++ b/10.0.14393.0/winrt/base.h
@@ -55,7 +55,7 @@ extern "C"
 #pragma comment(linker, "/alternatename:WINRT_SetRestrictedErrorInfo=SetRestrictedErrorInfo")
 #endif
 
-#ifdef _DEBUG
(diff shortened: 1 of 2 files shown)


leeter:
Note: I was thinking about this fix overnight and it may not be appropriate. `_DEBUG` is reserved for the implementation (it has a `_` prefix) and should probably not be used at all. The more appropriate fix may be switching that `#ifdef` to use `NDEBUG` instead.

Scottj1s:
@leeter, unfortunately, this fix would disable all WINRT_DEBUG facilities if NDEBUG is defined, which should really just target assert behavior.  As it happens, we fixed this issue in the latest update, which implements WINRT_VERIFY in terms of _ASSERTE, and not assert, so NDEBUG has no effect.  You can update to pick up this change, or just apply the line item fix to your older repo.

PR #268, opened 2017-11-06
CMake support

janosvitok:
This file helps to include C++/WinRT library in a CMake project.
For documentation see comments in the file.


kennykerr:
Thanks for the suggestion! We'll keep it in mind, however we are not in a position to take source code submissions just yet.

PR #275, opened 2017-11-15
Fixed compile error in event_revoker::operator bool()

Urmeli0815:
This code causes the compile error "...\cppwinrt\winrt\base.h(3858): error C2678: binary '!=': no operator found which takes a left-hand operand of type 'const winrt::weak_ref<I>' (or there is no acceptable conversion):

```
winrt::event_revoker<IAppServiceConnection> r;
if (r)
{
// ...
}
```


kennykerr:
Thanks for reporting this! I'll get this fixed in our internal repo. We cannot accept the pull request as it points to generated code.

PR #372, opened 2019-10-10
Add low-level coroutine suspension notifications

kennykerr:
The OS uses a form of activity tracking to log events that might complete asynchronously. When this is mixed with coroutines, the activity is tripped up because it does not realize that a suspension point may have switched the thread on which it is executing. This update provides a very low-level hook for use by WIL to notify the system that a coroutine is being suspended or resumed. As with [the winrt_to_hresult_handler hook](https://github.com/microsoft/xlang/pull/109), this is intended for low-level interop with WIL and I don't recommend developers using this as a general purpose notification mechanism.

The "get_awaiter" implementation here is pretty convoluted, but fortunately C++23 will have more library support for coroutines including a "get_awaiter" like function that will be part of the standard library.


oldnewthing:
Don't we need to add an `await_transform` to `fire_and_forget` too?

kennykerr:
@oldnewthing right, I'll add that.

strings/base_coroutine_foundation.h
```diff
@@ -260,6 +260,154 @@ namespace winrt::impl
...
+    }
+
+    template <typename T, std::enable_if_t<has_awaitable_free<T>::value, int> = 0>
+    auto get_awaiter(T&& value) noexcept
+    {
+        return free_await_adapter<T>{ static_cast<T&&>(value) };
+    }
+
+    template <typename T>
+    struct notify_awaiter
```
oldnewthing: Consider moving this out of `impl` so it can be used by user-defined promise types. (Like my `simple_task<T>`.)

kennykerr: This is a pretty invasive hook so I'll give it some time to bake, make sure it's what we really want, and then make a helper available outside the `impl` namespace.

PR #373, opened 2019-10-10
Fix build

wjk:
While trying to build the VSIX project to prepare another PR, I found and fixed approximately one metric crap-ton of build issues. It works for me now; I also added a quality-of-life batch file that builds the NuGet package. Enjoy!


kennykerr:
Thanks! We hadn't got around to getting that building after moving it over from xlang.

nuget/Microsoft.Windows.CppWinRT.nuspec
```diff
@@ -13,7 +13,6 @@
     <copyright>© Microsoft Corporation. All rights reserved.</copyright>
     <license type="file">LICENSE</license>
     <projectUrl>https://github.com/Microsoft/xlang/tree/master/src/package/cppwinrt/nuget</projectUrl>
-    <iconUrl>https://aka.ms/cppwinrt.ico</iconUrl>
```
Scottj1s: why was this removed?

vsix/vsix.csproj
```diff
@@ -65,8 +65,9 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Content Include="LICENSE">
+    <Content Include="..\LICENSE">
```
Scottj1s: this doesn't work (when tested with resulting vsix).  a limitation of vsix creation - the LICENSE file must be copied from parent (where it's shared by Microsoft.Windows.CppWinRT.nuspec) and .gitignored to not be checked in.

vsix/vsix.csproj
```diff
@@ -153,12 +154,11 @@
   </Target>
   <Import Project=".\packages\Microsoft.VSSDK.BuildTools.16.0.2258\build\Microsoft.VSSDK.BuildTools.targets" Condition="Exists('.\packages\Microsoft.VSSDK.BuildTools.16.0.2258\build\Microsoft.VSSDK.BuildTools.targets')" />
   <Target Name="GetCppWinRTVersion" Outputs="$(CppWinRTVersion)" />
-  <Target Name="PrepareBuild" BeforeTargets="Build">
+  <Target Name="PrepareBuild" BeforeTargets="PrepareForBuild">
     <Error Condition="'$(CppWinRTVersion)' == ''" Text="The project must be supplied with a CppWinRTVersion property value" />
     <Error Condition="'$(NatvisDirx86)' == ''" Text="The project must be supplied with a NatvisDirx86 property value" />
     <Error Condition="'$(NatvisDirx64)' == ''" Text="The project must be supplied with a NatvisDirx64 property value" />
     <Error Condition="'$(NupkgDir)' == ''" Text="The project must be supplied with a NupkgDir property value" />
-    <Copy SourceFiles="..\LICENSE" DestinationFolder="$(MSBuildProjectDirectory)" />
```
Scottj1s: this needs to stay - see above

nuget/build.cmd
```diff
@@ -0,0 +1,25 @@
+@echo off
```
Scottj1s: @kennykerr - we can take this, but it will probably need to be modified to support build pipeline

kennykerr: I'd rather just get the build up and running in one step. Might be simpler to close this PR (without merging) and just fix it all together (along with the internal build).

nuget/Microsoft.Windows.CppWinRT.nuspec
wjk: The `iconUrl` tag is [deprecated](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5048), and produces a warning during `nuget pack`. I attempted to download this URL and include the icon the the package using the non-deprecated `icon` tag, but the URL given 404ed when I tried that.

Scottj1s: FYI - I've updated the URL to the new repo

Scottj1s: Since iconUrl is deprecated, can you replace with an icon element pointing to vsix/cppwinrt.ico, and add that file to the nuspec?

Scottj1s: Since iconUrl is deprecated, can you replace with an icon element pointing to vsix/cppwinrt.ico, and add that file to the nuspec?

Scottj1s:
Thanks William!

PR #374, opened 2019-10-11
Race condition in resume_foreground(DispatcherQueue)

### LINKED ISSUE #370: Race condition in resume_foreground(DispatcherQueue)
```cpp
    [[nodiscard]] inline auto resume_foreground(
        Windows::System::DispatcherQueue const& dispatcher,
        Windows::System::DispatcherQueuePriority const priority = Windows::System::DispatcherQueuePriority::Normal) noexcept
    {
        struct awaitable
        {
...
            bool await_suspend(std::experimental::coroutine_handle<> handle)
            {
                m_queued = m_dispatcher.TryEnqueue(m_priority, [handle]
                    {
                        handle();
                    });

                return m_queued;
            }
```

There is a race condition here: If `TryEnqueue` succeeds, the coroutine may resume *before* `TryEnqueue` returns. In that case, the coroutine has been destructed, and our attempt to use `m_queued` will AV.

Unit tests randomly fail on my machine due to this bug.

oldnewthing:
It's possible for the lambda to run before control returns to the caller of TryEnqueue. This led to async_suspend accessing m_queued after it has already been destroyed.

We need to prevent the lambda from invoking the handler until async_suspend has stored the result of TryEnqueue into m_queued.

We do this by making the lambda wait until async_suspend is ready before invoking the handler. The slim_event cannot be a local in await_suspend because the lambda may run after await_suspend has returned. We put it in the awaiter. (Another option is putting it in the lambda, but that was messier.)

I chose a slim_event rather than a full kernel event because allocating a kernel event for every co_await seemed too expensive, especially because the wait is almost never going to block.

This commit introduces winrt::slim_event which is an event that starts out unsignaled but can be signaled. Once signaled, it cannot be reset. It is the size of a bool. For simplicity, only infinite waits are permitted. (Otherwise, I would have to recalculate the timeout when a spurious wake occurs.)

This fixes #370

### DIFF
--- a/strings/base_coroutine_system.h
+++ b/strings/base_coroutine_system.h
@@ -25,19 +25,22 @@ WINRT_EXPORT namespace winrt
 
             bool await_suspend(std::experimental::coroutine_handle<> handle)
             {
-                m_queued = m_dispatcher.TryEnqueue(m_priority, [handle]
-                    {
-                        handle();
-                    });
+                bool const queued = m_dispatcher.TryEnqueue(m_priority, [handle, this]
+                {
+                    m_ready.wait();
+                    handle();
+                });
+                m_queued = queued;
+                m_ready.signal(); // all member variables potentially invalid after this point
 
-                return m_queued;
+                return queued;
             }
 
         private:
-
             Windows::System::DispatcherQueue const& m_dispatcher;
             Windows::System::DispatcherQueuePriority const m_priority;
             bool m_queued{};
+            slim_event m_ready;
         };
 
         return awaitable{ dispatcher, priority };
--- a/strings/base_lock.h
+++ b/strings/base_lock.h
@@ -118,4 +118,29 @@ WINRT_EXPORT namespace winrt
     private:
         impl::condition_variable m_cv{};
     };
+
+    struct slim_event
+    {
+        slim_event(slim_event const&) = delete;
+        slim_event const& operator=(slim_event const&) = delete;
+        slim_event() noexcept = default;
+
+        void signal() noexcept
+        {
+            signaled = true;
+            WINRT_WakeByAddressAll(&signaled);
+        }
+
+        void wait() noexcept
+        {
+            while (!signaled)
+            {
+                bool not_ready = false;
+                WINRT_WaitOnAddress(&signaled, &not_ready, sizeof(not_ready), 0xFFFFFFFF); // INFINITE
+            }
+        }
+
+    private:
+        bool signaled = false;
+    };
 }
--- a/strings/base_extern.h
+++ b/strings/base_extern.h
@@ -62,6 +62,8 @@ extern "C"
     void    __stdcall WINRT_WakeAllConditionVariable(winrt::impl::condition_variable* cv) noexcept;
     void*   __stdcall WINRT_InterlockedPushEntrySList(void* head, void* entry) noexcept;
     void*   __stdcall WINRT_InterlockedFlushSList(void* head) noexcept;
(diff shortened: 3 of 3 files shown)


strings/base_coroutine_system.h
```diff
@@ -25,19 +25,22 @@ WINRT_EXPORT namespace winrt
...
+                m_ready.signal(); // all member variables potentially invalid after this point
+
+                return queued;
             }
 
         private:
-
             Windows::System::DispatcherQueue const& m_dispatcher;
             Windows::System::DispatcherQueuePriority const m_priority;
             bool m_queued{};
```
oldnewthing: If we really wanted to be miserly, we could merge `m_queued` with `m_ready` to save a byte. It's a false economy due to padding, but just saying that there is a no-extra-memory alternative.

```cpp
template<typename T>
struct slim_state<T>
{
    static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8);
    slim_state(T initial = {}) : value(initial) { }
   T get() { return value; }
   void set(T v) { value = v; WINRT_WakeByAddressAll(&value); }
   void wait_for(T v) { T original; while ((original = value) != v) WINRT_WaitOnAddress(&value, &original, sizeof(original), 0xFFFFFFFF); }
   void wait_for_not(T v) { while (value == v) WINRT_WaitOnAddress(&value, &v, sizeof(v), 0xFFFFFFFF); }
private: T value;
}
```

Then `m_queued` would be a tri-state of "unknown", "queued", and "not queued". `await_suspend` would `set()` the result, and the lambda would `wait_for_not(unknown)`.

kennykerr: Yeah, I'd rather keep it simple unless there's a measurable advantage.

strings/base_lock.h
```diff
@@ -118,4 +118,29 @@ WINRT_EXPORT namespace winrt
...
+        {
+            while (!signaled)
+            {
+                bool not_ready = false;
+                WINRT_WaitOnAddress(&signaled, &not_ready, sizeof(not_ready), 0xFFFFFFFF); // INFINITE
+            }
+        }
+
+    private:
+        bool signaled = false;
```
oldnewthing: May need to use `std::atomic<>`(can get away with `memory_order_relaxed`)  in case architecture does not support single-byte updates.

kennykerr: Wouldn't that introduce unnecessary memory barriers? Those tend to be more costly than a few extra bytes in memory. Perhaps uint32_t instead of bool is sufficient.

oldnewthing: Even with `uint32_t` we need a release barrier so that the compiler won't reorder the store to `m_queue` after the store to `m_ready.signaled`. If the reordering occurred, then the lambda could observe `signaled` and run ahead to complete the coroutine, and then the first thread crashes trying to write to an already-freed `m_queue`.

oldnewthing: Aha, I can use ` std::atomic_signal_fence(release)` which is purely a compiler reordering directive. No code is generated by a signal fence; it merely tells the compiler not to reorder across the fence.

oldnewthing: I think this means I should withdraw `slim_event` since it does not have the same memory consistency guarantees as a regular event. (Regular events are full barriers.)

kennykerr: Sounds good.

oldnewthing:
I'm going to abandon this PR and start up a fresh one with the revised approach as discussed here.

PR #403, opened 2019-10-27
Remove unnecessary xaml and windows includes.

jlaanstra:
Using code from the Windows.UI.Xaml namespace in a viewmodel is considered bad form as the goal of MVVM is to separate view from viewmodel (Windows.UI.Xaml.Data being the exception as the PropertyChanged classes live there). Furthermore, the default templates don't require any classes from that namespace. This makes the code simpler and doesn't by default include the massive Xaml namespace which significantly increases the pch.

Also defines WIN32_LEAN_AND_MEAN;WINRT_LEAN_AND_MEAN; to reduce unnecessary headers from Windows and WinRT.

### DIFF
--- a/vsix/ItemTemplates/ViewModel/ViewModel.cpp
+++ b/vsix/ItemTemplates/ViewModel/ViewModel.cpp
@@ -4,9 +4,6 @@
 #include "$safeitemname$.g.cpp"
 #endif
 
-using namespace winrt;
-using namespace Windows::UI::Xaml;
-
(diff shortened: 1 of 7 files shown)


kennykerr:
Project templates should also define [WINRT_LEAN_AND_MEAN](https://github.com/microsoft/xlang/pull/613) to further reduce build time.

jlaanstra:
@kennykerr done.

Scottj1s:
thanks for the tidy up!

PR #404, opened 2019-10-29
Add CppWinRTGenerateWindowsMetadata property to more quickly determine if a project outputs a winmd.

jlaanstra:
This PR adds the CppWinRTGenerateWindowsMetadata property to more quickly determine if a project outputs a winmd. This allows a project to determine this without evaluating all its project references for the presence of static libraries.

Tested NuGetTest.sln as well as various razzle projects.

Fixes #399

PR #405, opened 2019-10-29
Add a better default for CppWinRTGenerateWindowsMetadata for CX projects.

jlaanstra:
In #404, we added a property to enable Windows Metadata. We calculate a default is one is not specified, but there are a few cases where we know we should disable it without the need to calculate anything.

If the project specifically set a property to indicate it's a CX project, we will never generate a WinMD and we can therefore set CppWinRTGenerateWindowsMetadata to false.


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -23,6 +23,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <XamlNamespace Condition="'$(XamlNamespace)' == ''">Windows.UI.Xaml</XamlNamespace>
         <XamlMetaDataProviderIdl Condition="'$(XamlMetaDataProviderIdl)'== ''">$(GeneratedFilesDir)XamlMetaDataProvider.idl</XamlMetaDataProviderIdl>
         <XamlMetaDataProviderCpp Condition="'$(XamlMetaDataProviderCpp)'== ''">$(GeneratedFilesDir)XamlMetaDataProvider.cpp</XamlMetaDataProviderCpp>
+
+        <!-- For CX projects, CppWinRT will never output a winmd-->
+        <!-- NOTE: We don't set a default here as the default requires evaluation of project references
+             and this is done by the ComputeCppWinRTResolvedWinMD target. -->
+        <CppWinRTGenerateWindowsMetadata Condition="'$(CppWinRTGenerateWindowsMetadata)' == '' AND '$(CppWinRTProjectLanguage)' == 'C++/CX' ">false</CppWinRTGenerateWindowsMetadata>
```
Scottj1s: this line is redundant, same with the pair of checks below (line 33) - checking XamlLanguage == 'C++' should suffice

PR #409, opened 2019-11-07
Fixing CppWinRT_IncludePath and the include paths that build on it to point to the generated headers

manodasanW:
Currently CppWinRT_IncludePath points to the includes in the installed Windows SDK and not to the dynamically generated headers.  This causes for scenarios where if a nuget SDK only has a partial set of winmds, the headers for the excluded winmds are found in the installed Windows SDK path.  This change fixes that by making CppWinRT_IncludePath and a few include paths that build on it to point to the generated headers location by removing the C++/WinRT SDK path and including the generated headers path.

It would have been ideal to do this in a props, but given we rely on InitDir which is set after our props, I ended up doing this as part of a target.

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -32,6 +32,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
         <GeneratedFilesDir Condition="'$(GeneratedFilesDir)' == ''">$(IntDir)Generated Files\</GeneratedFilesDir>
         <!--Override SDK's uap.props setting to ensure version-matched headers-->
+        <CppWinRT_OldIncludePath>$(CppWinRT_IncludePath)</CppWinRT_OldIncludePath>
         <CppWinRT_IncludePath>$(GeneratedFilesDir)</CppWinRT_IncludePath>
         <!--TEMP: Override NuGet SDK's erroneous setting in uap.props -->
         <WindowsSDK_MetadataFoundationPath Condition="('$(WindowsSDK_MetadataFoundationPath)'!='') And !Exists($(WindowsSDK_MetadataFoundationPath))">$(WindowsSDK_MetadataPathVersioned)</WindowsSDK_MetadataFoundationPath>
@@ -40,7 +41,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             $(GetTargetPathDependsOn);ComputeCppWinRTResolvedWinMD;CppWinRTResolvedWinMD;
         </GetTargetPathDependsOn>
         <PrepareForBuildDependsOn>
-            $(PrepareForBuildDependsOn);CppWinRTVerifyKitVersion;
+            $(PrepareForBuildDependsOn);CppWinRTVerifyKitVersion;CppWinRTFixSDKIncludes;
         </PrepareForBuildDependsOn>
         <!-- Note: Before* targets run before Compute* targets. -->
         <BeforeMidlCompileTargets>
@@ -115,6 +116,28 @@ Copyright (C) Microsoft Corporation. All rights reserved.
                 <AdditionalOptions>%(ClCompile.AdditionalOptions) /DWINRT_NO_MAKE_DETECTION</AdditionalOptions>
             </ClCompile>
         </ItemGroup>
+    </Target>
+  
+    <!-- Fix includes to point to the generated C++/WinRT headers rather than the one in the SDK -->
+    <Target Name="CppWinRTFixSDKIncludes"
+            BeforeTargets="SetBuildDefaultEnvironmentVariables"
+            Condition="'$(CppWinRT_IncludePath)' != '$(CppWinRT_OldIncludePath)'">
+        <ItemGroup>
+            <_FixedSDKIncludes Remove="@(_FixedSDKIncludes)"/>
+            <_FixedSDKIncludes Include="$(WindowsSDK_IncludePath)"/>
+            <_FixedSDKIncludes Condition="'$(CppWinRT_OldIncludePath)' != ''" Remove="$(CppWinRT_OldIncludePath)"/>
+            <_FixedSDKIncludes Include="$(CppWinRT_IncludePath)"/>
+        </ItemGroup>
+        <ItemGroup>
+            <_FixedIncludes Remove="@(_FixedIncludes)"/>
+            <_FixedIncludes Include="$(IncludePath)"/>
+            <_FixedIncludes Condition="'$(CppWinRT_OldIncludePath)' != ''" Remove="$(CppWinRT_OldIncludePath)"/>
+            <_FixedIncludes Include="$(CppWinRT_IncludePath)"/>
+        </ItemGroup>
+        <PropertyGroup>
+            <WindowsSDK_IncludePath>@(_FixedSDKIncludes)</WindowsSDK_IncludePath>
+            <IncludePath>@(_FixedIncludes)</IncludePath>
+        </PropertyGroup>    
     </Target>
 
     <!-- Target used only to evaluate CppWinRTGenerateWindowsMetadata if it doesn't already have a value -->
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -32,7 +32,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
         <GeneratedFilesDir Condition="'$(GeneratedFilesDir)' == ''">$(IntDir)Generated Files\</GeneratedFilesDir>
         <!--Override SDK's uap.props setting to ensure version-matched headers-->
-        <CppWinRT_OldIncludePath>$(CppWinRT_IncludePath)</CppWinRT_OldIncludePath>
         <CppWinRT_IncludePath>$(GeneratedFilesDir)</CppWinRT_IncludePath>
         <!--TEMP: Override NuGet SDK's erroneous setting in uap.props -->
         <WindowsSDK_MetadataFoundationPath Condition="('$(WindowsSDK_MetadataFoundationPath)'!='') And !Exists($(WindowsSDK_MetadataFoundationPath))">$(WindowsSDK_MetadataPathVersioned)</WindowsSDK_MetadataFoundationPath>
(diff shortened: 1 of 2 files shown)


kennykerr:
Nice 👍

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -116,6 +117,28 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
     </Target>
+  
+    <!-- Fix includes to point to the generated C++/WinRT headers rather than the one in the SDK -->
+    <Target Name="CppWinRTFixSDKIncludes"
+            BeforeTargets="SetBuildDefaultEnvironmentVariables"
+            Condition="'$(CppWinRT_IncludePath)' != '$(CppWinRT_OldIncludePath)'">
+        <ItemGroup>
+            <_FixedSDKIncludes Remove="@(_FixedSDKIncludes)"/>
+            <_FixedSDKIncludes Include="$(WindowsSDK_IncludePath)"/>
+            <_FixedSDKIncludes Condition="'$(CppWinRT_OldIncludePath)' != ''" Remove="$(CppWinRT_OldIncludePath)"/>
```
Scottj1s: condition is probably redundant

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -116,6 +117,28 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+            <_FixedSDKIncludes Include="$(CppWinRT_IncludePath)"/>
+        </ItemGroup>
+        <ItemGroup>
+            <_FixedIncludes Remove="@(_FixedIncludes)"/>
+            <_FixedIncludes Include="$(IncludePath)"/>
+            <_FixedIncludes Condition="'$(CppWinRT_OldIncludePath)' != ''" Remove="$(CppWinRT_OldIncludePath)"/>
```
Scottj1s: same here

jlaanstra:
Why not simply override WindowsSDK_IncludePath using

```
<CppWinRT_IncludePath Condition="'$(CppWinRT_IncludePath)' == ''">path here</CppWinRT_IncludePath>
```

manodasanW:
> Why not simply override WindowsSDK_IncludePath using
> 
> ```
> <CppWinRT_IncludePath Condition="'$(CppWinRT_IncludePath)' == ''">path here</CppWinRT_IncludePath>
> ```

WindowsSDK_IncludePath is set as part of a props file, so it has already been set using the previous CppWinrt_IncludePath when we override it in the targets and therbey its value doesn't change.  And we can't move this definition to a props file without removing the reliance on InitDir or redefining it ourselves.

nuget/Microsoft.Windows.CppWinRT.targets
manodasanW: Agree, resolved.

nuget/Microsoft.Windows.CppWinRT.targets
manodasanW: Agree, resolved.

jlaanstra:
What is InitDir? Did you mean InitDir?

manodasanW:
> What is InitDir? Did you mean IntDir?

Yes I did mean IntDir.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -116,6 +117,28 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <ItemGroup>
+            <_FixedSDKIncludes Remove="@(_FixedSDKIncludes)"/>
+            <_FixedSDKIncludes Include="$(WindowsSDK_IncludePath)"/>
+            <_FixedSDKIncludes Remove="$(CppWinRT_OldIncludePath)"/>
+            <_FixedSDKIncludes Include="$(CppWinRT_IncludePath)"/>
+        </ItemGroup>
+        <ItemGroup>
+            <_FixedIncludes Remove="@(_FixedIncludes)"/>
+            <_FixedIncludes Include="$(IncludePath)"/>
+            <_FixedIncludes Remove="$(CppWinRT_OldIncludePath)"/>
```
jlaanstra: Does this actually work since IncludePath is a property? I don't think there is magic property to itemgroup conversion is there? So if my include path Contains multiple paths this wouldn't work?

jlaanstra:
The props file already uses $(GeneratedFilesDir), so I think you just define it there undconditionally like this:

```
<CppWinRT_IncludePath>$(GeneratedFilesDir)</CppWinRT_IncludePath>
```

Since the SDK doesn't overwrite, this should always take precedence.

nuget/Microsoft.Windows.CppWinRT.targets
manodasanW: There is actually a magic semi-colon separated list to item group conversion and vice versa which Scott informed me of.  See it below in action.

image

manodasanW:
> The props file already uses $(GeneratedFilesDir), so I think you can just define it there unconditionally like this:
> 
> ```
> <CppWinRT_IncludePath>$(GeneratedFilesDir)</CppWinRT_IncludePath>
> ```
> 
> Since the SDK doesn't overwrite, this should always take precedence.

That was one of the first things I tried.  But, I noticed when I used GeneratedFileDir in the props file, it had no value.  It is only defined in the target files with a dependency on IntDir.  I assumed looking at what it is used as part of in the props file (ClCompile) that it is evaluated later after the targets file defines it,  While since I am trying to use it to define a property, it is defined at that point and has no value.

jlaanstra:
$(GeneratedFilesDir) is definitely defined earlier than the targets file. The definition in the CppWinRT targets is merely a backup and having it undefined up till there would actually cause issues with the included module.g.cpp that is included by default in most templates.

My understanding is that you effectively don't want the sdk path to be included. If so, I would simply set it to a dummy value, in the props file, like 
```
<CppWinRT_IncludePath>$(MSBuildProjectDirectory)\</CppWinRT_IncludePath>
```
That value is always available and will make it effectively a no-op since the project directory is always included anyway.

Later in the targets file you will actually know exactly where the files are placed, so you can add that path to the AdditionalIncludeDirectories.

manodasanW:
> $(GeneratedFilesDir) is definitely defined earlier than the targets file. Various system props also use it. The definition in the CppWinRT targets is merely a backup and having it undefined up till there would actually cause issues with the included module.g.cpp that is included by default in most templates.
> 
> My understanding is that you effectively don't want the sdk path to be included. If so, I would simply set it to a dummy value, in the props file, like
> 
> ```
> <CppWinRT_IncludePath>$(MSBuildProjectDirectory)\</CppWinRT_IncludePath>
> ```
> 
> That value is always available and will make it effectively a no-op since the project directory is always included anyway.
> 
> Later in the targets file you will actually know exactly where the files are placed, so you can add that path to the AdditionalIncludeDirectories.

I tried removing the is empty condition we had on the GeneratedFilesDir in the target to see if we were replacing a previous value and it seems at least in a basic console app scenario with the C++/WinRT nuget package, the first time it is defined is in our targets file.

But with that aside, your idea of setting a fake value to prevent the SDK from setting it is definitely interesting.  I wouldn't set it to MSBuildProjectDirectory as a person may have a copy of a header file there that isn't supposed to be used and that can cause unexpected behavior.  But, setting it to a value like CppWinrtFakeIncludePath might be reasonable to expect not be an actual folder with headers.  This approach has similar disadvantages as the current approach with anything that uses CppWinrt_IncludePath in a props file has the wrong value, but it does avoid the MsBuild magic, is more straightforward, and allows to avoid any potential reference to the wrong set of C++/WinRT headers.

@Scottj1s Do you see any concerns I am missing with this proposed approach of setting a fake value instead?

jlaanstra:
> I tried removing the is empty condition we had on the GeneratedFilesDir in the target to see if we were replacing a previous value and it seems at least in a basic console app scenario with the C++/WinRT nuget package, the first time it is defined is in our targets file.

That seems problematic as our targets file doesn't modify the include path, so how does the compiler pickup the headers?

jlaanstra:
>  I wouldn't set it to MSBuildProjectDirectory as a person may have a copy of a header file there that isn't supposed to be used and that can cause unexpected behavior.

That's fair, though I think the project directory is pretty much always already included anyway. Any dummy value would work though. 

Would it be a good idea too to fix the SDK props too so we can simply turn the include path off with a boolean? That way over time, we don't need hacks anymore.

Scottj1s:
> @Scottj1s Do you see any concerns I am missing with this proposed approach of setting a fake value instead?

Aren't you still faced with a race between SDK and C++/WinRT nuget prop file inclusions?  If C++/WinRT props is imported first, no problem.  But if SDK is, then the bad value for CppWinRT_IncludePath is baked into WindowsSDK_IncludePath and it won't matter what you do with it.  Isn't that why you're having to fix things up after the fact?  Hacky yes, but seems necessary.

jlaanstra:
> Aren't you still faced with a race between SDK and C++/WinRT nuget prop file inclusions? 

C++/WinRT nuget props are always included first

manodasanW:
> > I tried removing the is empty condition we had on the GeneratedFilesDir in the target to see if we were replacing a previous value and it seems at least in a basic console app scenario with the C++/WinRT nuget package, the first time it is defined is in our targets file.
> 
> That seems problematic as our targets file doesn't modify the include path, so how does the compiler pickup the headers?

The C++/WinRT headers there come from the additional includes which gets evaluated as part of CICompile after we set GeneratedFilesDir in targets.



> > I wouldn't set it to MSBuildProjectDirectory as a person may have a copy of a header file there that isn't supposed to be used and that can cause unexpected behavior.
> 
> That's fair, though I think the project directory is pretty much always already included anyway. Any dummy value would work though.
> 
> Would it be a good idea too to fix the SDK props too so we can simply turn the include path off with a boolean? That way over time, we don't need hacks anymore.

Thats possible but given that we probably will always need to set the boolean and an non empty string just in case the developer is using an older SDK that doesn't recognize the boolean, I would stick with the approach of setting a fake value.

ChrisGuzak:
> since the project directory is always included anyway.

the project directory is no in the INCLUDE path. if it was you could `#include "string"` (a standard library header). instead you must `#include <string>`. see https://github.com/isocpp/CppCoreGuidelines/pull/1456

PR #422, opened 2019-11-15
Single-parameter constructors should be explicit

### LINKED ISSUE #413: Generated C++/WinRT constructors should be marked explicit
One of our devs just ran into a bug where we have a WinRT type with a constructor that takes Object:

```
unsealed runtimeclass ItemsSourceView: Windows.UI.Xaml.Interop.INotifyCollectionChanged
{
    ItemsSourceView(Object source);
```

And then we had a function with parameter typed as `winrt::ItemsSourceView` but the caller passed `winrt::IInspectable`. This compiled but did unexpected things at runtime because the ItemsSourceView constructor was being run.

I think that the generated constructors should be marked explicit to help avoid this problem.

kennykerr:
Fixes #413

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -2777,7 +2777,8 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
                     {
                         method_signature signature{ method };
 
-                        w.write("        %(%);\n",
+                        w.write("        %%(%);\n",
+                            signature.params().size() == 1 ? "explicit " : "",
                             type_name,
                             bind<write_consume_params>(signature));
                     }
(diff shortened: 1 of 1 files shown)


oldnewthing:
0-parameter and N-parameter constructors can also be converting constructors. I think we should allow the the 0-parameter converting constructor, and I'm not sure about whether we should allow the N-parameter one.

```cpp
struct C {
  C();
  C(Int32);
  C(Int32, Int32);
};

void F(C c);
```

| Statement | Prior to this change | After this change | Desired? |
|-|-|-|-|
| `F({});` | Allowed | Allowed | Probably |
| `F(42);` | Allowed | Disallowed | Probably not |
| `F({ 42, 96 });` | Allowed | Allowed | Unclear |
| `C c;` | Allowed | Allowed | Yes |
| `C c(42);` | Allowed | Allowed | Yes |
| `C c(42, 96);` | Allowed | Allowed | Yes |
| `C c = {};` | Allowed | Allowed | Probably |
| `C c = 42;` | Allowed | Disallowed | Probably not |
| `C c = { 42, 96 };` | Allowed | Allowed | Unclear |

kennykerr:
@oldnewthing yes, I'm not planning on changing anything about initialization in general and rather leaving the C++ language defaults as is. The `nullptr_t` is not explicit as doing so would be a bigger breaking changing and the value of taking that on is questionable. Having single-parameter constructors explicit solves the chief usability/pit-of-failure issue here.

PR #432, opened 2019-11-26
Override GetResolvedWinMD to make sure all native build targets are aware of the C++/WinRT produced WinMD.

### LINKED ISSUE #425: Unable to add C++/WinRT Runtime Component to UWP C# Application
Just want to make sure that I'm not doing something wrong here. I have created a Runtime Component project , implemented a bare minimum of a single Runtime class (following the example here: https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/author-events), but i could not reference this project from a C# UWP Application within the same solution. Trying to add a reference to the project would just end up with the message of `The method or opration is not implemented` . Is this a known issue with Visual Studio or C++/WinRT? Project referencing from a C++/WinRT Blank application doesn't have this issue.

Directly referencing the winmd files under Debug\Merged\ will just result in a runtime exception of something not activated class. 

Also, if i tried to manually reference the project by editing the csproj file manually, I have 2 cases happening. One project works just fine and the other one is complaining about not having the DLL of the Runtime Time component under its `Debug` folder. 

I can try to upload both projects here (the one that kinda-works and the one that doesn't) later but if anybody have any pointers, that would be great.

jlaanstra:
Since the project templates no longer set the flags that enable the MSBuild logic to tell MSBuild about the CX generated WinMD, we have to override the GetResolvedWinMD target to ensure that VS and MSBuild pick up the correct WinMD project output. Unfortunately the target is not very extensible, so we have to override it.

VS uses this to resolve project output and without this change the PrimaryWinMDFullPath itemgroup is empty resulting in VS being unable to add project references to C++/WinRT projects.

## TODO:

- [x] Verified NuGetTest.sln in VS and using MSBuild.
- [x] Verified that a C# app can reference and activate types from RuntimeComponent.
- [x] Verified Razzle builds.

Fixes #425


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -590,4 +590,37 @@ $(XamlMetaDataProviderPch)
...
+        Name="CopyWinMDArtifactsOutputGroup"
+        DependsOnTargets="GetResolvedWinMD;CppWinRTResolvedWinMD"
+        Returns="@(CopyWinMDArtifactsOutputGroupOutput)">
+
+        <ItemGroup>
+            <CopyWinMDArtifactsOutputGroupOutput Include="@(WinMDFullPath)" />
+        </ItemGroup>
+    </Target>
+
+    <!-- VS uses this target to resolve the primary WinMD to determine if the project rteference is compatible. -->
```
kennykerr: `rteference` - typo.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -590,4 +590,37 @@ $(XamlMetaDataProviderPch)
         </Link>
     </ItemDefinitionGroup>
 
+    <!--
+    ============================================================
```
kennykerr: There are a few too many negatives in this sentence. Consider rewriting for clarity.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -308,16 +308,16 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
-$(XamlMarkupIdlImport)
-namespace $(RootNamespace)
-{
-    $(FullXamlMetadataProviderAttribute)runtimeclass XamlMetaDataProvider : [default] $(XamlNamespace).Markup.IXamlMetadataProvider
-    {
-        XamlMetaDataProvider()%3b
-    }
-}
-</XamlMetaDataProviderIdlLines>
+                // This file is generated by the build to support Xaml apps
```
Scottj1s: please restore formatting

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -335,11 +335,11 @@ namespace $(RootNamespace)
             <_PCH>@(ClCompile->Metadata('PrecompiledHeaderFile')->Distinct())</_PCH>
             <XamlMetaDataProviderPch Condition="'$(_PCH)'!=''">#include "$(_PCH)"</XamlMetaDataProviderPch>
             <XamlMetaDataProviderCppLines>
-// This file is generated by the build to support Xaml apps
-$(XamlMetaDataProviderPch)
-#include "XamlMetaDataProvider.h"
-#include "XamlMetaDataProvider.g.cpp"
-</XamlMetaDataProviderCppLines>
+                // This file is generated by the build to support Xaml apps
```
Scottj1s: please restore formatting

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -590,4 +590,37 @@ $(XamlMetaDataProviderPch)
...
+    ============================================================
+    Override the CopyWinMDArtifactsOutputGroup to return the 
+    correct WinMD. Not that we still need to call GetResolvedWinMD
+    to make sure that projects that only use CppWinRT to consume
+    WinRT, but not to produce the WinMD, will still return the 
+    WinMD file produced by the CX target GetResolvedWinMD.
+    ============================================================
+    -->
+
+    <Target Name="WinMDArtifactsProjectOutputGroup"
```
Scottj1s: what's the purpose of this empty target?

nuget/Microsoft.Windows.CppWinRT.targets
Scottj1s: I think there's a typo (which would remove one negative):  "Note that we..."

Scottj1s:
I'm a bit nervous about overriding all these internal CX targets.  That has the potential to make our nuget brittle (or at least specific to VS versions).  What's wrong with the alternative of just setting GenerateWindowsMetadata back to true for the RC project template?  You describe that as a "workaround" - how so?

nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: Whoops sorry about that.

nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: It is. WIll add a comment.

jlaanstra:
These are standard OutputGroup targets, meant to be overriden.

nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: Yup, this is a standard output group target used for WinMDs and known to VS. Added someclarification.

jlaanstra:
@Scottj1s the other option is for us to override the GetResolvedWinMD target, to make it aware of the C++/WInRT WinMD. Thoughts?

jlaanstra:
@kennykerr Added todo's], so you can see what I have verified.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -122,21 +119,51 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
             <CppWinRTGenerateWindowsMetadata Condition="'@(CppWinRTMdMergeInputs)'!= ''">true</CppWinRTGenerateWindowsMetadata>
             <CppWinRTGenerateWindowsMetadata Condition="'@(CppWinRTMdMergeInputs)'== ''">false</CppWinRTGenerateWindowsMetadata>
         </PropertyGroup>
     </Target>
 
-    <Target Name="CppWinRTResolvedWinMD"
-            Condition="'$(CppWinRTGenerateWindowsMetadata)' == 'true'"
+    <!-- This target overrides the GetResolvedWinMD target used to resolve the WinMD for native projects
+         so it is aware of the C++/WinRT generated WinMD.
+         Since not every project that consumesd C++/WinRT uses it to generate a WinMD,
```
Scottj1s: typo

Scottj1s:
ok, let's give it a spin with vsix/razzle

test/nuget/TestApp/TestApp.vcxproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(ProjectDir)..\..\..\nuget\Microsoft.Windows.CppWinRT.props" />
```
kennykerr: This path seems rather brittle. No better way to know where this is store? Sometimes it's relative to the project and sometimes to the solution, no?

jlaanstra: For the test app it doesn't matter too much as we 100% control where the file is located. I can make it relative to $(SolutionDir) though if you prefer.

kennykerr: Ah, sorry I thought that was in the nuget files. Never mind.

PR #433, opened 2019-11-27
Workarounds to improve IntelliSense

### LINKED ISSUE #384: IntelliSense is getting slaughtered
Using the VS2019 Windows Runtime Component (C++/WinRT) template to create a component, you wind up with an unusable IntelliSense context.

Here are two variations.  The first, with nothing changed after the project is generated reports 5 errors.  The second, with only modest changes reports 5K errors.

image

image

kennykerr:
A collection of changes to address various issues with IntelliSense usability.

* Use `=` instead of `{ }` for variable templates to workaround IntelliSense bug
* Use `WINRT_IMPL_AUTO` to avoid poor IntelliSense performance due to return type deduction
* Address type conversion warnings from IntelliSense compiler

Also a tweak to avoid brittle ABI headers not defining GUID specializations.

Fixes #384

### DIFF
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -15,6 +15,12 @@
 
 #define WINRT_IMPL_SHIM(...) (*(abi_t<__VA_ARGS__>**)&static_cast<__VA_ARGS__ const&>(static_cast<D const&>(*this)))
 
+#ifdef __INTELLISENSE__
+#define WINRT_IMPL_AUTO(...) __VA_ARGS__
+#else
+#define WINRT_IMPL_AUTO(...) auto
+#endif
+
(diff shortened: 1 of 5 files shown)


strings/base_macros.h
```diff
@@ -15,6 +15,12 @@
 
 #define WINRT_IMPL_SHIM(...) (*(abi_t<__VA_ARGS__>**)&static_cast<__VA_ARGS__ const&>(static_cast<D const&>(*this)))
 
+#ifdef __INTELLISENSE__
```
LukaszMendakiewicz: Going through this macro route, it would be a "syntax changing macro" that probably should be included in the [hint file](https://docs.microsoft.com/en-us/cpp/build/reference/hint-files?view=vs-2019).

PR #445, opened 2019-12-07
Ensure we update the response files when the targets are run.

### LINKED ISSUE #444: C++/WinRT build targets don't play well with incremental builds.
Various build targets use the rsp files as output of a target, but an optimization to prevent unnecessary file writes to those files, prevents the output from those targets from being updated when the input change. This in turn causes the targets to be ran unnecessarily.

The fix should be to always write the rsp files when MSBuild determines that the target is out of date.

jlaanstra:
This PR makes sure that the Outputs of the various targets are updated when the targets run, to make sure the Outputs are newer than the Inputs after the target has ran.

/cc @rainersigwald to make sure I understand MSBuild's incremental build logic correctly.

Fixes #444


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -88,6 +88,13 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <IncludeCopyWinMDArtifactsOutputGroup>false</IncludeCopyWinMDArtifactsOutputGroup>
     </PropertyGroup>
 
+    <ItemDefinitionGroup>
+        <Midl Condition="'$(CppWinRTModernIDL)' != 'false'">
+            <AdditionalMetadataDirectories Condition="'%(AdditionalMetadataDirectories)' == '' And '$(WindowsSDK_MetadataFoundationPath)' != ''">$(WindowsSDK_MetadataFoundationPath);%(AdditionalMetadataDirectories)</AdditionalMetadataDirectories>
```
rainersigwald: Have you double-checked that this works as you expect? I think it will work if a project doesn't specify `AdditionalMetadataDirectories` for a given `Midl` item, but the old behavior (when it was in a target) was to prepend `$(WindowsSDK_MetadataFoundationPath)` unconditionally.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -395,8 +402,11 @@ $(XamlMetaDataProviderPch)
 
     <!--Insert Midl /references to Platform WinMDs, library reference WinMDs, and direct reference WinMDs-->
     <Target Name="CppWinRTSetMidlReferences"
-            DependsOnTargets="GetCppWinRTPlatformWinMDReferences;GetCppWinRTDirectWinMDReferences;GetCppWinRTProjectWinMDReferences;$(CppWinRTSetMidlReferencesDependsOn)">
-        <ItemGroup Condition="'$(CppWinRTModernIDL)' != 'false'">
+            Condition="'$(CppWinRTModernIDL)' != 'false'"
+            DependsOnTargets="GetCppWinRTPlatformWinMDReferences;GetCppWinRTDirectWinMDReferences;GetCppWinRTProjectWinMDReferences;$(CppWinRTSetMidlReferencesDependsOn)"
+            Inputs="@(CppWinRTDirectWinMDReferences);@(CppWinRTStaticProjectWinMDReferences);@(CppWinRTDynamicProjectWinMDReferences);@(CppWinRTPlatformWinMDReferences)"
```
rainersigwald: It's not directly related to this PR, but you might consider adding `$(MSBuildAllProjects)` to the `Inputs` in these targets. That's a common pattern used as a proxy for "changing a property or item". A common case where this is helpful is "deleting a reference"--the current project output is up to date with all of the inputs, but should be rebuilt anyway because the list of inputs has changed. There's no general solution to this in MSBuild but `$(MSBuildAllProjects)` helps.

nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: I believe this is just there to handle a case where C++/WinRT is used outside of a UWP since Microsoft.Cpp.AppContainerApplication.props already sets this for a UWP.

nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: Cool! I will look into this.

jlaanstra: @Rainersigwald I am trying to find example of how the Midl or the C++/C# compiler targets handle deletes from the Compile or ClCompile itemgroup but it doesn't look like they look at `$(MSBuildAllProjects)`. Am I missing something?

rainersigwald: C# is here: https://github.com/dotnet/roslyn/blob/b0df544ac729ff09031d3cf38e2cbfb3d69436a5/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets#L13

Note that with C#/VB, we have another more sophisticated approach, where we write the hash of the list to a file and include that as an input (microsoft/msbuild#1328). That's often overkill, and still misses the "some property passed to the compiler changed" scenario, so you still need `$(MSBuildAllProjects)`.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -5,6 +5,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+    <PropertyGroup>
+        <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
```
rainersigwald: Do you support MSBuild versions below 16.0? If not, delete this entirely: we do it automatically for you as of microsoft/msbuild#3605. If you do, please condition these out: unnecessarily adding to this property has been observed to cause a fair amount of wasted memory in VS scenarios.

jlaanstra: We do support VS2017 as far as I know which is why I added this.

jlaanstra: Like this?

nuget/Microsoft.Windows.CppWinRT.props
```diff
@@ -6,6 +6,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 <!--Set compiler and linker options for projects requiring C++/WinRT. -->
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+    <PropertyGroup>
+        <!-- Only do this for MSBuild versions below 16.0
+             as it is since done automatically, see https://github.com/microsoft/msbuild/pull/3605-->
+        <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)'  &lt;= '15'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
```
rainersigwald: This should be ok. We added `$(MSBuildVersion)` in 15.6, which is a little clearer, but it wasn't supported in 15.0 which is still in support for a little while.

PR #450, opened 2019-12-11
Provide the correct path to the implementation dll.

### LINKED ISSUE #449: Project references provide the wrong name for the implementation dll if the RootNamespace is different from the ProjectName.
We set the Implementation for a WinMD reference to:

```
<Implementation Condition="'$(TargetExt)' == '.dll'">$(WinMDImplementationPath)$(RootNamespace)$(TargetExt)</Implementation>
```

while it should be set to:

```
<Implementation Condition="'$(TargetExt)' == '.dll'">$(WinMDImplementationPath)$(TargetName)$(TargetExt)</Implementation>
```

jlaanstra:
This PR updates the path to the implementation dll for a WinMD to use ```$(TargetPath)``` instead of ```$(RootNamespace)``` which is incorrect when they differ. Also added a test project to make sure this keeps working going forward.

Fixes #449


kennykerr:
Hi @jlaanstra could you resolve this branch conflict. Then we can merge into master.

jlaanstra:
@kennykerr done.

sylveon:
My project does not build without this PR, is there any estimate as to when I should see a release on NuGet with this fix?

kennykerr:
@sylveon I just published 2.0.200107.1 with a rollup of fixes. 

https://www.nuget.org/packages/Microsoft.Windows.CppWinRT/

sylveon:
This version works, thank you!

PR #459, opened 2019-12-25
Check for STA and NA when deciding to call directly or post a callback.

### LINKED ISSUE #458: Coroutine resumption logic can cause RPC_E_WRONG_THREAD exception.
Some WinRT methods return back on a thread with apartment type APPTYPE_NA. This in turn can cause E_RPC_WRONG_THREAD when calling GetResults(). In the example code below `GetToastCollectionAsync` returns on a thread with apartment type APPTYPE_NA, causing the issue:

```
    winrt::fire_and_forget Test()
    {
        co_await resume_background();

        auto notificationManager = ToastNotificationManager::GetDefault();
        ToastCollectionManager collectionManager = notificationManager.GetToastCollectionManager();
        ToastCollection collection = co_await collectionManager.GetToastCollectionAsync(L"Test");

        InMemoryRandomAccessStream testStream{};
        auto decoder = co_await BitmapDecoder::CreateAsync(testStream);
    }
```

jlaanstra:
Checking for STA when deciding to post a callback or call directly, doesn't handle the case where the apartment type is APTTYPE_NA. This PR also adds a check for NA.

Fixes #458 

/cc @oldnewthing @kennykerr

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -291,10 +291,10 @@ WINRT_EXPORT namespace winrt
 
         void await_suspend(std::experimental::coroutine_handle<> handle) const
         {
-            impl::resume_apartment(sta_context, handle);
+            impl::resume_apartment(context, handle);
         }
 
-        com_ptr<impl::IContextCallback> sta_context = impl::sta_apartment_context();
+        com_ptr<impl::IContextCallback> context = impl::apartment_context();
     };
 
     [[nodiscard]] inline auto resume_after(Windows::Foundation::TimeSpan duration) noexcept
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -21,16 +21,16 @@ namespace winrt::impl
         return (0 == WINRT_CoGetApartmentType(&aptType, &aptTypeQualifier)) && ((aptType == 0 /*APTTYPE_STA*/) || (aptType == 3 /*APTTYPE_MAINSTA*/));
     }
 
-    inline bool is_mta() noexcept
+    inline bool requires_apartment_context() noexcept
     {
         int32_t aptType;
         int32_t aptTypeQualifier;
-        return (0 == WINRT_CoGetApartmentType(&aptType, &aptTypeQualifier)) && ((aptType == 1 /*APTTYPE_MTA*/));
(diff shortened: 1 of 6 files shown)


strings/base_coroutine_threadpool.h
```diff
@@ -21,6 +21,13 @@ namespace winrt::impl
         return (0 == WINRT_CoGetApartmentType(&aptType, &aptTypeQualifier)) && ((aptType == 0 /*APTTYPE_STA*/) || (aptType == 3 /*APTTYPE_MAINSTA*/));
     }
 
+    inline bool is_mta() noexcept
+    {
+        int32_t aptType;
+        int32_t aptTypeQualifier;
+        return (0 == WINRT_CoGetApartmentType(&aptType, &aptTypeQualifier)) && ((aptType == 1 /*APTTYPE_MTA*/));
```
YohDeadfall: Does it make sense to introduce an enum and have only one function to get the apartment type?

kennykerr:
The logic here is a little confusing and the PR misses the subtlety, but I think we can make it work. The `is_sta()` check is not really about checking for the STA as much as it is checking whether a call context needs to be captured at all. Reversing then does not quite achieve the original intent, as C++/WinRT can operate in a limited way without COM being loaded (at least initially). 

So I suspect you can get the desired behavior without perturbing unintended behavior simply by updating `is_sta` to include NA or by making `is_mta` mean anything that can be served by the thread pool e.g. MTA or no apartment.

strings/base_coroutine_threadpool.h
kennykerr: No, the question is simpler. Does a call context need to be captured.

jlaanstra:
Pushed some changes.

kennykerr:
Looks good. Just checking - we've confirmed this actually works with the API in question?

jlaanstra:
> Looks good. Just checking - we've confirmed this actually works with the API in question?

Correct, this change make the API in question work as expected.

kennykerr:
And you have a clean pass from `run_tests.cmd`?

jlaanstra:
> And you have a clean pass from `run_tests.cmd`?

D:\Dev\cppwinrt>run_tests.cmd x64 Release
===============================================================================
All tests passed (1972 assertions in 62 test cases)

===============================================================================
All tests passed (8 assertions in 2 test cases)

===============================================================================
All tests passed (11 assertions in 2 test cases)

===============================================================================
All tests passed (3775 assertions in 363 test cases)

===============================================================================
All tests passed (3 assertions in 1 test case)

===============================================================================
All tests passed (4 assertions in 1 test case)


D:\Dev\cppwinrt>

Issue #465, opened 2020-01-08
winrt::guid cannot be used with std::unordered_map


dfields-msft:
In order to use `winrt::guid` with `std::unordered_map`, `std::hash` must be specialized for `winrt::guid`.

PR #466, opened 2020-01-08
Specialize std::hash for winrt::guid

dfields-msft:
#465

### DIFF
--- a/strings/base_std_hash.h
+++ b/strings/base_std_hash.h
@@ -47,4 +47,12 @@ namespace std
     template<> struct hash<winrt::Windows::Foundation::IUnknown> : winrt::impl::hash_base {};
     template<> struct hash<winrt::Windows::Foundation::IInspectable> : winrt::impl::hash_base {};
     template<> struct hash<winrt::Windows::Foundation::IActivationFactory> : winrt::impl::hash_base {};
+    
+    template<> struct hash<winrt::guid>
+    {
+        size_t operator()(winrt::guid const& value) const noexcept
+        {
+            return winrt::impl::hash_data(&value, sizeof(value));
+        }
+    }
 }


strings/base_std_hash.h
```diff
@@ -47,4 +47,12 @@ namespace std
     template<> struct hash<winrt::Windows::Foundation::IUnknown> : winrt::impl::hash_base {};
     template<> struct hash<winrt::Windows::Foundation::IInspectable> : winrt::impl::hash_base {};
     template<> struct hash<winrt::Windows::Foundation::IActivationFactory> : winrt::impl::hash_base {};
```
oldnewthing: Maybe permit generalization to Point and other POD types by defining

```cpp
template<typename T>
struct hash_pod
{
  size_t operator()(T const& value) const noexcept
  {
    return winrt::impl::hash_data(&value, sizeof(value));
  }
}

template<> struct hash<guid> : winrt::impl::hash_pod {};
```

dfields-msft: I had the same thought... will do.

kennykerr: Please don't. While it may be a good idea to generate hash specializations for WinRT structs as well, we have to avoid introducing unnecessary template specializations/instantiations as that impacts build time. The suggestion above would double the number of specializations.

dfields-msft: Not sure I follow - the change I implemented just creates a helper for use when a specific POD type needs a hash. I only added one actual specialization, for `winrt::guid`.

dfields-msft: Are you saying that the additional template instantiation of `winrt::impl::hash_pod<winrt::guid>` is costly, and that it would be better to write this out individually for each new POD which needs specialization of `std::hash`?

kennykerr: Yes, the latter. One specialization for hash is unavoidable. Making it two per type is bloat.

kennykerr: Even with a single specialization, all of the hash specializations are very costly. That's why WINRT_LEAN_AND_MEAN strips them out.

dfields-msft: Got it; I'll revert the generalization.

PR #469, opened 2020-01-13
Support for Windows 7

kennykerr:
A number of projects need to support building a single binary that can run on Windows 10 all the way down to Windows 7. Strict requirements are placed on what these binaries may do to achieve this and the only mechanism that is allowed for light-up of features is `LoadLibrary`/`GetProcAddress`. Notably, the Anaheim (Edge) browser team need to use C++/WinRT but require this capability as they have a very rigid build system and cannot change the way the browser is built or add any new dependencies. All of their light-up must happen via `LoadLibrary`/`GetProcAddress`.

In order to support Windows 7 in this way, C++/WinRT is being updated to avoid linking to any Windows 8+ API functions directly. For example, the `RoOriginateLanguageException` function is now loaded via `LoadLibrary`/`GetProcAddress` and if that fails a fallback is provided. This works for all Windows 8+ functions with the exception of hstrings where performance necessitates an inlining of the hstring implementation. As C++/WinRT has done in the past with WinRT implementations like generic GUID calculations, C++/WinRT now includes a Windows 10 compatible implementation of hstring support that is inlined into the base library and can thus run down to Windows 7 without hurting Windows 10 performance.


MikeGitb:
Is there a performance regression expected when targeting Win10? If so, are there any benchmarks in place to quantify it?

kennykerr:
> Is there a performance regression expected when targeting Win10

No. Thanks to inlining of hstrings the performance on Windows 10 is at least as good as before (slightly faster in fact). The functions that are loaded dynamically are all in slow paths (e.g. error handling or behind the factory cache) where it doesn't matter.

MikeGitb:
Great to hear thanks.

strings/base_string.h
```diff
@@ -1,34 +1,143 @@
...
+    {
+        atomic_ref_count count;
+        wchar_t buffer[1];
+    };
+
+    inline void release_hstring(hstring_header* handle) noexcept
+    {
+        WINRT_ASSERT((handle->flags & hstring_reference_flag) == 0);
+
+        if (0 == --static_cast<shared_hstring_header*>(handle)->count)
```
DefaultRyan: Having a failfast to detect a negative refcount has turned out to be a highly valuable debugging tool and data point, not to mention instakilling the process before more memory corruption occurs. Consider doing that here.

kennykerr: Sounds good.

strings/base_string.h
```diff
@@ -1,34 +1,143 @@
...
+
+    inline hstring_header* create_hstring_on_heap(wchar_t const* value, uint32_t length)
+    {
+        if (!length)
+        {
+            return nullptr;
+        }
+
+        auto header = precreate_hstring_on_heap(length);
+        memcpy(header->buffer, value, sizeof(wchar_t) * length);
```
DefaultRyan: I can't think of a reason to use `memcpy` instead of `std::copy` or `std::copy_n`.

kennykerr: 👍

strings/base_string.h
```diff
@@ -1,34 +1,143 @@
...
 
-    inline void* create_string(wchar_t const* value, uint32_t const length)
+    inline shared_hstring_header* precreate_hstring_on_heap(uint32_t length)
     {
-        void* result = nullptr;
-        check_hresult(WINRT_WindowsCreateString(value, length, &result));
-        return result;
+        WINRT_ASSERT(length != 0);
+
+        auto header = static_cast<shared_hstring_header*>(WINRT_HeapAlloc(WINRT_GetProcessHeap(), 0, sizeof(shared_hstring_header) + sizeof(wchar_t) * length));
```
DefaultRyan: Both the multiplication and the addition can overflow. The Windows HSTRING implementation guards against it - so should this.

strings/base_string.h
```diff
@@ -1,34 +1,143 @@
...
+        uint32_t flags;
+        uint32_t length;
+        uint32_t padding1;
+        uint32_t padding2;
+        wchar_t const* ptr;
+    };
+
+    struct shared_hstring_header : hstring_header
+    {
+        atomic_ref_count count;
```
DefaultRyan: Nice to see `atomic_ref_count` pay for itself again. :)

kennykerr: Yeah, it's neat. 😄

strings/base_string.h
```diff
@@ -382,7 +501,15 @@ WINRT_EXPORT namespace winrt
 {
     inline bool embedded_null(hstring const& value) noexcept
     {
-        return impl::embedded_null(get_abi(value));
+        for (auto&& item : value)
```
DefaultRyan: Consider removing our own loop and replacing with
`return std::any_of(value.begin(), value.end(), [](auto elem){ return elem == 0; });`
Or similar.

strings/base_string_input.h
```diff
@@ -38,35 +35,23 @@ WINRT_EXPORT namespace winrt::param
         }
 
     private:
-        int32_t create_string_reference(wchar_t const* const data, size_t size) noexcept
+        void create_string_reference(wchar_t const* const data, size_t size) noexcept
         {
             auto size32 = static_cast<uint32_t>(size);
```
DefaultRyan: This could also overflow, technically, but it's only called from some other functions. Perhaps an assert is good enough here?

strings/base_string.h
kennykerr: 👍

strings/base_string.h
kennykerr: 👍

strings/base_string_input.h
kennykerr: Assert it is. 👍

DefaultRyan:
This project has reached the point that I think we need to beef up our testing strategy. Specifically, we need to figure out how to run our tests on multiple OS versions, like Win7, so that we have an automated way to ensure we don't break compat going forward.

I'm not going to block my review on this, but this should be strongly prioritized and done soon.

PR #477, opened 2020-01-16
Provide winrt_activation_handler to hook/mock activation

### LINKED ISSUE #476: Need a way to mock/hook WinRT runtime class activation for testing
Previously, we could override `WinRT_RoGetActivationFactory` to hook RoGetActivationFactory and return mock object instead. With #469 this is not supported. There is new reg-free activation support in that change, like searching dlls implicitly. For example, "Foo.Bar.Baz" is the runtime class name, then it will search "Foo.Bar.dll", "Foo.dll" and probe DllGetActivationFactory(). This only happens when it failed to activate system registered one.

In case of testing for authoring WinRT component, we need mocking mechanism for dependency runtime classes we can use even there is existing runtime class in the system.

kennykerr:
Fixes #476

This is required by the Windows shell team.


strings/base_activation.h
```diff
@@ -21,6 +21,11 @@ namespace winrt::impl
     template <typename Interface>
     hresult get_runtime_activation_factory(param::hstring const& name, void** result) noexcept
     {
+        if (winrt_activation_handler)
+        {
+            return winrt_activation_handler(*(void**)(&name), guid_of<Interface>(), result);
+        }
+
```
CarlosNihelton, line 28: This requires the custom activation handler to deal with all possible activation scenarios. In certain kinds of tests that might be an overkill. It would be nice to have a way for the custom handler to return a magic value signifying that it cannot handle that specific activation and the default activation procedure should just continue, something like:

```cpp
        if (winrt_activation_handler)
        {
            if (auto hr = winrt_activation_handler(*(void**)(&name), guid_of<Interface>(), result);
                hr != E_NOTIMPL) {
                    return hr;
                }
        }
        ...

Scottj1s: I agree - I've had need for this flexibility myself

PR #483, opened 2020-01-17
Clang conformance

kennykerr:
Some tweaks for Clang conformance. As a bonus, the Clang compiler picked up a source bug.

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -560,7 +560,7 @@ WINRT_EXPORT namespace winrt
         return pointer;
     }
 
-    [[noreturn]] inline void terminate() noexcept
+    inline void terminate() noexcept
     {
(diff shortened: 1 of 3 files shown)


strings/base_error.h
```diff
@@ -560,7 +560,7 @@ WINRT_EXPORT namespace winrt
         return pointer;
     }
 
-    [[noreturn]] inline void terminate() noexcept
+    inline void terminate() noexcept
```
DefaultRyan: Really? This one is surprising to me.

kennykerr: Yes, Clang can't tell that the function won't return and thus rejects the [[noreturn]]. Unfortunately, there seems to be no way to mark a lambda as [[noreturn]].

PR #493, opened 2020-01-25
Remove newline from the CAExcludePath.

### LINKED ISSUE #492: CAExcludePath doesn't work when using C++/WinRT NuGet package.
When the C++/WinRT NuGet is added to a project, headers in the CAExcludePath are no longer excluded from code analysis and will result in extra warnings.

jlaanstra:
CAExcludePath is used to set an environment variable. It looks like the way the build targets override the value might introduce a newline, causing the environment variable to not work. Defining on a single line fixes the issue.

Fixes #492

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -45,6 +45,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <CppWinRT_IncludePath>$(GeneratedFilesDir)</CppWinRT_IncludePath>
         <!--TEMP: Override NuGet SDK's erroneous setting in uap.props -->
         <WindowsSDK_MetadataFoundationPath Condition="('$(WindowsSDK_MetadataFoundationPath)'!='') And !Exists($(WindowsSDK_MetadataFoundationPath))">$(WindowsSDK_MetadataPathVersioned)</WindowsSDK_MetadataFoundationPath>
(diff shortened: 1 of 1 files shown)


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -45,6 +45,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <CppWinRT_IncludePath>$(GeneratedFilesDir)</CppWinRT_IncludePath>
         <!--TEMP: Override NuGet SDK's erroneous setting in uap.props -->
         <WindowsSDK_MetadataFoundationPath Condition="('$(WindowsSDK_MetadataFoundationPath)'!='') And !Exists($(WindowsSDK_MetadataFoundationPath))">$(WindowsSDK_MetadataPathVersioned)</WindowsSDK_MetadataFoundationPath>
+        <!-- CAExcludePath is used to set an environment variable, so make sure this is defined on a single line. -->
+        <CAExcludePath>$(GeneratedFilesDir);$(CAExcludePath)</CAExcludePath>
```
kennykerr: The position of the XML element changed. Does that matter?

The order within the include path also changed. Now GeneratedFilesDir comes first. Does that matter?

jlaanstra: I updated the order to match how the system props files add path to the variable. Shouldn't matter. The location is practically the same. All properties below the old and new location are used as the input to DependsOnTargets so the order doesn't matter. It is also not used in between.

PR #495, opened 2020-01-27
C++ modules conformance

kennykerr:
Now that Visual C++ 16.5 Preview 2 has decent support for C++ modules, I'm able to once again tweak C++/WinRT to support C++ modules as an alternative to precompiled headers. This is still experimental as there remain some minor issues, but it is getting very close. 

The `cppwinrt.exe` compiler will now generate a `winrt/winrt.ixx` file that may be compiled as a module as follows:

```
cl /c /I . /experimental:module /EHsc /bigobj /std:c++17 /await winrt/winrt.ixx
```

This will compile all of the projected namespaces into a single `winrt` module. This step takes quite a while - around 2 minutes on my machine. But once that's built, much like a PCH, you can import the module into any (compatible) TU very quickly as follows:

```
cl /experimental:module /EHsc /std:c++17 /await app.cpp winrt.obj
```

Here's a sample app that you can use to test this:

```C++
import winrt;
#include <stdio.h>

#pragma comment (lib, "oleaut32")

// Note: these two are only needed when compiling with modules.
#pragma comment (lib, "ole32")
#pragma comment (lib, "advapi32")

using namespace winrt;

int main()
{
    Uri uri(L"http://kennykerr.ca");
    printf("%ls\n", uri.Domain().c_str());
}
```

Note the extra libs that are required when compiling with modules. This is a C++ compiler bug.

There's also an issue with coroutines - waiting for a solution from the C++ team.


mediabuff:
Great! Another optimization that can help is to separate the platform header generation and app header generation - as the former changes infrequently. Additionally, if the platform headers can be shared across projects - this will help in the build speeds as well.

mediabuff:
I tried it with the new pull.

1>------ Build started: Project: cppwinrt_modules, Configuration: Debug Win32 ------
1>winrt.ixx
1>d:\agent\_work\8\s\src\vctools\Compiler\CxxFE\sl\p1\c\module\writer.cpp:5261: sorry: not yet implemented
1>c:\temp\cppwinrt_modules\cppwinrt_modules\Debug\Generated Files\winrt\winrt.ixx(491,1): fatal error C1001: Internal compiler error.
1>(compiler file 'd:\agent\_work\8\s\src\vctools\Compiler\CxxFE\sl\p1\c\module\utilities.h', line 39)
1> To work around this problem, try simplifying or changing the program near the locations listed above.

kennykerr:
@mediabuff You need to use Visual C++ 16.5 Preview 2 or later.

mediabuff:
Actually, it 16.5/P2.  But may be it's related to this error;

' c:\temp\cppwinrt_modules\cppwinrt_modules\main.cpp(1,13): error C2235: mismatching target architecture for compiled module interface for 'winrt' from 'winrt.ifc'
1>    c:\temp\cppwinrt_modules\cppwinrt_modules\main.cpp(1,13): error C2230: could not find module 'winrt''

image

kennykerr:
Could be. It looks like you're trying to build this from the IDE. I would suggest following the instructions *exactly* as I demonstrated above from a single dev command prompt. Keep in mind that C++ module support is still very new and experimental.

PR #497, opened 2020-01-29
Add support for generic interfaces on non-generic types (runtime classes)

Scottj1s:
Next up: add support for types returning generic interfaces from GetRuntimeClassName (e.g., C++/WinRT collection helpers)

### DIFF
--- a/natvis/object_visualizer.cpp
+++ b/natvis/object_visualizer.cpp
@@ -7,11 +7,7 @@ using namespace Microsoft::VisualStudio::Debugger;
 using namespace Microsoft::VisualStudio::Debugger::Evaluation;
 using namespace std::literals;
 using namespace winrt;
-using namespace winmd::impl;
 using namespace winmd::reader;
-
-template <typename...T> struct overloaded : T... { using T::operator()...; };
-template <typename...T> overloaded(T...)->overloaded<T...>;
 
 #define IID_IInspectable L"AF86E2E0-B12D-4C6A-9C5A-D7AA65101E90"
 #define IID_IStringable  L"96369F54-8EB6-48F0-ABCE-C1B211E627C3"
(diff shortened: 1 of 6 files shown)


natvis/object_visualizer.cpp
```diff
@@ -424,89 +375,85 @@ void GetInterfaceData(
...
-                        }
-
-                        propDisplayType = std::wstring(L"winrt::") + cppTypename;
-                        if(get_category(type) == category::class_type)
+                    // Types come back from winmd files with '.', need to be '::'
+                    // Ex. Windows.Foundation.Uri needs to be Windows::Foundation::Uri
+                    auto fullTypeName = ns + "::" + name;
+                    wchar_t cppTypename[500];
+                    size_t i, j;
+                    for (i = 0, j = 0; i < (fullTypeName.length() + 1); i++, j++)
```
kennykerr: Would this this loop not be a lot less wonky if `cppTypename` were a `std::wstring`?

Scottj1s: is there a simpler way to replace all occurrences of "." with "::" in a std::wstring?

kennykerr: No. :)

I just mean, it might be simpler as follows:

```C++
std::wstring transform(std::string_view source)
{
    std::wstring destination;

    for (auto c : source)
    {
        if (c == '.')
        {
            destination += L"::";
        }
        else
        {
            destination += c;
        }
    }

    return destination;
}
```

No raw buffers and wonky indexes to manage.

Scottj1s: in exchange for lots of reallocations, but yes looks nicer

kennykerr: https://en.cppreference.com/w/cpp/string/basic_string/reserve

PR #506, opened 2020-01-30
Fix bug where CppWinRT would drop static library references.

### LINKED ISSUE #504: A CX runtime component can't reference types from a C++/WinRT static library.
When a CX runtime component references a C++/WinRT static library that produces a WinMD, the project reference is dropped.

jlaanstra:
A WinMD file produced by a static library gets merged into the WinMD of the Dll project consuming the library. Because they often share a root namespace, the files might have the same name and they can overwrite each other in the output folder.

The way this was fixed is to simply remove the static lib from the list of project references. However this causes issues for the case where a CX dll project references a C++/WinRT static library as it doesn't see the reference. A better fix is to mark the reference as private so it won't be copied to the output directory of the Dll project.

## TODO:
- [x] Verify razzle.
- [x] Verify Visual Studio.

Fixes #504

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -67,7 +67,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             CppWinRTCopyWinMDToOutputDirectory;
         </AfterMidlTargets>
         <ResolveAssemblyReferencesDependsOn>
-            $(ResolveAssemblyReferencesDependsOn);GetCppWinRTProjectWinMDReferences;CppWinRTRemoveStaticLibraries;
+            $(ResolveAssemblyReferencesDependsOn);GetCppWinRTProjectWinMDReferences;CppWinRTMarkStaticLibrariesPrivate;
         </ResolveAssemblyReferencesDependsOn>
         <!-- Note: Before* targets run before Compute* targets. -->
         <BeforeClCompileTargets>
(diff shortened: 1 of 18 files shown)


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -204,15 +204,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <Message Text="GetResolvedWinMD: @(WinMDFullPath->'%(FullPath)')" Importance="$(CppWinRTVerbosity)"/>
     </Target>
 
-    <!-- Static library reference files are merged into the project that
-         references it. They shouldn't be included as actual project references.
-         Make sure that we've ran the GetCppWinRTProjectWinMDReferences target
-         and then remove them. -->
-    <Target Name="CppWinRTRemoveStaticLibraries"
-            DependsOnTargets="GetCppWinRTProjectWinMDReferences"
+    <!-- Static library reference WinMDs are merged into the project WinMD that
```
ChrisGuzak: referenced

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -204,15 +204,18 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+         references it and might have the same name because they often share namespace. 
+         Therefore they shouldn't be copied to the output folder
+         because they might override files in the output folder with the
+         same name, causing missing types. -->
```
ChrisGuzak: consider "missing types when linking"

PR #520, opened 2020-02-05
Remove use of umbrella libraries

### LINKED ISSUE #488: Remove all umbrella libs from project templates as they are unreliable
See #487 

Not only are they a moving target, the umbrella libs in the forthcoming SDK are actually broken. Best to avoid them entirely. Fortunately, C++/WinRT does not depend on any functions offered only through umbrella libs and only requires `ole32.lib` and `oleaut32.lib` beyond the default libs that the Visual C++ includes. `ole32.lib` itself is only required for Visual Studio debugger support.

May want to consider inlining `IIDFromString` as it is only used by NATVIS and that would avoid the extra link dependency for Visual Studio builds.

https://gist.github.com/kennykerr/6c948882de395c25b3218ad8d4daf362

wjk:
This reference was actually in the NuGet package, not the project templates.

Fixes #488.


nuget/CppWinrtRules.Project.xml
```diff
@@ -33,7 +33,7 @@
 
   <BoolProperty Name="CppWinRTLibs"
                 DisplayName="Umbrella Library"
-                Description="Adds the WindowsApp.lib umbrella library for Windows Runtime imports"
+                Description="Adds import libraries required for Windows Runtime support."
```
kennykerr: The pattern appears to be to *not* have a trailing `.`.

PR #522, opened 2020-02-10
IVector's IndexOf should handle queries for nullptr more gracefully

### LINKED ISSUE #518: IndexOf should return false when Xaml provides a nullptr value
For compatibility with Xaml, The IndexOf implementation that C++/WinRT provides for `convertible_observable_vector` should return false if the value is a nullptr. This avoids an exception/error that a developer cannot easily deal with.

kennykerr:
Fixes #518

### DIFF
--- a/strings/base_collections_vector.h
+++ b/strings/base_collections_vector.h
@@ -100,7 +100,7 @@ namespace winrt::impl
 
         bool IndexOf(Windows::Foundation::IInspectable const& value, uint32_t& index) const
         {
-            if constexpr (!std::is_base_of_v<Windows::Foundation::IInspectable, T>)
+            if constexpr (!std::is_base_of_v<Windows::Foundation::IUnknown, T>)
             {
                 if (value == nullptr)
                 {
--- a/strings/base_collections_vector.h
+++ b/strings/base_collections_vector.h
@@ -100,6 +100,14 @@ namespace winrt::impl
 
         bool IndexOf(Windows::Foundation::IInspectable const& value, uint32_t& index) const
         {
+            if constexpr (!std::is_base_of_v<Windows::Foundation::IInspectable, T>)
+            {
+                if (value == nullptr)
+                {
+                    return false;
+                }
+            }
+
             return IndexOf(unbox_value<T>(value), index);
         }
 
--- a/test/test/single_threaded_observable_vector.cpp
+++ b/test/test/single_threaded_observable_vector.cpp
@@ -217,6 +217,22 @@ TEST_CASE("single_threaded_observable_vector")
         REQUIRE((vector_i.IndexOf(2, index) && index == 1));
         index = 0;
         REQUIRE((vector_o.IndexOf(box_value(2), index) && index == 1));
+
+        REQUIRE(false == vector_o.IndexOf(nullptr, index));
+    }
+    {
+        Uri one(L"http://example.com/1");
+        Uri two(L"http://example.com/2");
+
+        IObservableVector<Uri> vector_i = single_threaded_observable_vector<Uri>({ one, nullptr, two });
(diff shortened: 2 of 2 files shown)


strings/base_collections_vector.h
```diff
@@ -100,6 +100,14 @@ namespace winrt::impl
...
         {
+            if constexpr (!std::is_base_of_v<Windows::Foundation::IUnknown, T>)
+            {
+                if (value == nullptr)
+                {
+                    return false;
+                }
+            }
+
             return IndexOf(unbox_value<T>(value), index);
```
oldnewthing: This will still crash if somebody searches for, say, an integer in an `IVector<hstring>`. Should return "not found".

I was thinking more along the lines of

```cpp
template<typename T>
std::optional<T> try_unbox(Windows::Foundation::IInspectable const& value)
{
        if (value)
        {
            if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, T>)
            {
                if (auto temp = value.try_as<T>())
                {
                    return temp;
                }
            }
            else if constexpr (std::is_enum_v<T>)
            {
                if (auto temp = value.try_as<Windows::Foundation::IReference<T>>())
                {
                    return temp.Value();
                }

                if (auto temp = value.try_as<Windows::Foundation::IReference<std::underlying_type_t<T>>>())
                {
                    return static_cast<T>(temp.Value());
                }
            }
            else
            {
                if (auto temp = value.try_as<Windows::Foundation::IReference<T>>())
                {
                    return temp.Value();
                }
            }
            return {};
        }
        else
        {
            // nullptr is valid for reference types.
            if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, T>)
            {
                return T{nullptr};
            }
            else
            {
                // nullptr is not valid for value types.
                return {};
            }
        }
}
```

And then we have

```cpp
bool IndexOf(Windows::Foundation::IInspectable const& value, uint32_t index) const
{
    auto unboxed = try_unbox<T>(value);
    if (!unboxed) return false; // wrong type, cannot possible be in the vector
    return IndexOf(unboxed.value(), index);
}

```

oldnewthing: (I think `try_unbox` would be generally useful.)

test/test/single_threaded_observable_vector.cpp
```diff
@@ -217,6 +217,22 @@ TEST_CASE("single_threaded_observable_vector")
         REQUIRE((vector_i.IndexOf(2, index) && index == 1));
         index = 0;
         REQUIRE((vector_o.IndexOf(box_value(2), index) && index == 1));
+
+        REQUIRE(false == vector_o.IndexOf(nullptr, index));
```
oldnewthing:
```
REQUIRE(false == vector_o.IndexOf(box_value(L"hello"), index));
```

Looking for a string in a vector of integers should fail rather than crash.

strings/base_collections_vector.h
kennykerr: Yeah that seems useful.

kennykerr:
Closing this PR as adding a `try_unbox` function, while a good idea, is bigger than this PR and would require a lot more test coverage.

oldnewthing:
But it means that the fix is incomplete. If I do `stringVector.IndexOf(42)`, I get an exception instead of `false`.

kennykerr:
Sure I'm not disputing that it would not be more resilient if it were able to handle such varied input - it's just more work than I can do right now.

oldnewthing:
As a stopgap fix, how about

```cpp
bool IndexOf(Windows::Foundation::IInspectable const& value, uint32_t& index) const
{
    try
    {
        return IndexOf(unbox_value<T>(value), index);
    }
    catch (hresult_no_interface const&)
    {
        return false; // unbox_value failed, "value" wasn't even a boxed T
    }
}
```

kennykerr:
Looks good.

PR #523, opened 2020-02-12
Avoid using memcpy

kennykerr:
C++/WinRT is used in places where `memcpy` is no longer permitted.

### DIFF
--- a/strings/base_string.h
+++ b/strings/base_string.h
@@ -105,7 +105,7 @@ namespace winrt::impl
         }
 
         auto header = precreate_hstring_on_heap(length);
-        memcpy(header->buffer, value, sizeof(wchar_t) * length);
+        memcpy_s(header->buffer, sizeof(wchar_t) * length, value, sizeof(wchar_t) * length);
         return header;
     }
 


walbourn:
FWIW I've had some complaints from people on my GitHub projects about using ``memcpy_s`` instead of ``memcpy`` on some toolsets. Also, not all compilers treat ``memcpy_s`` as an intrinsic where it does for ``memcpy``.

kennykerr:
Yeah, its a silly restriction in my opinion. There's nothing more "secure" about `memcpy_s`. Ideally we could just use `std::copy` but VC's implementation gets confused by array vs pointer types.

DefaultRyan:
Last I checked VC doesn't get confused about array vs pointer types - it checks the size of the output array and fires a runtime assert if it's too small for the incoming data. This is easily dodged by casting the buffer from an array to a pointer before calling `std::copy_n`, like this:
`std::copy_n(value, length, static_cast<wchar_t*>(header->buffer));`

Here's two comparisons with optimizations disabled, and `_DEBUG` defined to light up STL iterator debugging.
Before:
https://godbolt.org/z/_2zj5R

After:
https://godbolt.org/z/-Udg6P

Still calling memmove. but note that the "after" disassembly no longer has that call to `_CrtDbgReport`

PR #530, opened 2020-02-17
Improve <experimental/coroutine> feature detection

christophpurrer:
When compiling with LLVM/Clang on Windows, while using the Microsoft STL headers the current ```#ifndef __clang__``` does not work. The final binary would contain duplicate symbols from the std::experimental namespace

Instead of checking for a specific compiler we can check if the header file exists

### DIFF
--- a/strings/base_includes.h
+++ b/strings/base_includes.h
@@ -22,7 +22,7 @@
 #include <directxmath.h>
 #endif
 
-#ifndef __clang__
+#if __has_include(<experimental/coroutine>)
 
 #include <experimental/coroutine>
 


kennykerr:
Thanks for the contribution. Please create an issue with repro steps for what does not work and then illustrate what such a change would fix and how existing scenarios would continue to work. Then we can discuss possible problems and solutions. As it stands, the proposed change would break existing uses of C++/WinRT.

christophpurrer:
Sure. Thanks for taking a quick look already. I have opened an issue here: https://github.com/microsoft/cppwinrt/issues/531z

I hope we can find a setting to distinguish between MSVC, Clang and Clang + Microsoft STL headers

PR #535, opened 2020-02-20
Use relative paths in C++ projects

wjk:
`$(SolutionDir)` may have an incorrect value during command-line builds. This method is guaranteed to work, no matter what directory the user is building from. I also took an opportunity to update the gitignore. Thanks!

### DIFF
--- a/cppwinrt.props
+++ b/cppwinrt.props
@@ -39,7 +39,7 @@
     <CppWinRTBuildVersion Condition="'$(CppWinRTBuildVersion)'==''">2.3.4.5</CppWinRTBuildVersion>
     <CmakePlatform>$(Platform)</CmakePlatform>
     <CmakePlatform Condition="'$(Platform)'=='Win32'">x86</CmakePlatform>
-    <CmakeOutDir Condition="'$(CmakeOutDir)'==''">$(SolutionDir)_build\$(CmakePlatform)\$(Configuration)</CmakeOutDir>
+    <CmakeOutDir Condition="'$(CmakeOutDir)'==''">$(MSBuildThisFileDirectory)_build\$(CmakePlatform)\$(Configuration)</CmakeOutDir>
     <CppWinRTDir>$(CmakeOutDir)\</CppWinRTDir>
     <CppWinRTDir Condition="'$(Platform)'=='ARM' or '$(Platform)'=='ARM64'">$(SolutionDir)_build\x86\$(Configuration)\</CppWinRTDir>
     <OutDir>$(CmakeOutDir)\</OutDir>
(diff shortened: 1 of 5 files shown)


cppwinrt/cppwinrt.vcxproj
```diff
@@ -115,7 +115,7 @@
     <RootNamespace>cppwinrt</RootNamespace>
     <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
-  <Import Project="$(SolutionDir)\cppwinrt.props" />
+  <Import Project="$(MSBuildThisFileDirectory)..\cppwinrt.props" />
```
kennykerr: This is not the only vcxproj in the solution that refers to `$(SolutionDir)\cppwinrt.props`. Are the rest immune?

cppwinrt.props
```diff
@@ -39,7 +39,7 @@
     <CppWinRTBuildVersion Condition="'$(CppWinRTBuildVersion)'==''">2.3.4.5</CppWinRTBuildVersion>
     <CmakePlatform>$(Platform)</CmakePlatform>
     <CmakePlatform Condition="'$(Platform)'=='Win32'">x86</CmakePlatform>
-    <CmakeOutDir Condition="'$(CmakeOutDir)'==''">$(SolutionDir)_build\$(CmakePlatform)\$(Configuration)</CmakeOutDir>
+    <CmakeOutDir Condition="'$(CmakeOutDir)'==''">$(MSBuildThisFileDirectory)_build\$(CmakePlatform)\$(Configuration)</CmakeOutDir>
```
kennykerr: How were you building cppwinrt that necessitated this change?

wjk: Command-line MSBuild in the `fast_fwd`, `prebuild`, and `cppwinrt` directories. Attempting to build the top-level solution was not working, as the test projects were somehow incompatible with the version of cppwinrt in the repo at this time. I didn't know how (nor had the time) to fix it, so I went with the next best method.

kennykerr: You need to build the cppwinrt project first - and then build a projection - and then build the rest. 

https://github.com/microsoft/cppwinrt/blob/master/build_test_all.cmd

cppwinrt/cppwinrt.vcxproj
wjk: Oops! You're right. I didn't check the test projects, as I wasn't planning on using those, and weren't building for me anyway (see my reply to your comment below).

cppwinrt.props
kennykerr: I'm happy to look at a solution that makes this simpler, but short of that we need to keep with the status quo.

wjk:
@kennykerr Honestly, that being said I would prefer that we close this PR; I will file one with README updates that instruct users to run that batch file rather than blindly attempting to build the solution. Thanks!

kennykerr:
Sounds good.

Issue #519, opened 2020-02-05
Create an MSBuild SDK for cpp/winrt


stevenbrix:
[MSBuild SDK's](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2019) are the bees knees. I'm not sure if there will be any equivalent work done for C++ projects like there is with .NET projects. It would be fantastic to have one for projects using Cpp/WinRT that removes a lot of the boilerplate MSBuild project file goo that is required

An example cpp/winrt UWP app .vcxproj:
```xml
<Project Sdk="Microsoft.Windows.CppWinRT">
   <PropertyGroup>
     <OutputType>AppContainerExe</OutputType>
     <TargetPlatformVersion>10.0.18632.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17134.0</TargetPlatformMinVersion>
   </PropertyGroup>
</Project>
```

An example cpp/winrt Win32 Desktop app .vcxproj:
```xml
<Project Sdk="Microsoft.Windows.CppWinRT">
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
     <TargetPlatformVersion>10.0.18632.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17134.0</TargetPlatformMinVersion>
   </PropertyGroup>
</Project>
```

An example cpp/winrt library .vcxproj:
```xml
<Project Sdk="Microsoft.Windows.CppWinRT">
   <PropertyGroup>
     <OutputType>library</OutputType>
     <TargetPlatformVersion>10.0.18632.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.17134.0</TargetPlatformMinVersion>
   </PropertyGroup>
</Project>
```

kennykerr:
Thank Steven, that sounds neat. If you'd like to experiment and work on something like that for C++/WinRT that would be interesting, but I doubt we'll be able to find the time. Ideally, the Visual C++ team would consider such an approach. The C++ project system really needs some love.

jlaanstra:
SDK based projects heavily rely on the project system to support it (globbing etc). None of this is supported by the C++ project system as far as I know. @AugP is this something that is on the roadmap?

stevenbrix:
That doesn’t surprise me, although I wouldn’t think MSBuild would care? Or is it VS support that wouldn’t work?

I would be happy if it was everything sans globbing for ClCompile items

[...]

PR #536, opened 2020-02-21
Conditionalize adding of NuGet reference in item templates

wjk:
Does what the title says. If the `CppWinRTDisableAutoNuGetReference` MSBuild property is set to some variation of `true`, then the item templates will not automatically add a reference to the CppWinRT NuGet package. It is assumed that if this property is set, the user has already ensured these targets are imported through some other means. If that is not the case, a broken build is the expected result.

See https://github.com/microsoft/cppwinrt/issues/519#issuecomment-582474662 for some context. Once this is merged, I can start work on the MSBuild SDK discussed in that issue in earnest. Thanks!

### DIFF
--- a/vsix/CppWinRTPackageTemplateWizard.cs
+++ b/vsix/CppWinRTPackageTemplateWizard.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Reflection;
 using Microsoft.Build.Evaluation;
 using Microsoft.VisualStudio.TemplateWizard;
@@ -57,11 +58,30 @@ namespace Microsoft.Windows.CppWinRT
 
             if (addNuGetReference)
             {
-                Assembly asm = Assembly.Load("NuGet.VisualStudio.Interop, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a");
-                Type wizardType = asm.GetType("NuGet.VisualStudio.TemplateWizard");
+                try
+                {
+                    Assembly asm = Assembly.Load("NuGet.VisualStudio.Interop, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a");
+                    Type wizardType = asm.GetType("NuGet.VisualStudio.TemplateWizard");
 
-                IWizard nugetWizard = (IWizard)Activator.CreateInstance(wizardType);
-                nugetWizard.RunStarted(automationObject, replacementsDictionary, runKind, customParams);
+                    IWizard nugetWizard = (IWizard)Activator.CreateInstance(wizardType);
+                    nugetWizard.RunStarted(automationObject, replacementsDictionary, runKind, customParams);
+                }
+                catch (Exception ex)
+                {
+#if DEBUG
+                    string path = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), "exception.txt");
+                    using (Stream stream = File.Open(path, FileMode.Create))
+                    using (StreamWriter writer = new StreamWriter(stream))
+                    {
+                        writer.WriteLine($"Caught exception: {ex.GetType().FullName}");
+                        writer.WriteLine(ex.Message);
+                        writer.WriteLine();
+                        writer.WriteLine(ex.StackTrace);
+                    }
+#endif
+
+                    throw new WizardCancelledException();
+                }
             }
         }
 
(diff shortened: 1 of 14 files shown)


vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            catch (Exception ex)
+            {
+                ShowExceptionDialog(ex);
+            }
+        }
+
+        public void RunStarted(object automationObject, Dictionary<string, string> replacementsDictionary, WizardRunKind runKind, object[] customParams)
+        {
+            // Unfortunately, we cannot forward this call to the NuGet template wizard because we can't get the EnvDTE.Project right now.
```
Scottj1s: I assume that limitation is benign?

wjk: The NuGet wizard implementation [does run some code in RunStarted()](https://github.com/NuGet/NuGet.Client/blob/f67a6083f8a49502f52f0c50492db03452785359/src/NuGet.Clients/NuGet.VisualStudio.Implementation/VsTemplateWizard.cs#L317-L339). However, the packages don't actually get installed [until ProjectFinishedGenerating() is called](https://github.com/NuGet/NuGet.Client/blob/f67a6083f8a49502f52f0c50492db03452785359/src/NuGet.Clients/NuGet.VisualStudio.Implementation/VsTemplateWizard.cs#L260-L290). I tested this PR locally, and it does work. That being said, I can certainly refactor this code to call RunStarted() unconditionally; I don't think that would harm anything.

wjk: This was done.

wjk:
I have run into a very weird problem with this PR: When I add the Empty View Model template to a project, whether or not the NuGet reference is disabled, the header and C++ source files are neither added to the project file nor copied into the project's folder. The template is therefore broken, and I have no idea why.

Note that this problem did not occur yesterday on my desktop PC at home. (I am typing this from my laptop, which runs VS 16.5 Preview 3; my desktop PC at home runs VS 16.4.) It also does this if I use the template as it existed before this PR was created. I'd appreciate any pointers you can offer. Thanks!

kennykerr:
This is quite a substantial change, at least for a non-msbuild/vsix guy like me. Can you perhaps spend a bit more time describing the motivation and particulars of this change.

wjk:
In #519 we discussed the possibility of using an MSBuild SDK to reference cppwinrt. This would have the benefit of removing the janky packages.config usage from project that do so. However, the item templates include code that adds a reference to the NuGet package, thereby creating packages.config. If cppwinrt is already referenced through some other means (such as the MSBuild SDK), this would cause a duplicate copy of the MSBuild glue be included in the project. This PR adds a replacement wizard class to the item templates that, by default, calls through to the NuGet wizard class, which adds the NuGet package just as the existing templates do today (before this PR). However, if the `CppWinRTDisableAutoNuGetReference` property is set to true in the project file (either directly, or in an included targets file; the value is checked via the MSBuild evaluation API), the automatic NuGet reference in the item templates will be suppressed.

In terms of implementation, much of this PR's commits are refactoring of the vsix project to let me add the `CppWinRTPackageTemplateWizard` class, and have it picked up by Visual Studio at runtime. [This class](https://github.com/microsoft/cppwinrt/blob/37e8687df870cc778bc1ae0b44d5a095cc503720/vsix/CppWinRTPackageTemplateWizard.cs) is the only changed file that impacts actual functionality.

Personally, this PR (and the discussion of the MSBuild SDK) came about due to my deep-seated loathing of packages.config. The Visual C++ project system doesn't support PackageReference (even though I can kludge together code to trick it into working, doing so breaks the NuGet UI). Since I will be doing a lot of work with WinUI 3, I cannot use CMake due to https://github.com/microsoft/microsoft-ui-xaml/issues/1433.

I hope this clarifies my motivations behind this PR. Thanks!

wjk:
Honestly, after giving it a bit of thought I have figured out how to run cppwinrt, and compile WinUI XAML files if needed, using pure CMake. In the case of those XAML files, I can write out a temporary MSBuild project file that references simplified targets for XAML compilation ([something like this](https://github.com/windows-toolkit/Microsoft.Toolkit.Win32/blob/DesktopWindow/Microsoft.Toolkit.Win32.Xaml.targets)) and run it from within the CMake-generated build script. Therefore, this PR becomes unneeded.

One question for you, however: Once I get my CMake script for running cppwinrt finished, would you like it submitted as a PR to this repo? I think it would be useful to have it here so that people know it's an option. (I will of course send my XAML-compilation CMake scripts to Microsoft/windows-ui-xaml instead.) Thanks so much!

kennykerr:
Thanks for the experiments! I think a sample repo showing how to do Xaml with C++ and cmake would be outstanding. I don't think Scott will tolerate any cmake in this repo. 😄

PR #538, opened 2020-02-21
Apply EBO declspec to forward declaration 

Scottj1s:
as well as to specializations, to avoid inconsistent EBO application, which can also manifest as crashing ODR violations (not just unoptimized code).


jlaanstra:
Thanks for fixing this!

jlaanstra:
Can we push out a build with this fix to NuGet?

PR #541, opened 2020-02-25
Do not pass nullptr arrays to the ABI

oldnewthing:
For array parameters, ensure that nullptr is not passed as the array base, even for empty arrays. If the array is empty, the array base is not used, but RPC still requires it to be a non-null pointer.

We use a properly aligned non-null pointer as our fake array base, in case somebody cares about alignment.

Note that this fix does not address [out] parameters. The workaround for [out] parameters is more expensive (need to CoTaskMemAlloc(0)) and it is necessary only for out-of-process calls, so I don't want to incur the overhead for everybody. People returning empty com_arrays will have to remain vigilant.

Also fix BO in test Class when asked to fill an empty array.

### DIFF
--- a/strings/base_array.h
+++ b/strings/base_array.h
@@ -353,13 +353,15 @@ WINRT_EXPORT namespace winrt
     template <typename T>
     auto get_abi(array_view<T> object) noexcept
     {
+        auto data = object.size() ? object.data() : (T*)alignof(T);
+
         if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, T>)
         {
-            return (void**)object.data();
+            return (void**)data;
         }
         else
         {
-            return reinterpret_cast<impl::arg_out<std::remove_const_t<T>>>(const_cast<std::remove_const_t<T>*>(object.data()));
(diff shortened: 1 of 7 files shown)


cppwinrt/code_writers.h
```diff
@@ -575,7 +575,7 @@ namespace cppwinrt
                     w.write(param_name);
                     break;
                 case param_category::array_type:
-                    w.write("%.size(), get_abi(%)", param_name, param_name);
+                    w.write("%.size(), %.size() ? get_abi(%) : (decltype(get_abi(%)))&%", param_name, param_name, param_name, param_name, param_name);
```
kennykerr: Would it not be cleaner to update `get_abi` to return a non-null value when the array is empty? Then the logic is codified in one place rather than everywhere `get_abi` is used.

oldnewthing: For maximum compatibility, I limited the scope of the change to just the `consume_` paths. If you're okay with changing the behavior of `get_abi` globally, then go for it. I don't know what else `get_abi` is used for and didn't want to mess them up accidentally.

kennykerr: So the idea is that `get_abi` returns the ABI representation for the projected type. If that is a non-null value for arrays then `get_abi` should ensure that it faithfully returns a null-null value.

PR #544, opened 2020-02-28
Fix holes in IObservableVector<IInspectable>

oldnewthing:
ReplaceAll did not provide the strong exception guarantee: If not all the elements could be converted to T, the code left the vector in a partially-replaced state. Now it leaves the vector unchanged if the operation fails. (Note that vector_base::assign does not provide the strong exception guarantee if wrapping is required, so we can't quite declare victory yet.)

If the underlying vector is a vector of value types, then attempting to insert nullptr now fails with E_NOINTERFACE rather than crashing.

If the object's runtime type is not T, then IndexOf should report "not found" rather than reporting the index of the first nullptr (if T is a reference type) or crashing (if T is a value type). This fixes issue #518.

Fixed use-after-free bug in test, where we forgot to unregister event handlers, leaving them with references to destroyed local variables.

### DIFF
--- a/strings/base_collections_vector.h
+++ b/strings/base_collections_vector.h
@@ -102,7 +102,7 @@ namespace winrt::impl
         {
             try
             {
-                return IndexOf(extract_value(value), index);
+                return IndexOf(unbox_value<T>(value), index);
             }
             catch (hresult_no_interface const&)
             {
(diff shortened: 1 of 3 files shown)


strings/base_collections_vector.h
```diff
@@ -191,6 +198,18 @@ namespace winrt::impl
 
     private:
 
+        static auto extract_value(Windows::Foundation::IInspectable const& value)
+        {
+            if constexpr (!std::is_base_of_v<Windows::Foundation::IInspectable, T>)
+            {
+                if (!value)
```
oldnewthing: Upon further consideration, I think this should go into unbox_value. New iteration coming.

PR #546, opened 2020-03-03
Harden operator wstring_view

kennykerr:
Some existing code relies on data returning a valid string even when the view is empty.

### DIFF
--- a/strings/base_string.h
+++ b/strings/base_string.h
@@ -233,7 +233,7 @@ WINRT_EXPORT namespace winrt
             }
             else
             {
-                return {};
+                return { L"", 0 };
             }
         }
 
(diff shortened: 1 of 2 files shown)


strings/base_string.h
```diff
@@ -233,7 +233,7 @@ WINRT_EXPORT namespace winrt
             }
             else
             {
-                return {};
+                return { L"", 0 };
```
kennykerr: When the `hstring` is empty, the `wstring_view`'s default constructor causes the `data` member to return a `nullptr` as per the spec: 

https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view

Scottj1s: Specs? who needs em!

PR #557, opened 2020-03-14
Remove redundant empty line

mcfi:


kennykerr:
All of the `base_*.h` headers start with an empty line so that when they are amalgamated they have proper spacing.

PR #563, opened 2020-03-22
"Starts_With" should not be marked as inline, and write_component_static_forwarder should be.

pi1024e:
The reason for this comes down to clang's interpretation. Starts_with is already entirely in a struct, while "write_component_static_forwarder" is not, and should be marked as such.

### DIFF
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -289,7 +289,7 @@ catch (...) { return winrt::to_hresult(); }
             bind<write_consume_args>(signature));
     }
 
-    inline void write_component_static_forwarder(writer& w, MethodDef const& method)
+    static void write_component_static_forwarder(writer& w, MethodDef const& method)
     {
         auto format = R"(        auto %(%)
         {
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -289,7 +289,7 @@ catch (...) { return winrt::to_hresult(); }
             bind<write_consume_args>(signature));
     }
 
-    void write_component_static_forwarder(writer& w, MethodDef const& method)
+    inline void write_component_static_forwarder(writer& w, MethodDef const& method)
     {
         auto format = R"(        auto %(%)
         {
--- a/cppwinrt/helpers.h
+++ b/cppwinrt/helpers.h
@@ -498,7 +498,7 @@ namespace cppwinrt
         return result;
     }
 
-    staitc bool has_fastabi_tearoffs(writer& w, TypeDef const& type)
+    static bool has_fastabi_tearoffs(writer& w, TypeDef const& type)
     {
         for (auto&& [name, info] : get_interfaces(w, type))
         {
--- a/cppwinrt/helpers.h
+++ b/cppwinrt/helpers.h
@@ -384,7 +384,6 @@ namespace cppwinrt
             info.overridable = overridable || has_attribute(impl, "Windows.Foundation.Metadata", "OverridableAttribute");
             info.base = base;
             info.generic_param_stack = generic_param_stack;
-            writer::generic_param_guard guard;
 
             switch (type.type())
             {
@@ -412,7 +411,7 @@ namespace cppwinrt
 
                     info.generic_param_stack.push_back(std::move(names));
 
-                    guard = w.push_generic_params(type_signature.GenericTypeInst());
+                    writer::generic_param_guard guard = w.push_generic_params(type_signature.GenericTypeInst());
                     auto signature = type_signature.GenericTypeInst();
                     info.type = find_required(signature.GenericType());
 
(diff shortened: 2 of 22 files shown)


cppwinrt/component_writers.h
```diff
@@ -289,7 +289,7 @@ catch (...) { return winrt::to_hresult(); }
             bind<write_consume_args>(signature));
     }
 
-    void write_component_static_forwarder(writer& w, MethodDef const& method)
+    inline void write_component_static_forwarder(writer& w, MethodDef const& method)
```
kennykerr: This should be static not inline. Look at the rest of the functions in this source file. This helps the compiler issue a warning if the function is unused.

cppwinrt/helpers.h
```diff
@@ -384,7 +384,6 @@ namespace cppwinrt
             info.overridable = overridable || has_attribute(impl, "Windows.Foundation.Metadata", "OverridableAttribute");
             info.base = base;
             info.generic_param_stack = generic_param_stack;
-            writer::generic_param_guard guard;
```
kennykerr: Why did you move this? This is an unrelated to change to inlining.

cppwinrt/helpers.h
```diff
@@ -498,7 +497,7 @@ namespace cppwinrt
         return result;
     }
 
-    bool has_fastabi_tearoffs(writer& w, TypeDef const& type)
+    inline bool has_fastabi_tearoffs(writer& w, TypeDef const& type)
```
kennykerr: These again should be `static` not `inline`.

cppwinrt/type_writers.h
```diff
@@ -44,12 +44,12 @@ namespace cppwinrt
         }
     };
 
-    bool operator==(type_name const& left, type_name const& right)
+    inline bool operator==(type_name const& left, type_name const& right)
```
kennykerr: These don't strictly need to be inline because they're only ever included in a single TU. And again, static will issue a compiler warning if unused.

strings/base_extern.h
```diff
@@ -1,8 +1,8 @@
 
-__declspec(selectany) int32_t(__stdcall* winrt_to_hresult_handler)(void* address) noexcept {};
-__declspec(selectany) void(__stdcall* winrt_suspend_handler)(void const* token) noexcept {};
-__declspec(selectany) void(__stdcall* winrt_resume_handler)(void const* token) noexcept {};
-__declspec(selectany) int32_t(__stdcall* winrt_activation_handler)(void* classId, winrt::guid const& iid, void** factory) noexcept {};
+inline __declspec(selectany) int32_t(__stdcall* winrt_to_hresult_handler)(void* address) noexcept {};
```
kennykerr: Why did this need to be inline? These are used internally and have very specific linkage requirements and I'd rather not mess with them.

cppwinrt/type_writers.h
pi1024e: For all the inlines? or just that one?

strings/base_events.h
```diff
@@ -248,7 +248,6 @@ namespace winrt::impl
                 ((*reinterpret_cast<impl::abi_t<I>**>(&object))->*(Method))(m_token);
             }
         }
-    private:
```
kennykerr: Empty line here.

strings/base_version.h
```diff
@@ -3,7 +3,7 @@
 
 // WINRT_version is used by Microsoft to analyze C++/WinRT library adoption and inform future product decisions.
 extern "C"
-__declspec(selectany)
+inline __declspec(selectany)
```
kennykerr: Why is this necessary?

cppwinrt/helpers.h
```diff
@@ -498,7 +498,7 @@ namespace cppwinrt
         return result;
     }
 
-    bool has_fastabi_tearoffs(writer& w, TypeDef const& type)
+    staitc bool has_fastabi_tearoffs(writer& w, TypeDef const& type)
```
kennykerr: Typo.

strings/base_version.h
pi1024e: Because it is in a struct

pi1024e: The compiler is implicitly in-lining this anyway.

kennykerr: Then leave it alone. Unless it's causing compilation issues, this has worked as-is for a few years and I don't want to disrupt that unnecessarily.

cppwinrt/component_writers.h
```diff
@@ -1208,4 +1208,4 @@ namespace winrt::@::implementation
                 slot);
         }
     }
-}
```
kennykerr: Please leave a new line at the end of each file. Some compilers don't like it otherwise.

PR #564, opened 2020-03-22
Reserve Memory for type

pi1024e:
In code-writers, we are looping through a range of type in types, and we emplace_back using that value.

For this reason, we should preallocate memory for the struct using reserve.

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -2581,6 +2581,7 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
 
         std::vector<complex_struct> structs;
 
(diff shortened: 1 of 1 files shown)


cppwinrt/code_writers.h
```diff
@@ -2581,6 +2581,7 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
 
         std::vector<complex_struct> structs;
 
+        structs.reserve(types.size());
```
kennykerr: Please put an empty line after but not before.

pi1024e:
Done!

pi1024e:
@kennykerr

PR #565, opened 2020-03-22
Replace old headers with modern ones

pi1024e:
Most files already have switched to use the modern variation, such as "<ctime>", "<cassert>", etc. However, a few files do not, so this PR fixes that.

### DIFF
--- a/test/catch.hpp
+++ b/test/catch.hpp
@@ -10291,8 +10291,8 @@ namespace Catch {
 
 #if defined(CATCH_PLATFORM_MAC) || defined(CATCH_PLATFORM_IPHONE)
 
-#  include <cassert>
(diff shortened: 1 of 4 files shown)


test/catch.hpp
```diff
@@ -10291,8 +10291,8 @@ namespace Catch {
 
 #if defined(CATCH_PLATFORM_MAC) || defined(CATCH_PLATFORM_IPHONE)
 
-#  include <assert.h>
-#  include <stdbool.h>
+#  include <cassert>
```
kennykerr: This header is take from https://github.com/catchorg/Catch2 - please update it there.

pi1024e:
Done! @kennykerr

PR #567, opened 2020-03-23
Join declaration and assignment

pi1024e:
The writer::generic_param_guard in helpers.h is only used in a specific switch statement, so declaring it at the start of a function is unnecessary.


kennykerr:
The guard needs to outlive the switch statement - that's the whole point of the guard - to push the generics onto the stack while the subsequent code writers rely on those generic params.

PR #579, opened 2020-04-08
Random-access iterators over IVector/IVectorView

oldnewthing:
The built-in iterators for IVector and IVectorView (and more generally anything that has a GetAt method) now support random-access operations. They do not meet all the requirements of a random-access iterator, most notably:
* `reference` is not `value_type&` or `value_type const&`.
* `operator->` is not supported.

Design notes:
* The `value_type` had been misdeclared as `T` (e.g. `IVector<Something>`) rather than the element type (`Something`).
* The `reference` is not actually a reference because the return value of `operator*` must be convertible to `reference`.
* The element index is unsigned so that the relational operators like `<` produce correct results on large containers.
* The `difference_type` is `ptrdiff_t` so that `b-a` and related operations work on large containers.
* Return types for iterator operations are explicit to ensure they conform to iterator requirements.
* We now rely on `std::reverse_iterator` to implement our reverse iterator.

### DIFF
--- a/strings/base_collections.h
+++ b/strings/base_collections.h
@@ -7,10 +7,12 @@ namespace winrt::impl
     struct fast_iterator
     {
         using iterator_category = std::input_iterator_tag;
-        using value_type = T;
+        using value_type = decltype(std::declval<T>().GetAt(0));
         using difference_type = ptrdiff_t;
-        using pointer = T * ;
-        using reference = T & ;
+        using pointer = value_type*;
+        using reference = value_type;
+
+        fast_iterator() noexcept : m_collection(nullptr), m_index(0) {}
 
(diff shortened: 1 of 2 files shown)


strings/base_collections.h
```diff
@@ -23,68 +25,98 @@ namespace winrt::impl
...
 
-        auto operator*() const
+        reference operator*() const
         {
-            return m_collection->GetAt(static_cast<uint32_t>(m_index));
+            return m_collection->GetAt(m_index);
         }
 
-        bool operator==(rfast_iterator const& other) const noexcept
+        reference operator[](difference_type n) const noexcept
```
kennykerr, line 79: This isn't `noexcept`.

PR #597, opened 2020-04-20
Append additional metadata in targets file, to preserve project settings

### LINKED ISSUE #593: $(GeneratedFilesDir) not added to AdditionalIncludeDirectories if AdditionalIncludeDirectories is explicitly defined
After adding the cppwinrt NuGet package to a C++ project, it's not guaranteed that `AdditionalIncludeDirectories` will reference `$(GeneratedFilesDir)`, leading to unexpected compilation failure. This impacts the SDK-to-NuGet onboarding experience.

# Environment
cppwinrt NuGet package 2.0.200316.3
Visual Studio 2019 16.5.1
Windows 10 10.0.19041.207

# Steps to reproduce
1. Create a new C++ Console App project.
2. In the project's properties, set the `Additional Include Directories` field to `C:\Windows` or any other valid path.
3. Add the cppwinrt NuGet package to the project.
4. Edit the `[project].cpp` file, add `#include "winrt/base.h"` to the list of includes at the top of the file.
5. Build the project
6. Observe compilation error `Error C1083 Cannot open include file: 'winrt/base.h': No such file or directory [...]`

# Expected
Compilation success, no errors.

# Actual
* Compilation error: `Error	C1083	Cannot open include file: 'winrt/base.h': No such file or directory`
* Incorrect/unclear guidance in `readme.txt` after NuGet installation

# Additional information / thoughts

While troubleshooting this issue, I got confused, partially due to the `readme.txt` indicating:
> For any winmd file referenced by the project, C++/WinRT creates reference (consuming) projection headers. Client code can simply #include these headers, which are created in the Generated Files directory.

But in reality `CppWinRTMakePlatformProjection` seems to run in all cases, even without winmd references. I think modifying the `readme.txt` to clarify how projection headers are generated would help others that may run into issues.

Also missing from the `readme.txt` is mention of `$(GeneratedFilesDir)` and how it's required in scenarios where `AdditionalIncludeDirectories` is already explicitly set prior to install. Perhaps Microsoft could consider automagically appending `$(GeneratedFilesDir)` to the value of `AdditionalIncludeDirectories` in your MSBuild workflow *behind the scenes* given it's critical to success. (It's not clear if there are any scenarios in which a user should be able to override auto-generated projections.)

Scottj1s:
fixes #593


riverar:
In the case of a simple console app that simply:
* Adds the NuGet package
* Has a single winrt::init_apartment(); call

Projection headers get generated. As far as I know, in the above case, none of these are true (from README):
* NuGet package references containing winmd files
* Other project references producing winmd files
* Raw winmd file references

Which means this part of the README isn't true (emphasis mine):
> [...] For **any winmd file discovered above**, C++/WinRT creates reference (consuming) projection headers.

Am I missing something?

Scottj1s:
> In the case of a simple console app that simply:
> 
> * Adds the NuGet package
> * Has a single winrt::init_apartment(); call
> 
> Projection headers get generated. As far as I know, in the above case, none of these are true (from README):
> 
> * NuGet package references containing winmd files
> * Other project references producing winmd files
> * Raw winmd file references
> 
> Which means this part of the README isn't true (emphasis mine):
> 
> > [...] For **any winmd file discovered above**, C++/WinRT creates reference (consuming) projection headers.
> 
> Am I missing something?

The first bullet is true:
* Platform winmd files in the SDK (both MSI and NuGet)

These are not "referenced" directly by the project (so I'm avoiding that term), but are still "discovered" and processed by cppwinrt.exe to produced headers.

riverar:
Oh got it! The existence of the platform winmd files is the trigger.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -626,14 +620,21 @@ $(XamlMetaDataProviderPch)
...
-        <CppWinRTParameters>$(CppWinRTParameters) -fastabi</CppWinRTParameters>
-    </PropertyGroup>
-    <ItemDefinitionGroup Condition="'$(CppWinRTFastAbi)'=='true'">
-        <Link>
-            <AdditionalDependencies>%(AdditionalDependencies);$(CppWinRTPackageDir)build\native\lib\$(Platform)\cppwinrt_fast_forwarder.lib</AdditionalDependencies>
-        </Link>
+    <!--Append any additional item metadata after all default and project settings have been applied-->
+    <ItemDefinitionGroup>
+      <ClCompile>
+        <AdditionalOptions>%(AdditionalOptions) /bigobj /await</AdditionalOptions>
```
jlaanstra, line 626: I don't think we should force the use of await?

Scottj1s: This isn't new (just relocated) - without it, async projection headers fail.

jlaanstra: Same applies to /std:c++17 though, which we don't add here either.

Scottj1s: user could add /std::latest

PR #599, opened 2020-04-21
Weak references are objects too

oldnewthing:
If the source contributes to the module object count, then the weak reference also contributes, so that we do not unload prematurely.

Factored out module lock updating into helper base class `module_lock_updater` with specialization `update_module_lock`. This avoids a lot of code duplication.

### DIFF
--- a/strings/base_agile_ref.h
+++ b/strings/base_agile_ref.h
@@ -44,19 +44,39 @@ WINRT_EXPORT namespace winrt
 
 namespace winrt::impl
 {
-    struct agile_ref_fallback final : IAgileReference
+    template<bool UseModuleLock>
+    struct module_lock_updater;
+
+    template<>
+    struct module_lock_updater<true>
+    {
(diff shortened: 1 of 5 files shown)


oldnewthing:
Wondering why `weak_source_producer` was defined as
```cpp
    template <bool Agile>
    struct weak_source_producer
    {
    protected:
        weak_source<Agile> m_source;
    };
```

instead of

```cpp
    template <bool Agile>
    struct weak_source_producer : protected weak_source<Agile>
    {
    };
```
This alternate definition would greatly simplify `weak_ref::that()`:

```cpp
        weak_ref<Agile>* that() noexcept
        {
            return static_cast<weak_ref<Agile>*>(this);
        }
```

kennykerr:
@oldnewthing no particular reason - feel free to fix.

oldnewthing:
Oh, it's because `weak_source` is final.

Issue #600, opened 2020-04-21
Coroutine never resumes if awaiting an async op hosted by a separate process that has died


vidager:
WinRT object are often time hosted by remote servers.  When performing a co_await on an async method on a remote object, it's possible the server could terminate during that call.  If that happens, the coroutine will never resume after that co_await as the completed delegate will never be called.  

The C++/WinRT library should detect the server has died complete the async operation from the client side with a corresponding error.

PR #602, opened 2020-04-22
Detect RPC server crash when co_await'ing

oldnewthing:
During a `co_await` of an `IAsyncAction` or `IAsyncOperation`, if the completion delegate is released without ever being invoked, then assume that the remote server crashed. Resume the coroutine, and the RPC error will be raised when `await_resume()` calls `GetResults()`.

Note that this behavior applies only to `co_await`. If you manually attach a `Completed` handler, then it is your responsibility to detect server death.

The `disconnect_aware_handler` is basically the same as the lambda we had been using, just with the additional behavior of resuming the coroutine if it is destructed without ever having been invoked. We null out the coroutine handle on invoke. That lets us detect the error case in our destructor.

This fixes issue #600

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -97,11 +97,11 @@ namespace winrt::impl
 
     private:
         std::experimental::coroutine_handle<> m_handle;
-        impl::com_ref<impl::IContextCallback> m_context = apartment_context();
+        com_ptr<IContextCallback> m_context = apartment_context();
 
         void Complete()
         {
-            impl::resume_apartment(m_context, std::exchange(m_handle, {}));
+            resume_apartment(m_context, std::exchange(m_handle, {}));
         }
     };
 
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -76,6 +76,35 @@ namespace winrt::impl
         return async.GetResults();
     }
 
+    struct disconnect_aware_handler
+    {
+        disconnect_aware_handler(std::experimental::coroutine_handle<> handle)
+            : m_handle(handle) { }
+
+        disconnect_aware_handler(disconnect_aware_handler&& other)
+            : m_context(std::move(other.m_context))
+            , m_handle(std::exchange(other.m_handle, {})) { }
+
+        ~disconnect_aware_handler()
+        {
+            if (m_handle) Complete();
+        }
+
+        void operator()(Windows::Foundation::IAsyncInfo const&, Windows::Foundation::AsyncStatus)
+        {
+            Complete();
+        }
+
+    private:
+        std::experimental::coroutine_handle<> m_handle;
+        impl::com_ref<impl::IContextCallback> m_context = apartment_context();
+
+        void Complete()
+        {
+            impl::resume_apartment(m_context, std::exchange(m_handle, {}));
+        }
+    };
+
     template <typename Async>
     struct await_adapter
     {
@@ -88,10 +117,7 @@ namespace winrt::impl
 
         void await_suspend(std::experimental::coroutine_handle<> handle) const
         {
-            async.Completed([handle, context = impl::apartment_context()](auto&& ...)
-            {
-                impl::resume_apartment(context, handle);
-            });
+            async.Completed(disconnect_aware_handler{ handle });
         }
 
         auto await_resume() const
--- a/test/test/disconnected.cpp
+++ b/test/test/disconnected.cpp
@@ -129,6 +129,15 @@ struct non_agile_abandoned_action : implements<non_agile_abandoned_action, IAsyn
 {
     non_agile_abandoned_action(void* event_handle) : m_awaited(event_handle) {}
 
+    static fire_and_forget final_release(std::unique_ptr<non_agile_abandoned_action> self)
+    {
+        // The C++/WinRT m_handler is agile but not context-aware,
+        // so we need to make sure to release it from the context it
+        // was created from, which for this particular test is the MTA.
+        co_await resume_background();
+        // Now we can destruct.
+    }
+
     void Completed(AsyncActionCompletedHandler const& handler) {
         m_handler = handler;
         // Tell the test to disconnect the IAsyncAction, which simulates the server crash.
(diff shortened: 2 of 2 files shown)


strings/base_coroutine_foundation.h
```diff
@@ -76,6 +76,35 @@ namespace winrt::impl
...
+        }
+
+        void operator()(Windows::Foundation::IAsyncInfo const&, Windows::Foundation::AsyncStatus)
+        {
+            Complete();
+        }
+
+    private:
+        std::experimental::coroutine_handle<> m_handle;
+        impl::com_ref<impl::IContextCallback> m_context = apartment_context();
```
kennykerr, line 100: This can simply be `com_ptr`. No need for `impl::com_ref` unless you're not sure what kind of thing `T` is.

test/test/disconnected.cpp
```diff
@@ -122,3 +123,73 @@ TEST_CASE("disconnected")
...
+            action = make<non_agile_abandoned_action>(signal.get());
+        });
+
+    auto result = [](IAsyncAction action) -> IAsyncAction
+        {
+            co_await action;
+            co_return;
+        }(action.get());
+
+    REQUIRE_THROWS_AS(result.get(), hresult_error);
```
kennykerr, line 194: Worth asserting what specific error was raised?

test/test/disconnected.cpp
```diff
@@ -122,3 +123,73 @@ TEST_CASE("disconnected")
...
+    agile_ref<IAsyncAction> action;
+    InvokeInContext(private_context.get(), [&]()
+        {
+            action = make<non_agile_abandoned_action>(signal.get());
+        });
+
+    auto result = [](IAsyncAction action) -> IAsyncAction
+        {
+            co_await action;
+            co_return;
```
kennykerr, line 191: The `co_return` is superfluous.

test/test/disconnected.cpp
oldnewthing: Unfortunately, the error generated from our simulated server is not the same RPC error that is generated from a real server. A real server generates `RPC_E_SERVER_DIED` but our fake one generates `RPC_E_WRONG_THREAD`. I was happy to get even a *simulation* of an OOP server without having to create an actual OOP server.

test/test/disconnected.cpp
oldnewthing: I have a vague memory that an earlier version of the coroutine spec required `co_return` for all coroutines, so I get kind of paranoid about it. Looks like the current spec allows `co_return` to be elided for `return_void` promises.

test/test/disconnected.cpp
oldnewthing: Okay, this test is bogus. The marshaling error occurs even when I don't disconnect the context. Need to figure out how to demonstrate both success and failure.

oldnewthing: Okay, it's because `apartment_context` and `resume_apartment` do not understand COM custom contexts. They say "Oh look, I'm in the MTA, and I was captured from the MTA, so I don't need to do anything special and can just invoke directly!" But it can't because it's in the wrong context. I don't think we want to go so far as adding private context support to C++/WinRT (unless I can find a super-cheap way of doing it) since they are very rare. Will have to work around in the test.

kennykerr: This is why I thought of adding OOP server support first to make this testing easier, but if you can find a way to test this easily then go for it.

strings/base_coroutine_threadpool.h
```diff
@@ -21,16 +21,14 @@ namespace winrt::impl
         return (0 == WINRT_IMPL_CoGetApartmentType(&aptType, &aptTypeQualifier)) && ((aptType == 0 /*APTTYPE_STA*/) || (aptType == 3 /*APTTYPE_MAINSTA*/));
     }
 
-    inline bool requires_apartment_context() noexcept
+    inline auto current_context()
     {
-        int32_t aptType;
-        int32_t aptTypeQualifier;
-        return (0 == WINRT_IMPL_CoGetApartmentType(&aptType, &aptTypeQualifier)) && ((aptType == 0 /*APTTYPE_STA*/) || (aptType == 2 /*APTTYPE_NA*/) || (aptType == 3 /*APTTYPE_MAINSTA*/));
+        return capture<IContextCallback>(WINRT_IMPL_CoGetObjectContext);
```
oldnewthing, line 26: Note that this change means that you can no longer use the `winrt::apartment_context()` function or `co_await` without first initializing COM. The previous code forgave you and assumed that if COM was not initialized, then you were in the default MTA.

strings/base_coroutine_threadpool.h
```diff
@@ -21,16 +21,14 @@ namespace winrt::impl
...
     }
 
-    inline auto apartment_context()
+    inline auto is_default_mta_context(com_ptr<IContextCallback> const& context)
     {
-        return requires_apartment_context() ? capture<IContextCallback>(WINRT_IMPL_CoGetObjectContext) : nullptr;
+        return context == capture<IContextCallback>(WINRT_IMPL_CoGetDefaultContext, 1 /* APTTYPE_MTA */);
```
oldnewthing, line 31: AARGH! CoGetDefaultContext is not callable from UWP!

strings/base_coroutine_threadpool.h
ChrisGuzak: seems ok to me

strings/base_coroutine_threadpool.h
ChrisGuzak: Ignore my email. I now see why this is a problem. I filed [this task](https://microsoft.visualstudio.com/OS/_workitems/edit/26091860/) for you. How often will the WACK detect this? maybe it won't.

oldnewthing:
Backed out the custom context stuff due to UWP issues. Returning to just the disconnection detection.

strings/base_coroutine_threadpool.h
oldnewthing: The problem is more than WACK. It's the SDK. Windowsapp.lib doesn't contain CoGetDefaultContext. You'd need a new SDK to get a version that links.

PR #603, opened 2020-04-23
Use EncodePointer to obfuscate event token values

dunhor:


strings/base_events.h
```diff
@@ -492,7 +492,7 @@ WINRT_EXPORT namespace winrt
 
         event_token get_token(delegate_type const& delegate) const noexcept
         {
-            return event_token{ reinterpret_cast<int64_t>(get_abi(delegate)) };
+            return event_token{ reinterpret_cast<int64_t>(WINRT_IMPL_EncodePointer(get_abi(delegate))) };
```
dunhor, line 495: One further possible option might be to call `DecodePointer` on the token in `remove` and then compare pointers so that there's only a single call out instead of one per entry.

Issue #480, opened 2020-01-17
array_view<const T> should be constructible from array_view<T>


DefaultRyan:
The following code doesn't compile, but should:
```
using namespace winrt;
using namespace winrt::Windows::Foundation;
IInspectable repackage(IPropertyValue pv)
{
  com_array<uint8_t> data;
  pv.GetUInt8Array(data);
  return PropertyValue::CreateUInt8Array(data);
}
```
This breaks because no conversion exists from `array_view<T>` to `array_view<const T>`.

Adding this constructor for `array_view` should fix things.
```
template <typename U, std::enable_if_t<std::is_convertible_v<U*, T*>, int> = 0>
array_view(array_view<U> const& other)
  : array_view(other.data(), other.size())
{}
```

oldnewthing:
Defect in the proposed constructor is that it would allow things like

```cpp
struct Base { int b; };
struct Derived : Base { int d; };

array_view<Base> b{...};
array_view<Derived> v = b; // oops
```

Can be repaired by requiring element type to be the same, minus cv-qualifiers, and for T to be at least as qualified as U.

```cpp
template <typename U, std::enable_if_t<std::is_same_v<std::remove_cv_t<T>, std::remove_cv_t<U>>, int> = 0>
array_view(array_view<U> const& other)
  : array_view(other.data(), other.size())
 // let the constructor validate that T is at least as qualified a U
{}
```

Another point of friction between `array_view<T const>` and `array_view<T>` is the inability to compare them. Can be repaired by allowing `array_view<T>` and `array_view<U>` to be compared, as long as `T` and `U` are comparable.

```cpp
    template <typename T, typename U>
    bool operator==(array_view<T> const& left, array_view<U> const& right) noexcept
    {
        return std::equal(left.begin(), left.end(), right.begin(), right.end());
    }
   // similarly with the other comparisons
```

This takes advantage of the fact that `std::equal` and `std::lexicographical_compare` support comparisons of ranges with different element types.

DefaultRyan:
You read my mind on the conversion issues. :) I agree, the comparison friction would be great to address at the same time.

PR #604, opened 2020-04-29
Allow conversion and comparison of cv-qualified array_view

DefaultRyan:
This fixes issue #480.

### DIFF
--- a/test/old_tests/UnitTests/array.cpp
+++ b/test/old_tests/UnitTests/array.cpp
@@ -57,6 +57,8 @@ TEST_CASE("array,SmsBinaryMessage")
     REQUIRE(1 == a[0]);
     REQUIRE(2 == a[1]);
     REQUIRE(3 == a[2]);
+
+    query.Thumbprint(a);
 }
 
 //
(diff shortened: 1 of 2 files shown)


test/old_tests/UnitTests/array.cpp
```diff
@@ -909,6 +911,31 @@ TEST_CASE("array_view,array")
     }
 }
 
+// Tests array_view construction for various cv-qualified conversions
+TEST_CASE("array_view,cv array_view")
```
kennykerr, line 915: Can we add a com_array to array_view test since that was the original issue, just to be safe?

test/old_tests/UnitTests/array.cpp
```diff
@@ -57,6 +57,8 @@ TEST_CASE("array,SmsBinaryMessage")
     REQUIRE(1 == a[0]);
     REQUIRE(2 == a[1]);
     REQUIRE(3 == a[2]);
+
+    query.Thumbprint(a);
 }
```
kennykerr, line 62: Oh I think that com_array to array_view test is happening here! 👍

kennykerr:
Thanks!

PR #608, opened 2020-05-05
weak_ref<T> can be compared for equality

oldnewthing:
This compares whether the weak references are copies of each other, not whether they refer to the same object.

### DIFF
--- a/strings/base_weak_ref.h
+++ b/strings/base_weak_ref.h
@@ -56,20 +56,51 @@ WINRT_EXPORT namespace winrt
             return static_cast<bool>(m_ref);
         }
 
-        bool operator==(weak_ref const& other) const noexcept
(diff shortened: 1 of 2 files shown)


oldnewthing:
Background notes:

* I've run across code that maintained vectors of weak references and wanted to delete items from that vector by value. In the absence of equality comparisons, they resorted to calling `get()` and comparing the results, Since `get()` is commonly used to obtain raw pointers from smart pointers, people didn't realize that they are comparing strengthened pointers rather than raw pointers. Comparing strengthened pointers is mostly-harmless (since all broken weak pointers are functionally equivalent), but being able to compare weak pointers directly saves a lot of effort. It also aids debugging because you won't blame leaks on the wrong weak pointer.

Issue #607, opened 2020-04-30
Use compact encoding for runtime names to reduce binary size


kennykerr:

kennykerr:
Turns out Xaml is very dependent on UTF-16 string literals and any deviation has a negative impact on code size for WinUI.

PR #610, opened 2020-05-06
utf-8 encoding of runtime names

kennykerr:
Part of #607 to store runtime names in binaries using utf-8 rather than utf-16. 

The next step is to fold common prefixes to reduce binary size even further.


kennykerr:
@DefaultRyan I tried using `u8"literals"` as you suggested but that doesn't seem to work in a way that will work with both C++17 and C++20. When compiled with C++20 those literals are `char8_t` string literals but `char8_t` is new to C++20 so I can't use it in C++17. I think I'll just leave it until we switch to C++20 and leave C++17 behind.

DefaultRyan:
An approach I've taken elsewhere is to conditionally define the char type using the `char8_t` feature test macro:

```
#ifdef __cpp_char8_t
using char_type = char8_t;
#else
using char_type = char;
#endif
```

strings/base_foundation.h
```diff
@@ -102,9 +102,9 @@ WINRT_EXPORT namespace winrt::Windows::Foundation
 
 namespace winrt::impl
 {
-    template <> inline constexpr auto& name_v<Windows::Foundation::Point> = L"Windows.Foundation.Point";
-    template <> inline constexpr auto& name_v<Windows::Foundation::Size> = L"Windows.Foundation.Size";
-    template <> inline constexpr auto& name_v<Windows::Foundation::Rect> = L"Windows.Foundation.Rect";
+    template <> inline constexpr auto& name_v<Windows::Foundation::Point> = "Windows.Foundation.Point";
```
DefaultRyan, line 105: Seems to be a few leftovers here.

kennykerr: Perhaps you're looking at an older commit.

kennykerr: Hmm, maybe I didn't push that.

YohDeadfall:
@kennykerr looks like you're doing changes from a different machine or the git config doesn't contain the mail associated with your account.

@walbourn Since all processes which share a library have separate memory they can store a flag to indicate does the current process uses UTF-8 or UTF-16. So if an app has UTF-8 strings, then there should be no conversion. Otherwise, conversion is required. Near the same way `Nt` and `Zw` functions work in Windows.

In case of WinRT where strings represented as handles and where the string lifetime is controlled by a reference counter, RAII could help: `c_str` should return a temporary struct which is implicitly convertible to `char8_t`/`char`, which does a conversion in the constructor and deallocation in the destructor if the flag reports that an app uses UTF-16. In case of UTF-8 no actions are needed.

strings/base_string.h
```diff
@@ -638,3 +638,27 @@ WINRT_EXPORT namespace winrt
         return result;
     }
 }
+
+namespace winrt::impl
+{
+    template <size_t Size>
+    hstring literal_to_hstring(std::array<char_type, Size> const& value) noexcept
+    {
+        return to_hstring(std::string_view(reinterpret_cast<char const*>(value.data()), Size - 1));
```
DefaultRyan, line 647: Recommend adding an overload of `to_hstring` for `std::u8string_view`, which will reduce developer friction ~~if~~ when they use `char8_t` strings. Then, the `reinterpret_cast` can live there, as its only purpose is to translate from `char8_t` to the `char` required by `MultiByteToWideChar`.

test/old_tests/UnitTests/constexpr.cpp
```diff
@@ -38,7 +38,7 @@ TEST_CASE("constexpr")
 
     REQUIRE(winrt::name_of<winrt_container>() == midl_container::z_get_rc_name_impl());
 
-    constexpr auto name = winrt::name_of<winrt::guid>();
+    auto name = winrt::name_of<winrt::guid>();
```
DefaultRyan, line 41: The change makes this not really a constexpr test any more. Shouldn't this test still be using `name_v` and comparing it to `u8"Guid"`?

strings/base_string.h
kennykerr: Yeah, I was hoping to use `u8string_view` but it looks like its new to C++20.

test/old_tests/UnitTests/constexpr.cpp
kennykerr: Yeah I should just remove this test.

strings/base_string.h
DefaultRyan: `u8string_view` is C++to, but `basic_string_view<winrt::impl::char_type>` always works.

PR #612, opened 2020-05-07
Set better defaults for references and project references to keep MdMerge happy.

### LINKED ISSUE #452: MDM2009 duplicate type error when using a grandchild reference
Scenario: creating a CppWinRT app "App" with two WinRT components "B" and "C" with the following dependencies/references:

- App references B and C
- B references C

B will copy C.winmd to its output directory since it's a dependency.  App will pass B and C's output directories as metadata directories to mdmerge since they are dependencies.  However, because C.winmd is present in both B and C's output directories, App's mdmerge will fail because it detects duplicate types in the two C.winmds it evaluates.

jlaanstra:
Current released versions of MdMerge don't like duplicate definitions. This PR attempts to reduce the duplication by not copying reference output to the output folder by default unless this is an application. An application is always a leaf node and its winmd should never be merged into something else anyway. 

Overrides are still possible since this is defined in the props file which is imported very early in the build process, to allows folks to change the default if needed. Unit test projects are an example where this might be needed, but there is not a good way to detect them. 

A non-C++/WinRT project won't get these defaults and those projects might still require manually setting Private=false, if it is reference by a C++/WinRT project, so this solution isn't perfect but it helps somewhat.

TODO:
- [x] Test razzle build

Fixes #452

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.props
+++ b/nuget/Microsoft.Windows.CppWinRT.props
@@ -43,6 +43,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <ProxyFileName Condition="'%(Midl.ProxyFileName)'==''">nul</ProxyFileName>
             <TypeLibraryName Condition="'%(Midl.TypeLibraryName)'==''"></TypeLibraryName>
         </Midl>
+        <ProjectReference Condition="'$(XamlLanguage)' != 'C++'">
+            <!-- By default, for a C++/WinRT project,
+                 don't copy project reference output to the output directory
+                 of the current project unless this is an exe. -->
+            <Private>false</Private>
+            <Private Condition="'$(ConfigurationType)' == 'Application' or '$(OutputType)'=='winexe' or '$(OutputType)'=='exe' or '$(OutputType)'=='appcontainerexe'">true</Private>
+        </ProjectReference>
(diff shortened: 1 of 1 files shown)


nuget/Microsoft.Windows.CppWinRT.props
```diff
@@ -43,6 +43,20 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <ProxyFileName Condition="'%(Midl.ProxyFileName)'==''">nul</ProxyFileName>
             <TypeLibraryName Condition="'%(Midl.TypeLibraryName)'==''"></TypeLibraryName>
         </Midl>
+        <ProjectReference Condition="'$(XamlLanguage)' != 'C++'">
```
Scottj1s, line 46: these are defaults.  if necessary, users could explicitly override in the IDE?

jlaanstra: Correct. Overrides are possible per reference, for all references or conditionally.

kennykerr:
I like the "to keep MdMerge happy" part. 😉

ChrisGuzak:
> leave node

leaf node

kennykerr:
What's the status of this PR?

jlaanstra:
@kennykerr I need to test the razzle build. Will try to do that tomorrow.

PR #615, opened 2020-05-11
Use canonical error message format for standard build tool parsing

Scottj1s:


cppwinrt/main.cpp
```diff
@@ -363,7 +363,7 @@ Where <spec> is one or more of:
         }
         catch (std::exception const& e)
         {
-            w.write(" error: %\n", e.what());
+            w.write("cppwinrt : error %\n", e.what());
```
kennykerr, line 366: Is this to avoid the `cppwinrt.cmd` in Razzle?

PR #632, opened 2020-05-20
Add SDK to NuGet package

### LINKED ISSUE #519: Create an MSBuild SDK for cpp/winrt

wjk:
This is a crude, but effective, workaround for the lack of `PackageReference` support in Visual C++ projects. If this PR is merged, users can add the below code to their vcxproj file (or to imported props/targets), and MSBuild will download and cache the specified version of C++/WinRT automatically during build or when loading the project in VS. After that, the package will behave exactly as if it was referenced using packages.config.

Since C++/WinRT is not applicable to non-C++ projects, I added a warning if the SDK is imported in any other type of project. This warning can be suppressed, which makes referencing the SDK from props/targets files in a mixed-language repository easier.

```xml
<!-- At top of project, or in .props file -->
<Import Project="Sdk.props" Sdk="Microsoft.Windows.CppWinRT" Version="..." />

<!-- At bottom of project, or in .targets file -->
<Import Project="Sdk.targets" Sdk="Microsoft.Windows.CppWinRT" Version="..." />
```

Fixes #519.

### DIFF
--- a/nuget/Sdk/Sdk.props
+++ b/nuget/Sdk/Sdk.props
@@ -1,3 +1,7 @@
 <Project>
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.vcxproj'>
+    <CppWinRTDisableAutoNuGetReference>true</CppWinRTDisableAutoNuGetReference>
+  </PropertyGroup>
+
   <Import Project="..\build\native\Microsoft.Windows.CppWinRT.props" Condition="'$(MSBuildProjectExtension)' == '.vcxproj'" />
 </Project>
--- a/nuget/Sdk/Sdk.props
+++ b/nuget/Sdk/Sdk.props
@@ -1,5 +1,5 @@
 <Project>
-  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.vcxproj'>
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.vcxproj'">
     <CppWinRTDisableAutoNuGetReference>true</CppWinRTDisableAutoNuGetReference>
   </PropertyGroup>
 
--- a/nuget/Sdk/Sdk.props
+++ b/nuget/Sdk/Sdk.props
@@ -1,0 +1,3 @@
+<Project>
+  <Import Project="..\build\native\Microsoft.Windows.CppWinRT.props" Condition="'$(MSBuildProjectExtension)' == '.vcxproj'" />
+</Project>
--- a/nuget/Microsoft.Windows.CppWinRT.nuspec
+++ b/nuget/Microsoft.Windows.CppWinRT.nuspec
@@ -25,6 +25,8 @@
     <file src="Microsoft.Windows.CppWinRT.props" target="build\native"/>
     <file src="Microsoft.Windows.CppWinRT.targets" target="build\native"/>
     <file src="CppWinrtRules.Project.xml" target="build\native"/>
+    <file src="Sdk\Sdk.props" target="Sdk" />
+    <file src="Sdk\Sdk.targets" target="Sdk" />
     <file src="readme.txt"/>
   </files>
 </package>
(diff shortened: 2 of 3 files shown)


wjk:
See also #636, which works in tandem with this PR. Thanks!

kennykerr:
Thanks William! I'll add @Scottj1s  and @jlaanstra for the vs/msbuild dark magic I know nothing about. 😉

nuget/Sdk/Sdk.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.vcxproj'>
+    <CppWinRTDisableAutoNuGetReference>true</CppWinRTDisableAutoNuGetReference>
```
jlaanstra, line 3: Remove this as it is unrelated tot eh changes in this PR.

wjk: I would like to respectfully disagree. If this PR is merged without this change applied, then we would have the following situation:

1. A user configures their vcxproj to reference C++/WinRT via NuGet SDK. This does not entail adding the NuGet package to packages.config.
2. The user adds a C++/WinRT file to the project using one of the item templates. The templates [automatically add a packages.config entry](https://github.com/microsoft/cppwinrt/blob/2f2911f8b03eaf4c8f63bab19a69639a7f0e588b/vsix/ItemTemplates/BlankUserControl/cppwinrt_BlankUserControl.vstemplate#L23-L27) when used. Now the user has two references to the C++/WinRT NuGet package (one from the NuGet SDK in this PR, and one using packages.config from the item template). This will cause problems, especially if the two reference have different versions.

PR #633 modifies the item and project templates to automatically detect if the NuGet SDK is in use, and does not add the packages.config reference automatically if that is the case. This is the property the code in that PR uses to detect if the NuGet SDK is used instead of the packages.config file.

If you would prefer, I could move this change to the other PR, but then we'd need to fully merge this PR first before I could reapply the change there, as this file will not exist to modify there until I can rebase it on top of this PR’s changes. I hope this explains things. Thanks!

jlaanstra: If that's the case we should hold off on merging this until concensus is reached on the other PR.

wjk: Very well, but I still think this PR and #633 should be reviewed in parallel. They really are two halves of one project.

kennykerr: What about merging the PRs?

PR #633, opened 2020-05-20
Add VS extension support for #632

wjk:
This PR is a redo of #536. I have since rebased it on top of master and fixed the problem described here: https://github.com/microsoft/cppwinrt/pull/536#issuecomment-589748242. The problem there was a bug in VS wherein a template adding nested files to a vcxproj would fail if the vcxproj already has filters in it. If I remove the three default filters from the project, then the templates work as expected.

### DIFF
--- a/vsix/AssemblyInfo.cs
+++ b/vsix/AssemblyInfo.cs
@@ -7,7 +7,7 @@ using System.Runtime.InteropServices;
 // General Information about an assembly is controlled through the following 
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
-[assembly: AssemblyTitle("CppWinRT Visual Studio Extension")]
+[assembly: AssemblyTitle("C++/WinRT Visual Studio Extension")]
 [assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("Microsoft Corporation")]
--- a/vsix/AssemblyInfo.cs
+++ b/vsix/AssemblyInfo.cs
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("CppWinRT Visual Studio Extension")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Microsoft Corporation")]
+
+// Version information for an assembly consists of the following four values:
+//
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0")]
+[assembly: AssemblyFileVersion("1.0")]
+[assembly: ComVisible(false)]
--- a/vsix/CppWinRTPackageTemplateWizard.cs
+++ b/vsix/CppWinRTPackageTemplateWizard.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Reflection;
 using Microsoft.Build.Evaluation;
 using Microsoft.VisualStudio.TemplateWizard;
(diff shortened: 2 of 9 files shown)


vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                ShowExceptionDialog(ex);
+            }
+        }
+
+        public void ProjectFinishedGenerating(DTEProject project)
+        {
+            try
+            {
+                VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
```
YohDeadfall, line 41: Is it correct to throw an exception in a try-catch block in a VS extension?

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+
+                Project msbuildProject = new Project(project.FullName);
+                if (msbuildProject.GetPropertyValue("CppWinRTDisableAutoNuGetReference").Equals("true", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Don't forward this call to the NuGet wizard, as it has been explicitly disabled.
+                }
+                else
+                {
+                    wizardImpl?.ProjectFinishedGenerating(project);
+                }
```
YohDeadfall, line 51:
```suggestion
                // Forward the call to the NuGet wizard, as it hasn't been explicitly disabled.
                if (msbuildProject
                    .GetPropertyValue("CppWinRTDisableAutoNuGetReference")
                    .Equals("false", StringComparison.OrdinalIgnoreCase))
                {
                    wizardImpl?.ProjectFinishedGenerating(project);
                }
```

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            catch (Exception ex)
+            {
+                ShowExceptionDialog(ex);
+            }
+        }
+
+        public void ProjectItemFinishedGenerating(DTEProjectItem projectItem)
+        {
+            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
```
YohDeadfall, line 63: And here you're not invoking the method in a try-catch block, so let's determine which way is right (:

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+                Project msbuildProject = new Project(projectItem.ContainingProject.FullName);
+                if (msbuildProject.GetPropertyValue("CppWinRTDisableAutoNuGetReference").Equals("true", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Don't forward this call to the NuGet wizard, as it has been explicitly disabled.
+                    // NuGet packages are also installed in ProjectItemFinishedGenerating().
+                }
+                else
+                {
+                    wizardImpl?.ProjectItemFinishedGenerating(projectItem);
+                }
```
YohDeadfall, line 76: This expression can be simplified as the previous one.

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        public void ProjectFinishedGenerating(DTEProject project)
+        {
+            try
+            {
+                VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+
+                Project msbuildProject = new Project(project.FullName);
```
YohDeadfall, line 43: As I know `var` is preferred over an actual type if it's specified on the right side. At least, this convention is used by .NET repos.

```suggestion
                var msbuildProject = new Project(project.FullName);
```

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+
+        public void ProjectItemFinishedGenerating(DTEProjectItem projectItem)
+        {
+            VisualStudio.Shell.ThreadHelper.ThrowIfNotOnUIThread();
+
+            try
+            {
+                Project msbuildProject = new Project(projectItem.ContainingProject.FullName);
```
YohDeadfall, line 67: The same as before.

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+                ShowExceptionDialog(ex);
+            }
+        }
+
+        public bool ShouldAddProjectItem(string filePath)
+        {
+            try
+            {
+                if (wizardImpl != null) return wizardImpl.ShouldAddProjectItem(filePath);
+                else return true;
```
YohDeadfall, line 119:
```suggestion
                return wizardImpl == null || wizardImpl.ShouldAddProjectItem(filePath);
```

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            catch (Exception ex)
+            {
+                ShowExceptionDialog(ex);
+                return true; // technically not reached
```
YohDeadfall, line 124: It's better to use `throw` while it takes no effect.

vsix/AssemblyInfo.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("CppWinRT Visual Studio Extension")]
```
YohDeadfall, line 10: Maybe change this to C++/WinRT like the projects name?

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
...
+using DTEProject = EnvDTE.Project;
+using DTEProjectItem = EnvDTE.ProjectItem;
+
+namespace Microsoft.Windows.CppWinRT
+{
+    internal sealed class CppWinRTPackageTemplateWizard : IWizard
+    {
+        private IWizard wizardImpl = null;
+
+        private void ShowExceptionDialog(Exception ex)
```
YohDeadfall, line 18: And since the method doesn`t use any instance members, it can be made `static`.

vsix/CppWinRTPackageTemplateWizard.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using Microsoft.Build.Evaluation;
+using Microsoft.VisualStudio.TemplateWizard;
+using System.Windows.Forms;
```
YohDeadfall, line 7: First part of usings are ordered, so let's put the last one to the right place.

vsix/CppWinRTPackageTemplateWizard.cs
wjk: I don't like this idea; I think this style (conditional on left, method call or `throw` on the right) is much less readable.

vsix/CppWinRTPackageTemplateWizard.cs
wjk: See my commit message on 4c0efeb.

vsix/CppWinRTPackageTemplateWizard.cs
YohDeadfall: Not a problem since both variants will be jitted identically, but let's put under if-else. This style even is used in C++ files here:

```
if (condition)
    return call();
else
    return true;
```

YohDeadfall: Or even better make `return true` the last line in the method. This will allow you to remove needless `throw` from the `catch` block.

kennykerr:
@wjk I'm going to close these PRs for now. Unfortunately we don't have anyone on the team who is able to oversee and support changes to the msbuild/vs support at the moment. However, a few people over here at Microsoft have been impressed by your contributions and would love it if you would consider [applying for a fulltime opportunity](https://careers.microsoft.com/students/us/en/job/643738/Full-Time-Opportunities-for-Students-or-Recent-Graduates-Software-Engineering-Program-Management) at Microsoft. 😉

Issue #639, opened 2020-05-22
Co_awaiting apartment_context from another STA causes RO_E_BLOCKED_CROSS_ASTA_CALL


jlaanstra:
In an application with multiple CoreDispatchers, using apartment_context as recommended in the docs can cause RO_E_BLOCKED_CROSS_ASTA_CALL when trying to switch from one dispatcher to the other.

kennykerr:
Had a chat with Johan. We're going to close this issue for now and revisit when we have more to share.

PR #640, opened 2020-05-22
When switching from STA to STA, jump through the threadpool to prevent RO_E_BLOCKED_CROSS_ASTA_CALL.

jlaanstra:
Possible fix for #639. The use of the lambda and capture of the params type is probably wrong as I am not sure if WINRT_IMPL_TrySubmitThreadpoolCallback is guaranteed to run before it returns.

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -39,14 +39,39 @@ namespace winrt::impl
         return 0;
     };
 
+    inline auto resume_context(com_ptr<IContextCallback> const& context, std::experimental::coroutine_handle<> handle)
+    {
+        com_callback_args args{};
+        args.data = handle.address();
+
+        check_hresult(context->ContextCallback(resume_apartment_callback, &args, guid_of<ICallbackWithNoReentrancyToApplicationSTA>(), 5, nullptr));
+    }
+
     inline auto resume_apartment(com_ptr<IContextCallback> const& context, std::experimental::coroutine_handle<> handle)
     {
         if (context)
         {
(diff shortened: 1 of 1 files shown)


strings/base_coroutine_threadpool.h
```diff
@@ -39,14 +39,39 @@ namespace winrt::impl
...
+            {
+                struct params
+                {
+                    com_ptr<IContextCallback> context;
+                    std::experimental::coroutine_handle<> handle;
+                };
+                params p{ context, handle };
+                if (!WINRT_IMPL_TrySubmitThreadpoolCallback([](void*, void* context) -> void
+                    {
+                        params p = *reinterpret_cast<params*>(context);
```
ChrisGuzak, line 64: `auto p = *static_cast<params*>(context);`

strings/base_coroutine_threadpool.h
```diff
@@ -39,14 +39,39 @@ namespace winrt::impl
...
-
-            check_hresult(context->ContextCallback(resume_apartment_callback, &args, guid_of<ICallbackWithNoReentrancyToApplicationSTA>(), 5, nullptr));
+            if (requires_apartment_context())
+            {
+                struct params
+                {
+                    com_ptr<IContextCallback> context;
+                    std::experimental::coroutine_handle<> handle;
+                };
+                params p{ context, handle };
```
ChrisGuzak, line 61: the lifetime of this needs to be managed, passing a shared_ptr or transferring ownership of a unique_ptr, as I don't see anything that forces the thread function to be synchronized with this.

PR #645, opened 2020-05-29
Edge cases in when_all

oldnewthing:
* Fix the case where you call `when_all` with coroutines whose return values have an overloaded comma operator. The existing code would accidentally comma the coroutine results together instead of discarding them as intended. Fix by explicitly comma-ing with void. You cannot overload   the comma operator when void is one of the parameters, so this forces the built-in comma operator to be used.

* Fix the case where you call `when_all` with no parameter. The existing code would end up with an empty body, which   lacked any `co_await` or `co_return` statements and therefore   wasn't a coroutine any more. Fix by adding a `co_return`.

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -712,7 +712,8 @@ WINRT_EXPORT namespace winrt
     template <typename... T>
     Windows::Foundation::IAsyncAction when_all(T... async)
     {
-        (co_await async, ...);
+        ((co_await async, void()), ...);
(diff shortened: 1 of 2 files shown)


oldnewthing:
D'oh, I found an easier solution this morning:
```cpp
        (void(co_await async), ...);
```

but the version I originally submitted still works, so I guess not worth tweaking it now.

kennykerr:
Always open to improvements that make it simpler. 👍

Issue #651, opened 2020-06-04
Add the ability for static lifetime store to be explicitly cleared


ChrisGuzak:
in the case of a Win32 service that uses the static lifetime store (using [`static_lifetime`](https://docs.microsoft.com/en-us/uwp/cpp-ref-for-winrt/static-lifetime)) we need to respond to the service shutdown request and un-register our factories. to do this we need to be able to clear the entry in the static lifetime store

kennykerr:
The static lifetime store is designed to clear itself when the last apartment is torn down. Is this not working?

oldnewthing:
The issue here is a shared service. One service needs to clean up all its COM objects, but other services are still using COM. I am working on a PR to address this specific scenario.

PR #654, opened 2020-06-06
Can clear a specific factory from static lifetime store

oldnewthing:
Services need to clean up all their COM objects without running down COM. Provide a way to remove a specific factory or factories from the static lifetime store.

```cpp
clear_factory_static_lifetime<Class1, Class2, Class3>();
```

The classes must be factory classes (in the `factory_implementation` namespace), to match `make_self`. We use the `instance_class` subtype to verify this.

This fixes issue #651


ChrisGuzak:
thank you Raymond.

PR #655, opened 2020-06-06
Fix ABI conformance of IMap::Remove, add TryRemove

oldnewthing:
`single_threaded_map()`'s IMap::Remove did not throw `hresult_out_of_range` on attempts to remove a nonexistent key. Now it throws.

Note that this is a breaking change. Code that assumed nonexistent objects could be harmlessly removed will encounter exceptions when run against C++/WinRT implementations when they hadn't before this change. This was, however, a pre-existing bug, because implementations from other projections (C#, C++/CX, WRL) always throw under those conditions.

Added a TryRemove() method for people who want the nonthrowing version.

Note that fixing the ABI conformance is required in order for TryRemove to return the correct result, because TryRemove relies on the call to Remove failing if the key doesn't exist.

(JavaScript doesn't project objects as maps, so there is nothing to validate there.)

Tightened the behavior of TryLookup and TryRemove so they swallow only "key not found" and propagate other errors. The inability to remove the item could be due to the server being unavailable, or the app accessing the map from the wrong thread, and that's not the same as the item not existing in the collection. Previous code treated these errors the same as "The item doesn't exist", which is not true: The item could exist, we simply don't know since the call never got off the ground.

Failure to remove a nonexistent key no longer invalidates iterators, because nothing actually changed. This brings the map implementation in line with the implementations in C# and WRL. (C++/CX invalidates prematurely. JavaScript never invalidates iterators.)

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1305,7 +1305,7 @@ namespace cppwinrt
         else if (type_name == "Windows.Foundation.Collections.IMapView`2")
         {
             w.write(R"(
-        auto TryLookup(param_type<K> const& key) const noexcept
+        auto TryLookup(param_type<K> const& key) const
         {
             if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, V>)
             {
@@ -1331,7 +1331,7 @@ namespace cppwinrt
         else if (type_name == "Windows.Foundation.Collections.IMap`2")
         {
             w.write(R"(
-        auto TryLookup(param_type<K> const& key) const noexcept
+        auto TryLookup(param_type<K> const& key) const
         {
             if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, V>)
             {
@@ -1353,7 +1353,7 @@ namespace cppwinrt
             }
         }
 
-        auto TryRemove(param_type<K> const& key) const noexcept
+        auto TryRemove(param_type<K> const& key) const
         {
             return 0 == impl::check_hresult_allow_bounds(WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMap<K, V>)->Remove(get_abi(key)));
         }
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1310,7 +1310,7 @@ namespace cppwinrt
             if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, V>)
             {
                 V result{ nullptr };
-                WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMapView<K, V>)->Lookup(get_abi(key), put_abi(result));
+                impl::check_hresult_allow_bounds(WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMapView<K, V>)->Lookup(get_abi(key), put_abi(result)));
                 return result;
             }
             else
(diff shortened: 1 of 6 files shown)


cppwinrt/code_writers.h
```diff
@@ -1310,15 +1310,15 @@ namespace cppwinrt
             if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, V>)
             {
                 V result{ nullptr };
-                WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMapView<K, V>)->Lookup(get_abi(key), put_abi(result));
+                impl::check_hresult_allow_bounds(WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMapView<K, V>)->Lookup(get_abi(key), put_abi(result)));
```
kennykerr, line 1313: This throws and yet `TryLookup` has always been `noexcept`.

cppwinrt/code_writers.h
```diff
@@ -1336,22 +1336,27 @@ namespace cppwinrt
...
                     result = std::move(value);
                 }
 
                 return result;
             }
         }
+
+        auto TryRemove(param_type<K> const& key) const noexcept
+        {
+            return 0 == impl::check_hresult_allow_bounds(WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMap<K, V>)->Remove(get_abi(key)));
```
kennykerr, line 1358: Same here - throwing inside a `noexcept` functions will terminate. Is that what you're going for?

strings/base_collections_base.h
```diff
@@ -415,8 +414,14 @@ WINRT_EXPORT namespace winrt
 
         void Remove(K const& key)
         {
+            auto& container = static_cast<D&>(*this).get_container();
+            auto found = container.find(static_cast<D const&>(*this).wrap_value(key));
+            if (found == container.end())
+            {
+                throw hresult_out_of_bounds();
```
kennykerr, line 421: Is there a reason why it's important that Remove throws when the key is not found, beyond consistency with other implementations?

cppwinrt/code_writers.h
oldnewthing: Good catch. I need to remove the noexcept from here and TryRemove. The point is that there are three outcomes:

1. Item is definitely present (and was looked up / removed)
2. Item is definitely missing (and was not looked up / not removed)
3. Operation was unsuccessful

Right now, cases 2 and 3 are conflated, which means that you have bugs where the remote server is unavailable (maybe it is hung or broken into the debugger and you got `RPC_E_TIMEOUT`), and you do a TryLookup, see that it's not there, and assume "Well, then that means it was never added, so I don't need to clean up my client tracking data." But it actually means "I couldn't perform the desired operation at all. The thing you added might or might not be there, we can't tell."

strings/base_collections_base.h
oldnewthing: It prevents `TryRemove` from working correctly.

kennykerr: Yes, but it sounds like a circular argument. `TryRemove` needs `Remove` to change. `TryRemove` is needed because `Remove` has changed. 😉

oldnewthing: TryRemove is needed because if the map came from somebody other than C++/WinRT, **it will throw**.

kennykerr: Got it - thanks.

PR #662, opened 2020-06-13
Resuming neutral context from STA should force background thread

oldnewthing:
The neutral apartment is not a thread, but rather steals whatever thread you happen to be using by temporarily converting it to a neutral apartment, and then restoring the original apartment when it's done.

COM will secretly create neutral apartments for issuing callbacks, and code may find itself inadvertently capturing a neutral apartment when it thought it was capturing an MTA:

```cpp
fire_and_forget OnSomething()
{
    auto previous = apartment_context();
    co_await resume_foreground(dispatcher);
    .do_ui_stuff();
    co_await previous;
    do_slow_stuff();
}
```

If this code is called from a neutral apartment on a borrowed MTA thread, the final `co_await previous` will switch to the neutral apartment while staying on the UI thread, even though the code started on an MTA thread. This is surprising behavior that has been identified as the root cause of some hangs.

This change revises C++/WinRT to minimize the surprise by resuming neutral apartments on an MTA thread if coming from an STA thread.

Previously, the `apartment_context` worried about only two causes (STA and MTA), and it encoded the MTA as a null pointer. Now, we have three cases to worry about, so the trick of encoding the MTA as `nullptr` is not going to work. The `apartment_context` now needs to capture both the `IContextCallback` as well as the apartment type of that context.

Explicitly carrying the apartment type removes the need for the special `nullptr` sentinel value, so we get rid of it.

**Switching to a neutral apartment from an STA now switches explicitly to an MTA thread before converting the MTA thread to a neutral apartment.** In other words, what used to be

```cpp
co_await resume_on_nta;
```

is now, conceptually,

```cpp
co_await resume_background();
co_await resume_on_nta;
```

Implementing this "extra `co_await`" is not as pretty as it sounds because we cannot use the coroutine machinery to do the extra `co_await`. We have to code it up explicitly.

New internal function `get_apartment_type()` returns a `std::pair` of the apartment type and apartment type qualifier. If COM is not initialized, then we pretend that we are in the implicit MTA.

Internal function `impl::is_sta()` is renamed to `impl::is_sta_thread()` since it now checks the apartment of the underlying thread, even if the current apartment is neutral.

Removed dependency of unit test on now-defunct internal function `impl::is_sta()`.

This PR incorporates #663 because I need `try_capture`.

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -1,6 +1,14 @@
 
 namespace winrt::impl
 {
+    inline auto submit_threadpool_callback(void(__stdcall* callback)(void*, void* context), void* context)
+    {
+        if (!WINRT_IMPL_TrySubmitThreadpoolCallback(callback, context, nullptr))
+        {
+            throw_last_error();
+        }
+    }
+
     inline void __stdcall resume_background_callback(void*, void* context) noexcept
     {
         std::experimental::coroutine_handle<>::from_address(context)();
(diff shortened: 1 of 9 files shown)


strings/base_coroutine_threadpool.h
```diff
@@ -1,63 +1,120 @@
...
+        }
+        return false;
     }
 
-    inline auto apartment_context()
+    struct resume_apartment_context
     {
-        return requires_apartment_context() ? capture<IContextCallback>(WINRT_IMPL_CoGetObjectContext) : nullptr;
-    }
+        com_ptr<IContextCallback> m_context = capture<IContextCallback>(WINRT_IMPL_CoGetObjectContext);
```
oldnewthing, line 53: This is wrong. It will throw if COM is not initialized. Fix coming.

oldnewthing:
The revised `resume_apartment` avoids going through `IContextCallback` in the common case where the `resume_apartment` is used to get from the completion context to a matching awaiter context (e.g., coroutine completed on STA, and awaiter wants to resume on STA). This conserves some stack space.

ChrisGuzak:
> Switching to a neutral apartment from an STA now switches explicitly to an MTA thread before converting the MTA thread to a neutral apartment.
This is the crux, avoiding the problem of converting an STA to an NA. Emphasize this point.

strings/base_coroutine_threadpool.h
```diff
@@ -1,63 +1,108 @@
...
     {
-        if (!WINRT_IMPL_TrySubmitThreadpoolCallback(resume_background_callback, handle.address(), nullptr))
-        {
-            throw_last_error();
-        }
+        submit_threadpool_callback(resume_background_callback, handle.address());
     }
 
-    inline bool is_sta() noexcept
+    inline std::pair<int32_t, int32_t> get_apartment_type() noexcept
```
ChrisGuzak, line 22: consider mentioning the results hold `APTTYPE_*` and `APTTYPEQUALIFIER_*` values

oldnewthing: Already mentioned in the PR description.

test/old_tests/UnitTests/apartment_context.cpp
```diff
@@ -12,9 +14,56 @@ namespace
 
         co_await context;
     }
+
+    template<typename TLambda>
+    void InvokeInContext(IContextCallback* context, TLambda&& lambda)
```
kennykerr, line 19: invoke_in_context

kennykerr: Oh, this is a test. Never mind.

PR #663, opened 2020-06-13
try_capture and try_create_instance

oldnewthing:
Nonthrowing versions of `capture` and `create_instance`.


oldnewthing:
This will be used in #662

Issue #665, opened 2020-06-18
winrt::array_view lacks a public (pointer, size) constructor


dfields-msft:
In my experience, I typically have a pointer and size in hand, rather than begin/end pair, and have to repeat the pointer in the second argument and manually add the size to it to emulate begin/end semantics, just for array_view to subtract these again to pass to the private constructor. I always wish I could just say (pointer, size) instead. Concretely:

```cpp
void ArrayViewExample(uint8_t* buffer, uint32_t size)
{
    for (const auto& byte : winrt::array_view(buffer, buffer + size)) { /* ... */ }
 
    // vs
 
    for (const auto& byte : winrt::array_view(buffer, size)) { /* ... */ }
}
```

PR #666, opened 2020-06-18
Make array_view(pointer, size) constructor public

dfields-msft:
Addressing #665 by promoting the `array_view(pointer, size)` constructor to be public rather than protected, which is consistent with C++20's `std::span` that `winrt::array_view` is designed to emulate.

### DIFF
--- a/strings/base_array.h
+++ b/strings/base_array.h
@@ -17,6 +17,11 @@ WINRT_EXPORT namespace winrt
 
         array_view() noexcept = default;
 
+        array_view(pointer data, size_type size) noexcept :
+            m_data(data),
+            m_size(size)
+        {}
+
         array_view(pointer first, pointer last) noexcept :
             m_data(first),
             m_size(static_cast<size_type>(last - first))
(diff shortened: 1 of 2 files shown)


oldnewthing:
Please add a unit test and verify that the deduction guide works.

dfields-msft:
@oldnewthing please let me know if the test I added is covering what you had in mind re: deduction guides.

dfields-msft:
@kennykerr this formulation requires callers to explicitly cast to `size_type` which is done implicitly for all other constructors. Fixing this would be possible, but would require a bit of extra template goop and I know that you're sensitive to the overhead of this. Do you have an opinion on what the contract of this constructor should be?

oldnewthing:
I was thinking of adding a new case to the `com_array,ctad` test:

```cpp
    uint8_t* p = &arc[0];
    REQUIRE_DEDUCED_AS(uint8_t, { p, 3 });
```

to verify that

```cpp
com_array a{ p, 3 };
```

deduces as `com_array<uint8_t>`.

dfields-msft:
@oldnewthing now I see what you're talking about; done.

PR #675, opened 2020-06-30
Fix concat_hstring for 0-length string

oldnewthing:
The zero-length string is represented by nullptr, not a heap-allocated string of length zero.

### DIFF
--- a/strings/base_string.h
+++ b/strings/base_string.h
@@ -423,13 +423,13 @@ namespace winrt::impl
         hstring_builder& operator=(hstring_builder const&) = delete;
 
         explicit hstring_builder(uint32_t const size) :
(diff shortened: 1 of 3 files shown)


strings/base_string.h
```diff
@@ -423,13 +423,13 @@ namespace winrt::impl
         hstring_builder& operator=(hstring_builder const&) = delete;
 
         explicit hstring_builder(uint32_t const size) :
-            m_handle(impl::precreate_hstring_on_heap(size))
+            m_handle(size ? impl::precreate_hstring_on_heap(size) : nullptr)
```
kennykerr, line 426: It doesn't make sense to create an empty `hstring_builder` - it's internal so I'm ok with this for now but it would seem to invite misuse. This class is also only used by the `concat_hstring` internal helper so it would seem like a better fix would not be too difficult.

PR #678, opened 2020-07-03
[ARM64] Fix CFG helper calling convention and simplify flow of call target address.

pmsjt:
Unlike other architectures, on ARM64 __guard_check_icall_fptr follows a custom calling convention. X15 takes the address of the target method and that value is preserved across the call.

As a result, there is no need to enlist the help of the x19 non-volatile register. In turn, no need to spill/fill it.


kennykerr:
Thanks for the contribution! 🎉

PR #681, opened 2020-07-05
Allow directly comparing references to a value

sylveon:
This makes it more natural to compare references to a certain value, brings the projection more in line with C# behavior (`int?` can be compared to `int` over there), and helps avoiding null pointer reads (less inclined users might write `ref.Value() == 10` which will crash if ref is null).


test/old_tests/UnitTests/References.cpp
```diff
@@ -27,4 +27,18 @@ TEST_CASE("References")
...
+
+    REQUIRE(a != 124);
+    REQUIRE(124 != a);
+
+    // Tests short circuit when reference is empty.
+    REQUIRE_FALSE(e == 123);
+    REQUIRE_FALSE(123 == e);
+
+    REQUIRE(e != 123);
+    REQUIRE(123 != e);
```
kennykerr, line 43: Can you add tests for strings? They have a way of weeding out issues.

sylveon: Trying to use strings will "box" here because the raw string will be converted to `hstring` beforehand (because of `operator==(IReference<hstring>, hstring)`)... Maybe add `param::hstring` overloads and independent tests for those cases?

kennykerr: Yep, it gets expensive real quick. `param::hstring` should help. Look at how `box_value` is overloaded for strings.

sylveon: Another idea would be to make the non-reference parameter generic and letting `hstring`'s == overload handle it:

```cpp
template <typename T, typename U>
bool operator==(IReference<T> const& left, U const& right)
{
    if (!left)
    {
        return false;
    }

    return left.Value() == right;
}
```

kennykerr: It's worth a try. 😉

sylveon: After trying it seems that this breaks comparing an IReference with nullptr so I'll add hstring overloads

sylveon:
@kennykerr I added some `IReference<hstring>` overloads (using `std::wstring_view` since I realized `param::hstring` still called `WindowsCreateStringReference` when we don't need it). Also made it possible to compare references to different types (so that `IReference<int>` and `IReference<unsigned long long>` can be tested for equality).

Sidenote, but when testing those new changes I used incremental building from `build_test_all.cmd` and reliably got segmentation faults or test failures in unrelated tests until I cleaned the `_build` folder... Are incremental builds broken?

test/old_tests/UnitTests/References.cpp
```diff
@@ -27,4 +28,44 @@ TEST_CASE("References")
...
+
+    REQUIRE(str != L"bar");
+    REQUIRE(L"bar" != str);
+
+    // Tests short circuit when reference is empty.
+    REQUIRE_FALSE(nullstr == L"foo");
+    REQUIRE_FALSE(L"foo" == nullstr);
+
+    REQUIRE(nullstr != L"bar");
+    REQUIRE(L"bar" != nullstr);
```
DefaultRyan, line 70: Test these combinations also:

- IReference<hstring> and hstring
- IReference<hstring> and std::wstring
- IReference<hstring> and std::wstring_view
- IReference<hstring> and char[] (non-literal)
- IReference<hstring> and char* (non-literal)

test/old_tests/UnitTests/References.cpp
```diff
@@ -27,4 +28,44 @@ TEST_CASE("References")
...
+
+    REQUIRE(a != 124);
+    REQUIRE(124 != a);
+
+    // Tests short circuit when reference is empty.
+    REQUIRE_FALSE(e == 123);
+    REQUIRE_FALSE(123 == e);
+
+    REQUIRE(e != 123);
+    REQUIRE(123 != e);
```
DefaultRyan, line 50: Let's also add tests for floats, doubles, and bools. Testing against bool in particular will also be good for weeding out any surprises, most likely around types that are convertible to bool. It makes me wonder if we are going to inadvertently allow things like this to compile:
```
IReference<bool> == L"foo";
IReference<bool> == 42;
IReference<bool> == IInspectable{};
```
etc

oldnewthing: It also means that if `foo` is an `IReference<bool>`, then

```cpp
if (foo != false) ...
if (foo) ...
```
are not the same thing, which could be the source of subtle refactoring bugs where somebody takes the rule "Don't compare a boolean against true or false, just test the boolean" and inadvertently applies it to the first version to "simplify" it to the second version, and ends up breaking it.

kennykerr:
Regarding incremental builds, see [this](https://github.com/microsoft/cppwinrt#building-cwinrt) for how we generally build and test changes. Incremental builds don't really work as you have to rebuild the projection, which invalidates pretty much everything.

> compare references to different types

Why is that useful? Just a little worried about adding complexity without compelling justification.

test/old_tests/UnitTests/References.cpp
kennykerr: Yes, we need to be very careful not to introduce a breaking change here.

sylveon:
I was thinking "what if I wanted to compare `IReference<int>` and `IReference<short>`?" Today I'd have to write `a && b && a.Value() == b.Value()` so I thought I could shorten that.

test/old_tests/UnitTests/References.cpp
sylveon: Raymond raises a valid point, and you can see this in how C# handles a `bool?`: you can't test those just by doing `if (foo)`, it results in a compilation error, you have to write `if (foo == true)`, making it effectively a tri-state bool. Doing this in C++ would be possible but it would obviously be source breaking.

sylveon: FWIW I think `if (foo)` is already misleading because it only tests nullability, not if it's actually true.

oldnewthing: Other things that would be not be equivalent after this change:

* `if (!v)` and `if (v == false)`
* `if (v)` and `if (v != 0)`

The case of `v != 0` is problematic because `0` is implicitly convertible to `nullptr`, so a test like `if (v != 0)` does different things depending on whether `v` is `IReference<int>` or `IReference<not-int>`.

I admire the intent behind these convenience operators, but I think they open too big of a can of worms.

kennykerr: > I admire the intent behind these convenience operators, but I think they open too big of a can of worms.

Agreed, C++ lets you express some amazing things but it can often get you into trouble. In this case, we have a *lot* of code that has to keep running on C++/WinRT and I'm just not sure the convenience here is worth the risk.

kennykerr:
> I was thinking "what if I wanted to compare IReference<int> and IReference<short>?" Today I'd have to write a && b && a.Value() == b.Value() so I thought I could shorten that.

I think you might be better served by some helper functions for comparing values. These don't even need to be part of C++/WinRT.

test/old_tests/UnitTests/References.cpp
sylveon: Right, and it gets worse with the `NULL` macro being actually defined to 0 so if somebody uses that to check for null with an `IReference<int>`, this would actually be breaking. If I really want this feature, I can implement those operators manually in my code.

I still think there might be value in merging the `std::wstring_view` overloads and ability to compare different types however, provided I address @DefaultRyan's comment.

sylveon:
I'll remove that and the comparing values part, but keep the `std::wstring_view` overloads and add more test coverage for those then.

test/old_tests/UnitTests/References.cpp
DefaultRyan: I think I'm with Raymond and Kenny here. Adding overloads for `operator==` to pre-existing types is a change that is fundamentally invasive to existing code - code that was testing/comparing `IReference<T>` may now acquire very different semantics, because `operator==` is fundamental to the language and unavoidable.

Adding a new helper function (and to a lesser degree, modifying an existing helper) is far less invasive, because it allows for more "opt in" on the developers' part.

But in either case, the benefit of this change seems relatively meager. All it enables is replacing this
```
if (ref && ref.Value() == value)
```
with this
```
if (ref == value)
```

The old pattern isn't onerous, hostile, or overly verbose, so the convenience almost borders on syntactic sugar. Like Ramond, I admire the intent, but the edge cases are a bit too subtle and potentially breaking in my opinion.

oldnewthing: This might be more suited to a helper library like [`wil\include\cppwinrt.h`](https://github.com/microsoft/wil/blob/master/include/wil/cppwinrt.h).

```cpp
namespace wil
{
   template<typename T, typename T>
    bool nullable_value_equal(IReference<T> const& left, T const& right)
    {
        return left && left.Value() == right;
    }

   // etc.
}
```

Then people could opt into it explicitly:

```cpp
  if (wil::nullable_value_equal(foo, true)) { ... }
```

PR #687, opened 2020-07-10
Accommodate other "key not found" errors

### LINKED ISSUE #685: Application.Current.Resources.TryLookup throws hresult_error despite it being a Try* method
In the cppwinrt generated header we have this in windows.foundation.collections.0.h, and we go down the first branch in the code, which will throw if the item is not found (in this case, XAML's ResourceDictionary returns `E_FAIL`):

```cpp
        auto TryLookup(param_type<K> const& key) const
        {
            if constexpr (std::is_base_of_v<Windows::Foundation::IUnknown, V>)
            {
                V result{ nullptr };
                impl::check_hresult_allow_bounds(WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMap<K, V>)->Lookup(get_abi(key), put_abi(result)));
                return result;
            }
            else
            {
                std::optional<V> result;
                V value{ empty_value<V>() };

                if (0 == impl::check_hresult_allow_bounds(WINRT_IMPL_SHIM(Windows::Foundation::Collections::IMap<K, V>)->Lookup(get_abi(key), put_abi(value))))
                {
                    result = std::move(value);
                }

                return result;
            }
        }
```

oldnewthing:
XAML's ResourceDictionary has a custom implementation of `IMap` that returns `E_FAIL` and `HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)` to indicate that the key is not present. Accommodate this nonstandard behavior.

Fixes #685

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -584,7 +584,7 @@ namespace winrt::impl
 {
     inline hresult check_hresult_allow_bounds(hresult const result)
     {
-        if (result != impl::error_out_of_bounds)
+        if (result != impl::error_out_of_bounds && result != impl::error_fail && result != impl::error_file_not_found)
         {
             check_hresult(result);
         }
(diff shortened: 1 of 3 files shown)


strings/base_error.h
```diff
@@ -584,7 +584,7 @@ namespace winrt::impl
 {
     inline hresult check_hresult_allow_bounds(hresult const result)
     {
-        if (result != impl::error_out_of_bounds)
+        if (result != impl::error_out_of_bounds && result != impl::error_fail && result != impl::error_file_not_found)
```
kennykerr, line 587: Wish there was some way to incentivize the implementations to clean up their act as this isn't sustainable and they're not likely to fix this on their end.

jonwis: It's a contract breaking change now if we "fix" the XAML implementation.

PR #688, opened 2020-07-10
Add LookupAs/TryLookupAs to complement Lookup/TryLookup

### LINKED ISSUE #427: Maps (propertysets) of K to Object could get an unboxing [Try]Lookup
There's a bunch of code like this:

```c++
ValueSet v = /* ... */
auto g = winrt::unbox_value<winrt::guid>(v.Lookup(L"someKey"));
auto q = winrt::unbox_value_or<winrt::hstring>(v.TryLookup(L"anotherKey"), winrt::hstring{});
```

It'd be neat if 'unbox object to type' was directly supported so I could have written:

```c++
ValueSet v = /* ... */
auto g = v.Lookup<winrt::guid>(L"someKey");
auto q = v.TryLookup<winrt::hstring>(L"anotherKey");
```

kennykerr:
Fixes #427 by adding unboxing helpers for maps.

### DIFF
--- a/strings/base_reference_produce.h
+++ b/strings/base_reference_produce.h
@@ -378,5 +378,55 @@ WINRT_EXPORT namespace winrt
     }
 
     template <typename T>
+    auto try_unbox_value(Windows::Foundation::IInspectable const& value)
+    {
+        if constexpr (std::is_base_of_v<Windows::Foundation::IInspectable, T>)
+        {
+            return value.try_as<T>();
+        }
+        else
+        {
+            if (!value)
+            {
+                return std::optional<T>();
+            }
+
+            if constexpr (std::is_enum_v<T>)
+            {
+                if (auto temp = value.try_as<Windows::Foundation::IReference<T>>())
+                {
+                    return std::optional<T>(temp.Value());
+                }
+
+                if (auto temp = value.try_as<Windows::Foundation::IReference<std::underlying_type_t<T>>>())
+                {
+                    return std::optional<T>(static_cast<T>(temp.Value()));
+                }
+
+                return std::optional<T>();
+            }
+#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
+            else if constexpr (std::is_same_v<T, GUID>)
+            {
+                if (auto temp = value.try_as<Windows::Foundation::IReference<guid>>())
+                {
+                    return std::optional<T>(temp.Value());
+                }
+
+                return std::optional<T>();
+            }
+#endif
+            else if (auto temp = value.try_as<Windows::Foundation::IReference<T>>())
+            {
+                return std::optional<T>(temp.Value());
+            }
+            else
+            {
+                return std::optional<T>();
+            }
+        }
+    }
+
+    template <typename T>
     using optional = Windows::Foundation::IReference<T>;
 }
--- a/test/test/lookup_as.cpp
+++ b/test/test/lookup_as.cpp
@@ -4,68 +4,77 @@ using namespace winrt;
 using namespace Windows::Foundation;
 using namespace Windows::Foundation::Collections;
 
+namespace
+{
+    template <typename M>
+    void test(M const& m)
+    {
+        // Good LookupAs
+
+        REQUIRE(m.LookupAs<Uri>(L"uri").Domain() == L"kennykerr.ca");
+        REQUIRE(m.LookupAs<IStringable>(L"uri").ToString() == L"http://kennykerr.ca/");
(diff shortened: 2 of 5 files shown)


strings/base_reference_produce.h
```diff
@@ -377,6 +377,56 @@ WINRT_EXPORT namespace winrt
         return default_value;
     }
 
+    template <typename T>
+    auto try_unbox_value(Windows::Foundation::IInspectable const& value)
+    {
+        if constexpr (std::is_base_of_v<Windows::Foundation::IInspectable, T>)
+        {
+            return value.try_as<T>();
```
jonwis, line 385: TryLookup returns std::optional<T>{} when not present, yes?  Does that .try_as<T> return nullptr if the value is empty? Or does the `if (!value)` below need to move upwards?

test/test/lookup_as.cpp
```diff
@@ -0,0 +1,82 @@
...
+        REQUIRE(m.TryLookupAs<IStringable>(L"uri").ToString() == L"http://kennykerr.ca/");
+        REQUIRE(m.TryLookupAs<AsyncStatus>(L"enum").value() == AsyncStatus::Completed);
+        REQUIRE(m.TryLookupAs<AsyncStatus>(L"underlying_type").value() == AsyncStatus::Error);
+        REQUIRE(m.TryLookupAs<guid>(L"guid").value() == guid_of<Uri>());
+        REQUIRE(m.TryLookupAs<GUID>(L"guid").value() == static_cast<GUID>(guid_of<Uri>()));
+        REQUIRE(m.TryLookupAs<hstring>(L"hstring").value() == L"Test");
+
+        // TryLookupAs edge cases
+
+        REQUIRE_FALSE(m.TryLookupAs<bool>(L"INVALID").has_value());
```
jonwis, line 49: Does "nullptr" here represent the invalid value? Or should there be a test for `m.TryLookupAs<Uri>(L"INVALID")` here?

strings/base_reference_produce.h
oldnewthing: On TryLookup failure, it returns `nullptr` for reference types and `std:optional{}` for value types.

test/test/lookup_as.cpp
oldnewthing: `TryLookupAs<bool>` will return a `std::optional<bool>`, and we are verifying that it returns a valueless object.

In the second test, `TryLookupAs<Uri>` will return a `Uri`, and we are using the `Uri == nullptr` test to see if the `Uri` object is empty.

mlalic:
> I preferred the proposal to add support for value types to `IInspectable.as<>` and `IInspectable.try_as<>` (treating them as unboxing), because that is useful in places other than map lookups. For example, converting XAML tags and navigation parameters.

FWIW, in addition to the Xaml examples Raymond listed, we have a new UIAutomation platform API (currently in insider preview SDK builds) that would also benefit from a uniform way of converting an `IInspectable` to a runtimeclass, COM pointer, or an unboxed value.

For example;

```c++
// today
auto elementResult = result.GetOperand(a).as<IUIAutomationElement>(); // this is a COM interface
auto textRangeResult = result.GetOperand(b).as<winrt::AutomationTextRange>(); // this is a WinRT runtimeclass
auto stringResult = winrt::unbox_value<winrt::hstring>(result.GetOperand(c));


// new hotness :)
auto elementResult = result.GetOperand(a).as<IUIAutomationElement>();
auto textRangeResult = result.GetOperand(b).as<winrt::AutomationTextRange>();
auto stringResult = result.GetOperand(c).as<winrt::hstring>();
```

The consistency looks really appealing from a consumer's perspective, IMO.

kennykerr:
Yes, I also prefer overloading `as` and `try_as` - that was my original suggestion. The trouble is (beyond the tricky metaprogramming) that this introduces a dependency from `winrt/base.h` (where `IUnknown`/`IInspectable`/`as`/`try_as` live) to `winrt/windows.foundation.h` (where boxing support lives). And that's just a much bigger change than I'd like to do for this feature.

strings/base_reference_produce.h
jonwis: Thanks, try_as of nullpter is what I didn't know.

test/test/lookup_as.cpp
jonwis: How about a test that validates `TryLookupAs<IInspectableDerivedThing>(L"VALUE THAT IS NOT IN THE MAP")`?

Issue #676, opened 2020-07-03
Implement upcoming STL C++20 coroutines support


sylveon:
The Microsoft STL has recently merged non-experimental coroutine support: https://github.com/microsoft/STL/pull/894

This support, once shipped (expected around 16.8 preview 1), will prevent mixing code using the new coroutines and the experimental ones.

Using the `/await` command line (which is forced in various places by the C++/WinRT project files) will prevent including `<coroutine>`, telling you to use `<experimental/coroutine>`. Similarly, using the `/std:c++latest` command line (my code makes use of many new C++ 20 only features), will prevent including `<experimental/coroutine>`, telling you to instead use `<coroutine>`.

This puts me in a weird spot: I can't include `<coroutine>` because C++/WinRT uses `<experimental/coroutine>` and `/await`, but neither can I include `<experimental/coroutine>` because I use `/std:c++latest`.

Ideally, the C++/WinRT headers should detect the presence of C++20 coroutines (using `__cpp_impl_coroutine`) and use the right header. Additionally, the project files should not force `/await` in VS 16.8 and newer, or provide a way to opt-out from this behavior.

kennykerr:
Yes, it might soon be time to create a C++17 branch for servicing and then switch master over to C++20. I'd like to avoid supporting both C++17 and C++20 as it can become quite complex especially since Clang and Visual C++ are never quite in line.

DefaultRyan:
It looks like you're saying VC++ in C++17 mode does not define `__cpp_impl_coroutine`. Did I read that correctly?

sylveon:
https://github.com/microsoft/STL/blob/21acd23ad81463628e95cb6c571c56ba7200cdba/stl/inc/yvals_core.h#L1162-L1164

https://github.com/microsoft/STL/blob/21acd23ad81463628e95cb6c571c56ba7200cdba/stl/inc/coroutine#L12-L21

Yes

sylveon:
It happened, VS 16.8 Preview 1 is out and my code doesn't compile anymore:

![image](https://user-images.githubusercontent.com/6440374/89701538-e481d800-d905-11ea-9efc-5adc1a24c864.png)

sylveon:
[...]

PR #692, opened 2020-07-14
Add multi-threaded collections

dunhor:
This mainly re-uses the existing collections, but adds support for doing shared/exclusive operations. For the existing single threaded collections, nothing special is done and operations are performed without any synchronization. For the new multi-threaded collections, an SRWLOCK is acquired in either shared or exclusive mode.

The most noteworthy "flaw" is that mutating iterator operations lock the collection exclusively. This is mostly a limitation imposed by the base IIterable type which uses opaque iterators for the IIterator objects. In theory, the collection only needs to be locked in shared mode, however that would require the IIterator objects themselves to have a separate lock, which seemed like overkill and unlikely to be worth it. Compare this to the existing ABI collection types which know the underlying vector/map types and can perform interlocked operations - in the case of vector - or forgo sequential consistency while still guaranteeing safety - in the case of map.

The following were also discussed, but are not addressed here as they can be done separately as a later change:
* The vector/map types still implement IVectorView/IMapView and GetView returns a pointer to itself, meaning that you can still QI an IVectorView -> IVector
* The IObservable* types still allow modification during a change callout. This can pretty easily be addressed by an atomic_bool as locks aren't really helpful here anyway.

### DIFF
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -35,6 +35,48 @@ namespace winrt::impl
 
         mutable slim_mutex m_mutex;
     };
+
+    template <typename D>
+    using container_type_t = std::decay_t<decltype(std::declval<D>().get_container())>;
+
+    template <typename D, typename = void>
+    struct removed_values
+    {
+        void assign(container_type_t<D>& value)
+        {
+            // Trivially destructible; okay to run destructors under lock and clearing allows potential re-use of buffers
+            value.clear();
+        }
+    };
+
+    template <typename D>
+    struct removed_values<D, std::enable_if_t<!std::is_trivially_destructible_v<typename container_type_t<D>::value_type>>>
+    {
+        container_type_t<D> m_value;
+
+        void assign(container_type_t<D>& value)
+        {
+            m_value.swap(value);
+        }
+    };
+
+    template <typename T, typename = void>
+    struct removed_value
+    {
+        // Trivially destructible; okay to run destructor under lock
+        void assign(T&) {}
+    };
+
+    template <typename T>
+    struct removed_value<T, std::enable_if_t<std::is_move_constructible_v<T> && !std::is_trivially_destructible_v<T>>>
+    {
+        std::optional<T> m_value;
+
+        void assign(T& value)
+        {
+            m_value.emplace(std::move(value));
+        }
+    };
 }
 
 WINRT_EXPORT namespace winrt
@@ -64,7 +106,7 @@ WINRT_EXPORT namespace winrt
         auto perform_shared(Func&& fn) const
         {
             // Support for concurrent "shared" operations is optional
-            return perform_exclusive(fn);
+            return static_cast<D const&>(*this).perform_exclusive(fn);
         }
 
         auto First()
@@ -286,6 +328,7 @@ WINRT_EXPORT namespace winrt
 
         void SetAt(uint32_t const index, T const& value)
         {
+            impl::removed_value<typename impl::container_type_t<D>::value_type> oldValue;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 if (index >= static_cast<D const&>(*this).get_container().size())
@@ -294,7 +337,9 @@ WINRT_EXPORT namespace winrt
                 }
 
                 this->increment_version();
-                static_cast<D&>(*this).get_container()[index] = static_cast<D const&>(*this).wrap_value(value);
+                auto& pos = static_cast<D&>(*this).get_container()[index];
+                oldValue.assign(pos);
+                pos = static_cast<D const&>(*this).wrap_value(value);
             });
         }
 
@@ -314,6 +359,7 @@ WINRT_EXPORT namespace winrt
 
         void RemoveAt(uint32_t const index)
         {
+            impl::removed_value<typename impl::container_type_t<D>::value_type> removedValue;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 if (index >= static_cast<D const&>(*this).get_container().size())
@@ -322,7 +368,9 @@ WINRT_EXPORT namespace winrt
                 }
 
                 this->increment_version();
-                static_cast<D&>(*this).get_container().erase(static_cast<D const&>(*this).get_container().begin() + index);
+                auto itr = static_cast<D&>(*this).get_container().begin() + index;
+                removedValue.assign(*itr);
+                static_cast<D&>(*this).get_container().erase(itr);
             });
         }
 
@@ -337,6 +385,7 @@ WINRT_EXPORT namespace winrt
 
         void RemoveAtEnd()
         {
+            impl::removed_value<typename impl::container_type_t<D>::value_type> removedValue;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 if (static_cast<D const&>(*this).get_container().empty())
@@ -345,24 +394,28 @@ WINRT_EXPORT namespace winrt
                 }
 
                 this->increment_version();
+                removedValue.assign(static_cast<D&>(*this).get_container().back());
                 static_cast<D&>(*this).get_container().pop_back();
             });
         }
 
         void Clear() noexcept
         {
+            impl::removed_values<D> oldContainer;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 this->increment_version();
-                static_cast<D&>(*this).get_container().clear();
+                oldContainer.assign(static_cast<D&>(*this).get_container());
             });
         }
 
         void ReplaceAll(array_view<T const> value)
         {
+            impl::removed_values<D> oldContainer;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 this->increment_version();
+                oldContainer.assign(static_cast<D&>(*this).get_container());
                 assign(value.begin(), value.end());
             });
         }
@@ -372,16 +425,14 @@ WINRT_EXPORT namespace winrt
         template <typename InputIt>
         void assign(InputIt first, InputIt last)
         {
-            using container_type = std::remove_reference_t<decltype(static_cast<D&>(*this).get_container())>;
-
-            if constexpr (std::is_same_v<T, typename container_type::value_type>)
+            if constexpr (std::is_same_v<T, typename impl::container_type_t<D>::value_type>)
             {
                 static_cast<D&>(*this).get_container().assign(first, last);
             }
             else
             {
                 auto& container = static_cast<D&>(*this).get_container();
-                container.clear();
+                WINRT_ASSERT(container.empty());
                 container.reserve(std::distance(first, last));
 
                 std::transform(first, last, std::back_inserter(container), [&](auto&& value)
@@ -534,16 +585,24 @@ WINRT_EXPORT namespace winrt
 
         bool Insert(K const& key, V const& value)
         {
+            impl::removed_value<typename impl::container_type_t<D>::mapped_type> oldValue;
             return static_cast<D&>(*this).perform_exclusive([&]
             {
                 this->increment_version();
-                auto pair = static_cast<D&>(*this).get_container().insert_or_assign(static_cast<D const&>(*this).wrap_value(key), static_cast<D const&>(*this).wrap_value(value));
-                return !pair.second;
+                auto [itr, added] = static_cast<D&>(*this).get_container().emplace(static_cast<D const&>(*this).wrap_value(key), static_cast<D const&>(*this).wrap_value(value));
+                if (!added)
+                {
+                    oldValue.assign(itr->second);
+                    itr->second = static_cast<D const&>(*this).wrap_value(value);
+                }
+
+                return !added;
             });
         }
 
         void Remove(K const& key)
         {
+            typename impl::container_type_t<D>::node_type removedNode;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 auto& container = static_cast<D&>(*this).get_container();
@@ -553,16 +612,17 @@ WINRT_EXPORT namespace winrt
                     throw hresult_out_of_bounds();
                 }
                 this->increment_version();
-                container.erase(found);
+                removedNode = container.extract(found);
             });
         }
 
         void Clear() noexcept
         {
+            impl::removed_values<D> oldContainer;
             static_cast<D&>(*this).perform_exclusive([&]
             {
                 this->increment_version();
-                static_cast<D&>(*this).get_container().clear();
+                oldContainer.assign(static_cast<D&>(*this).get_container());
             });
         }
     };
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -68,7 +68,6 @@ WINRT_EXPORT namespace winrt
             void abi_enter()
             {
                 m_owner->abi_enter();
-                this->check_version(*m_owner);
             }
 
             void abi_exit()
@@ -86,12 +85,56 @@ WINRT_EXPORT namespace winrt
 
             T Current() const
             {
-                // TODO: Locking and validity check?
-                if (m_current == m_end)
+                return m_owner->perform_shared([&]
                 {
-                    throw hresult_out_of_bounds();
-                }
+                    this->check_version(*m_owner);
 
+                    if (m_current == m_end)
+                    {
+                        throw hresult_out_of_bounds();
+                    }
+
+                    return current_value();
+                });
+            }
+
+            bool HasCurrent() const noexcept
+            {
+                return m_owner->perform_shared([&]
+                {
+                    this->check_version(*m_owner);
+                    return m_current != m_end;
+                });
+            }
+
+            bool MoveNext() noexcept
+            {
+                return m_owner->perform_shared([&]
+                {
+                    this->check_version(*m_owner);
+                    if (m_current != m_end)
+                    {
+                        ++m_current;
+                    }
+
+                    return m_current != m_end;
+                });
+            }
+
+            uint32_t GetMany(array_view<T> values)
+            {
+                return m_owner->perform_shared([&]
+                {
+                    this->check_version(*m_owner);
+                    return GetMany(values, typename std::iterator_traits<iterator_type>::iterator_category());
+                });
+            }
+
+        private:
+
+            T current_value() const
+            {
+                WINRT_ASSERT(m_current != m_end);
                 if constexpr (!impl::is_key_value_pair<T>::value)
                 {
                     return m_owner->unwrap_value(*m_current);
@@ -101,31 +144,6 @@ WINRT_EXPORT namespace winrt
                     return make<impl::key_value_pair<T>>(m_owner->unwrap_value(m_current->first), m_owner->unwrap_value(m_current->second));
                 }
             }
-
-            bool HasCurrent() const noexcept
-            {
-                // TODO: Validity check?
-                return m_current != m_end;
-            }
-
-            bool MoveNext() noexcept
-            {
-                // TODO: Locking and validity check?
-                if (m_current != m_end)
-                {
-                    ++m_current;
-                }
-
-                return HasCurrent();
-            }
-
-            uint32_t GetMany(array_view<T> values)
-            {
-                // TODO: Locking and validity check?
-                return GetMany(values, typename std::iterator_traits<iterator_type>::iterator_category());
-            }
-
-        private:
 
             uint32_t GetMany(array_view<T> values, std::random_access_iterator_tag)
             {
@@ -141,7 +159,7 @@ WINRT_EXPORT namespace winrt
 
                 while (output < values.end() && m_current != m_end)
                 {
-                    *output = Current();
+                    *output = current_value();
                     ++output;
                     ++m_current;
                 }
@@ -573,5 +591,26 @@ WINRT_EXPORT namespace winrt
             Windows::Foundation::Collections::CollectionChange const m_change;
             K const m_key;
         };
+    };
+
+    struct multi_threaded_collection_base
+    {
+        template <typename Func>
+        auto perform_exclusive(Func&& fn) const
+        {
+            slim_lock_guard lock(m_mutex);
+            return fn();
+        }
+
+        template <typename Func>
+        auto perform_shared(Func&& fn) const
+        {
+            slim_lock_guard_shared lock(m_mutex);
+            return fn();
+        }
+
+    private:
+
+        mutable slim_mutex m_mutex;
     };
 }
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -1,5 +1,64 @@
 WINRT_EXPORT namespace winrt
 {
+    namespace impl
+    {
+        template <typename D, typename = void>
+        struct has_exclusive_op : std::false_type {};
+
+        template <typename D>
+        struct has_exclusive_op<D, std::void_t<decltype(std::declval<D*>()->exclusive_op(std::true_type{}))>> : std::true_type {};
+
+        template <typename D, typename = void>
+        struct has_shared_op : std::false_type {};
+
+        template <typename D>
+        struct has_shared_op<D, std::void_t<decltype(std::declval<D*>()->shared_op(std::true_type{}))>> : std::true_type {};
+
+        struct single_threaded_collection_base
+        {
+#if 0
+            template <typename Func>
+            auto exclusive_op(Func&& fn) const
+            {
+                slim_lock_guard lock(m_mutex);
+                return fn();
+            }
+
+            template <typename Func>
+            auto shared_op(Func&& fn) const
+            {
+                slim_lock_guard_shared lock(m_mutex);
+                return fn();
+            }
+
+        private:
+
+            mutable slim_mutex m_mutex;
+#endif
+        };
+
+        struct multi_threaded_collection_base
+        {
+            template <typename Func>
+            auto exclusive_op(Func&& fn) const
+            {
+                slim_lock_guard lock(m_mutex);
+                return fn();
+            }
+
+            template <typename Func>
+            auto shared_op(Func&& fn) const
+            {
+                slim_lock_guard_shared lock(m_mutex);
+                return fn();
+            }
+
+        private:
+
+            mutable slim_mutex m_mutex;
+        };
+    }
+
     template <typename D, typename T, typename Version = impl::no_collection_version>
     struct iterable_base : Version
     {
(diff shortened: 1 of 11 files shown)


strings/base_collections_base.h
```diff
@@ -1,5 +1,45 @@
 WINRT_EXPORT namespace winrt
 {
+    namespace impl
```
kennykerr, line 3: Don't nest namespaces. This causes the impl namespace to be exported as part of a C++ module. Instead, start a new root namespace definition every time e.g. `namespace winrt::impl`

strings/base_collections_base.h
```diff
@@ -70,43 +140,66 @@ WINRT_EXPORT namespace winrt
...
+                    if (m_current == m_end)
+                    {
+                        throw hresult_out_of_bounds();
+                    }
+
+                    return current_value();
+                });
             }
 
-            bool HasCurrent() const noexcept
```
jonwis, line 88: Is losing noexcept here ok?  Previously this was nofailish, now there's an issue with version checking.

strings/base_collections_base.h
```diff
@@ -70,43 +140,66 @@ WINRT_EXPORT namespace winrt
...
 
-            bool MoveNext() noexcept
+            bool MoveNext()
             {
-                if (m_current != m_end)
+                return m_owner->perform_exclusive([&]
                 {
-                    ++m_current;
-                }
+                    this->check_version(*m_owner);
```
jonwis, line 169: Why is exclusive required for iterator movement? If you acquire the shared lock and verify the version info, then m_end should still be accurate, and this otherwise does not mutate the target collection.

strings/base_collections_base.h
```diff
@@ -70,43 +140,66 @@ WINRT_EXPORT namespace winrt
...
+                return m_owner->perform_exclusive([&]
+                {
+                    this->check_version(*m_owner);
+                    return GetMany(values, typename std::iterator_traits<iterator_type>::iterator_category());
+                });
             }
 
         private:
 
+            T current_value() const
```
jonwis, line 190: Should there be a naming convention for "no locks in this method"?  `current_value` seems to only be called under a lock.  (Not sure where we stand on SAL, but there's a "require no locks held" annotation.)

strings/base_collections_base.h
```diff
@@ -189,64 +301,85 @@ WINRT_EXPORT namespace winrt
...
+                this->increment_version();
+                static_cast<D&>(*this).get_container().pop_back();
+            });
         }
 
         void Clear() noexcept
         {
-            this->increment_version();
-            static_cast<D&>(*this).get_container().clear();
+            static_cast<D&>(*this).perform_exclusive([&]
```
jonwis, line 369: I get the pattern, but why call `(*this).perform_exclusive` here vs. types that return locks like `auto e = lock_exclusive()` / `auto s = lock_shared()` which return "the right type of unlocker"? Might reduce the code churn here and some compile time cost?

strings/base_collections_map.h
```diff
@@ -2,13 +2,17 @@
 namespace winrt::impl
 {
     template <typename K, typename V, typename Container>
-    struct observable_map :
```
jonwis, line 5: Should the lock type or lock instance be configurable?

strings/base_collections_base.h
```diff
@@ -1,3 +1,43 @@
...
+    struct has_exclusive_op<D, std::void_t<decltype(std::declval<D*>()->exclusive_op(std::true_type{}))>> : std::true_type {};
+
+    template <typename D, typename = void>
+    struct has_shared_op : std::false_type {};
+
+    template <typename D>
+    struct has_shared_op<D, std::void_t<decltype(std::declval<D*>()->shared_op(std::true_type{}))>> : std::true_type {};
+
+    struct single_threaded_collection_base
+    {
```
DefaultRyan, line 16: Seems like some code would be simplified if we removed the `has_exclusive_op` and `has_shared_op` traits, and the checks for said traits, and simply added `exclusive_op` and `shared_op` to `single_threaded_collection_base`, with these implementations simply calling the op.

strings/base_collections_base.h
DefaultRyan: I believe the version check is correct, and was missing from the old code. `m_current` and `m_end` can be invalidated by collection changes.

strings/base_collections_base.h
DefaultRyan: The iterator increment (`++m_current`) needs to be done under *some* exclusive lock, because the underlying container (e.g. `std::vector`) iterators provide no guarantees of atomicity.

My guess is this is a compromise to avoid adding a lock object to every iterator instance - when an iterator needs to lock itself exclusively, it simply piggy-backs on its owner's exclusive lock. This also means we only need to take a single lock, instead of taking the iterator's lock exclusive and the owner's lock shared (or would it be in the other order?). Doing a dance with two locks raises the complexity and makes it more likely that we'll accidentally introduce a deadlock.

But as a result, we're taking an exclusive lock, blocking other benign concurrent operations from running.

So I think I get the rationale, but I'm not completely, totally, 100% convinced it's the right decision. More like 80%.

strings/base_collections_base.h
```diff
@@ -70,43 +140,66 @@ WINRT_EXPORT namespace winrt
...
+                });
             }
 
             uint32_t GetMany(array_view<T> values)
             {
-                return GetMany(values, typename std::iterator_traits<iterator_type>::iterator_category());
+                return m_owner->perform_exclusive([&]
+                {
+                    this->check_version(*m_owner);
+                    return GetMany(values, typename std::iterator_traits<iterator_type>::iterator_category());
```
DefaultRyan, line 184: Also wondering why this is locking exclusive. I'd never noticed that `GetMany()` is modifying `m_current`. That can't be right, can it, @kennykerr? `Get()` doesn't modify it, so why would `GetMany()`?

strings/base_collections_base.h
```diff
@@ -142,40 +235,59 @@ WINRT_EXPORT namespace winrt
...
+        }
 
-            uint32_t const actual = (std::min)(Size() - startIndex, values.size());
-            this->copy_n(static_cast<D const&>(*this).get_container().begin() + startIndex, actual, values.begin());
-            return actual;
+    private:
+
+        uint32_t container_size() const noexcept
+        {
+            return static_cast<uint32_t>(std::distance(static_cast<D const&>(*this).get_container().begin(), static_cast<D const&>(*this).get_container().end()));
```
DefaultRyan, line 290: We seem inconsistent about the requirements we place on the container. Here, we only assume that it has `begin()` and `end()`, but make no assumptions about `size()`. But `vector_base` assumes that `size()` exists. I wonder why?

strings/base_collections_base.h
jonwis: Ah, I see. vector::iterator::operator++ might "do something" to the underlying vector (like, say, in debug mode?) so we have to go exclusive here as we can't assume anything about it.

strings/base_collections_base.h
kennykerr: Yes, that's correct. The GetMany on the collection doesn't modify, but the GetMany on the iterator consumes the iterator.

strings/base_collections_base.h
kennykerr: The priority for this PR should be:

1. Correctness
2. Simplicity of implementation
3. Performance

No one in their right mind is going to use a COM-based collection as a high-performance concurrent collection. This is about correctness in the face of accidental or intentional misuse. Obviously it needs to be correct, but beyond that we need to strive for the simplest practical solution as complexity here can really make this difficult to maintain.

strings/base_collections_base.h
kennykerr: We don't use SAL inside C++/WinRT. We do however adopt standard C++ attributes that are slowly taking the place of SAL.

strings/base_collections_map.h
kennykerr: I'd like to avoid feature creep unless there's a compelling need for this at the moment. SRW locks are fantastic and making it configurable won't necessarily make it usable by those with "interesting" locking requirements.

strings/base_collections_base.h
jonwis: Ok, no SAL, but maybe "current_value_withlock()" ?

strings/base_collections_base.h
dunhor: That's actually how I had it initially. The problem there is that these types are not in the `impl` namespace and I wanted to avoid a breaking change. In general, the following two approaches were considered:

OPTION 1: Have the `perform_exclusive` function in `iterable_base` just call the function object and then have `perform_shared` do the forwarding call to `perform_exclusive`, and then the implementation class can provide alternative definition(s) of those two functions. This is actually why there's still a bunch of CRTP casting done when calling `perform_*` that I neglected to remove. Oops. The reason I abandoned that route was because it made the implementations more difficult to write. I wanted the `*_threaded_collection_base` types that I could derive from, but then you'd have two base classes that define the same functions, which would require awkward `using Base::perform_*` statements. Overall, it's probably not that bad - at the time it was a little bit more awkward based on the way things were structured at the time. Now that the single/multi-threaded definitions are just usings, it's probably not a huge deal any more.

OPTION 2: Take the breaking change and add `ThreadingBase` as a template argument that persists all the way down to `iterable_base`. This is actually more-or-less required for doing per-iterator locking should we decide we want that. It's a pretty minor breaking change, so perhaps we don't care all that much.

strings/base_collections_base.h
dunhor: To add on to Ryan's comment, the invalidation checking was previously done by `abi_enter` which is called before each function. Since this check needs to be synchronized it was moved here, which no longer made the function `noexcept`.

strings/base_collections_base.h
dunhor: If you look at the existing internal collections - I was surprised by this behavior, too - you'll see that they do the same thing. An additional thing to note is that only vector's iterator guarantees sequential consistency here (i.e. concurrent calls to `GetMany` won't "overlap"). Map's iterator type forgoes sequential consistency while still being thread safe.

strings/base_collections_base.h
dunhor: Ryan's pretty much summarized my thoughts. From the PR description:

> The most noteworthy "flaw" is that mutating iterator operations lock the collection exclusively. This is mostly a limitation imposed by the base IIterable type which uses opaque iterators for the IIterator objects. In theory, the collection only needs to be locked in shared mode, however that would require the IIterator objects themselves to have a separate lock, which seemed like overkill and unlikely to be worth it. Compare this to the existing ABI collection types which know the underlying vector/map types and can perform interlocked operations - in the case of vector - or forgo sequential consistency while still guaranteeing safety - in the case of map.

Personally I don't think it's worth the complexity and the breaking change associated with "piping" information down to the base iterable type, but I'm of course open to other opinions.

Deadlock is always something to look out with when multiple locks are involved, however I'm fairly certain that acquiring the iterator lock first and then the iterable's lock second is guaranteed to be correct.

strings/base_collections_base.h
dunhor: 🤷‍♂️

kennykerr: Views can be constructed from iterator pairs and there's no `size` on `winrt::impl::range_container`.

strings/base_collections_base.h
dunhor: There's a couple of things to note here, but ultimately it's just taste. The most obvious limitation to doing it that way requires the return type to _at least_ be move constructible. E.g `std::lock_guard` wouldn't be an option since it is non-copyable and non-movable (of course `std::unique_lock` would be a viable alternative). In this case, it would seem as though `slim_lock_guard`/`slim_shared_lock_guard` *are* copy/move constructible, but it's actually unsafe to do so; this is a bug.

You also have the issue where the function would have to return _something_ as you can't do `auto guard = fn();` for a `void` returning function, e.g. in the case of the single threaded collections. A natural alternative might be to return a null `void*`, but then you'd need to consistently decorate the assignment with `[[maybe_unused]]`.

strings/base_collections_base.h
dunhor: The latest commit basically does option 1 from above. It's slightly awkward since both `single_threaded_collection_base` _and_ `iterable_base` define the same thing for `perform_shared`/`perform_exclusive`, but that's just a side effect of `using ThreadingBase::perform_*` needing to reference something that exists.

strings/base_collections_base.h
DefaultRyan: Good to know on `GetMany` - I had no idea. I suppose that if the existing collection behaviors all do this, it's a _de facto_ behavior and should be documented as such. I created a bug on the docs here: https://github.com/MicrosoftDocs/winrt-api/issues/1732

strings/base_collections_base.h
DefaultRyan: Makes sense. On a side note, I've been toying with the thought of introducing C++20 concepts to C++/WinRT in few places. The collection-related code, in particular, is an area that would benefit.

strings/base_collections_base.h
kennykerr: The idea is that a WinRT iterator can model a forward-only iterator so GetMany has to move the iterator forward.

strings/base_collections_base.h
kennykerr: Yes, there are a few C++20 features I'd like to adopt including span and coroutines #676 - we should think about forking cppwinrt with a servicing C++17 branch and start work on C++20 features in master. Just a little nervous about how long it will take for C++20 to become mainstream since Visual C++ doesn't even have a /std:c++20 flag yet.

strings/base_collections_base.h
DefaultRyan: Agreed - it makes total sense, especially when you think about how a C# IEnumerator would need to be projected as IIterator. IEnumerator has no GetMany - just Current and MoveNext.

It's just not immediately intuitive when our brains are used to IVector.GetMany being non-mutating. Hence, the UWP doc bug. :)

strings/base_collections_base.h
DefaultRyan: Agreed. At least with span, concepts, char8_t, consteval, chrono updates, and otheres, there are clear-cut feature test macros. Coroutine support has the potential to be messier.

test/test/multi_threaded_map.cpp
```diff
@@ -0,0 +1,313 @@
...
+    {
+        t.join();
+    }
+
+    REQUIRE(map.Size() == (size * threadCount));
+
+    // sum(i = 1 -> N){i} = N * (N + 1) / 2
+    auto sum = std::accumulate(begin(map), end(map), 0, [](int curr, auto&& pair) { return curr + conditional_unbox(pair.Value()); });
+    REQUIRE(sum == ((threadCount * (size - 1) * size) / 2));
+}
```
DefaultRyan, line 223: I suppose it could be possible that an element gets inserted, but into the wrong part of the map's binary tree. It's probably more likely that in debug mode, at least, this would trip an assert. But in release mode, it's conceivable this state could occur - moreso if an insertion happened while the tree was rebalancing itself.

It's probably rare, but it's also super-easy to iterate the map and verify that the keys are all strictly increasing, so consider adding that little check. Aside from that, these tests look solid.

strings/base_collections_base.h
oldnewthing: I was expecting something more like

```cpp
struct single_threaded_collection_base
{
    struct nop {};
    auto with_lock_guard() { return nop{}; }
    auto with_shared_lock_guard() { return nop{}; }
};
struct multi_threaded_collection_base
{
    auto with_lock_guard() { return lock_guard(m_mutex); }
    auto with_shared_lock_guard() { return shared_lock_guard(m_mutex); }
private:
   srwlock m_mutex;
};
```

and then the classes would do

```cpp
auto guard = this->with_lock_guard();
blah blah blah;
```

or

```cpp
auto guard = this->with_shared_lock_guard();
blah blah blah;
```

Most of the changes would be just dropping `auto guard = ...` into various methods, which makes it easier to review because that preserves existing indentation levels and increases confidence that no funny business snuck in.

The wrapped lambda is more complicated to use because you have to double-return (once to return a value from the lambda, and another to propagate that value out of the function). It also removes reliance on the compiler to inline the lambda.

strings/base_collections_base.h
```diff
@@ -189,64 +286,85 @@ WINRT_EXPORT namespace winrt
...
-            }
+                if (index >= static_cast<D const&>(*this).get_container().size())
+                {
+                    throw hresult_out_of_bounds();
+                }
 
-            this->increment_version();
-            static_cast<D&>(*this).get_container().erase(static_cast<D const&>(*this).get_container().begin() + index);
+                this->increment_version();
+                static_cast<D&>(*this).get_container().erase(static_cast<D const&>(*this).get_container().begin() + index);
```
oldnewthing, line 325: This also destructs the old item under the lock. Old value needs to be moved to an object that is created outside the lock, so that the destruction happens outside the lock.

strings/base_collections_base.h
```diff
@@ -189,64 +286,85 @@ WINRT_EXPORT namespace winrt
...
-            }
+                if (static_cast<D const&>(*this).get_container().empty())
+                {
+                    throw hresult_out_of_bounds();
+                }
 
-            this->increment_version();
-            static_cast<D&>(*this).get_container().pop_back();
+                this->increment_version();
+                static_cast<D&>(*this).get_container().pop_back();
```
oldnewthing, line 348: Again, destructing a foreign object inside the lock.

strings/base_collections_base.h
```diff
@@ -189,64 +286,85 @@ WINRT_EXPORT namespace winrt
...
         }
 
         void ReplaceAll(array_view<T const> value)
         {
-            this->increment_version();
-            assign(value.begin(), value.end());
+            static_cast<D&>(*this).perform_exclusive([&]
+            {
+                this->increment_version();
+                assign(value.begin(), value.end());
```
oldnewthing, line 366: Need to retain the old values so they are destructed outside the lock.

strings/base_collections_base.h
```diff
@@ -370,24 +488,33 @@ WINRT_EXPORT namespace winrt
...
         bool HasKey(K const& key) const noexcept
         {
-            return static_cast<D const&>(*this).get_container().find(static_cast<D const&>(*this).wrap_value(key)) != static_cast<D const&>(*this).get_container().end();
+            return static_cast<D const&>(*this).perform_shared([&]
+            {
+                return static_cast<D const&>(*this).get_container().find(static_cast<D const&>(*this).wrap_value(key)) != static_cast<D const&>(*this).get_container().end();
+            });
         }
 
         void Split(Windows::Foundation::Collections::IMapView<K, V>& first, Windows::Foundation::Collections::IMapView<K, V>& second) const noexcept
```
oldnewthing, line 520: I never did figure out what Split does.

strings/base_collections_base.h
```diff
@@ -407,27 +534,36 @@ WINRT_EXPORT namespace winrt
 
         bool Insert(K const& key, V const& value)
         {
-            this->increment_version();
-            auto pair = static_cast<D&>(*this).get_container().insert_or_assign(static_cast<D const&>(*this).wrap_value(key), static_cast<D const&>(*this).wrap_value(value));
-            return !pair.second;
+            return static_cast<D&>(*this).perform_exclusive([&]
+            {
+                this->increment_version();
+                auto pair = static_cast<D&>(*this).get_container().insert_or_assign(static_cast<D const&>(*this).wrap_value(key), static_cast<D const&>(*this).wrap_value(value));
```
oldnewthing, line 540: Need to save the old value so it can be destructed outside the lock.

strings/base_collections_base.h
```diff
@@ -407,27 +534,36 @@ WINRT_EXPORT namespace winrt
...
-            this->increment_version();
-            container.erase(found);
+                auto& container = static_cast<D&>(*this).get_container();
+                auto found = container.find(static_cast<D const&>(*this).wrap_value(key));
+                if (found == container.end())
+                {
+                    throw hresult_out_of_bounds();
+                }
+                this->increment_version();
+                container.erase(found);
```
oldnewthing, line 556: Need to save the old value so it can be destructed outside the lock.

strings/base_collections_base.h
```diff
@@ -407,27 +534,36 @@ WINRT_EXPORT namespace winrt
...
         }
 
         void Clear() noexcept
         {
-            this->increment_version();
-            static_cast<D&>(*this).get_container().clear();
+            static_cast<D&>(*this).perform_exclusive([&]
+            {
+                this->increment_version();
+                static_cast<D&>(*this).get_container().clear();
```
oldnewthing, line 565: Need to save the old values so they can be destructed outside the lock. Probably should just swap in an empty container.

test/test/multi_threaded_map.cpp
```diff
@@ -0,0 +1,313 @@
...
+    // the implementation does not attempt to recursively acquire the mutex.
+    template <typename Func>
+    auto exclusive_op(Func&& fn) const
+    {
+        slim_lock_guard guard(mutex);
+        return fn();
+    }
+};
+
+static void deadlock_test()
```
oldnewthing, line 253: Need a reentrancy test that validates that objects are not destructed under the lock.
```cpp
// Exercises the case where an object's destructor results in a call
// back into the container it was removed from. This test makes the call directly,
// but in real life, the path is usually more circuitous.
// E.g., destructor of object X releases the final reference to object Y,
// and object Y's destructor tries to remove itself from the same container.

template<typename Container>
struct CallsBackInDestructor : implements<CallsBackInDestructor, IInspectable>
{
    Container m_owner;
    CallsBackInDestructor(Container const& owner) : m_owner(owner) { }
    ~CallsBackInDestructor() { m_owner.Size(); }
};

void reentrancy_test()
{
    auto map = multi_threaded_map<int, IInspectable>();
    map.Insert(1, make<CallsBackInDestructor>(map));
    map.Insert(1, make<CallsBackInDestructor>(map)); // should not deadlock
    map.Remove(1); // should not deadlock

    map.Insert(1, make<CallsBackInDestructor>(map));
    map.Clear(); // should not deadlock
}
```

strings/base_collections_base.h
```diff
@@ -189,64 +286,85 @@ WINRT_EXPORT namespace winrt
...
-            }
+                if (index >= static_cast<D const&>(*this).get_container().size())
+                {
+                    throw hresult_out_of_bounds();
+                }
 
-            this->increment_version();
-            static_cast<D&>(*this).get_container()[index] = static_cast<D const&>(*this).wrap_value(value);
+                this->increment_version();
+                static_cast<D&>(*this).get_container()[index] = static_cast<D const&>(*this).wrap_value(value);
```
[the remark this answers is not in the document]
dunhor: Good catch

strings/base_collections_base.h
dunhor: Because non-movable types cannot be used as the return type of a function. [Example](https://godbolt.org/z/zv44E4). `std::lock_guard` is an example from the STL. `slim_lock_guard` is designed similarly. Of course, there are alternatives. `std::unique_lock` supports "invalidation" and therefore can be moved. Similarly we can add `unique_slim_lock_guard`, etc. Now you're trading off reliance that the compiler will inline lambdas/forwarding calls - something compilers are really good at - with reliance that the compiler will optimize validation checks - something compilers are also pretty good at.

jonwis: Locks might not be movable, but lock guards can be?  Yes you'd need to use a pointer back to the original lock, or some "do the unlock" flag if you keep the reference mode.  I don't get what "optimize validation checks" has to do with this.  Ultimately the "return a guard" or "initialize a guard" models just seem more readable with no loss of functionality.

kennykerr: I know it's frowned upon, but I've always preferred guards that can be returned from a function. Having said that, we'd like to remove all of the public locking primitives from C++/WinRT in favor of std alternatives.

dunhor: > I don't get what "optimize validation checks" has to do with this
Because the destructor would have to look like this:
```c++
~unique_lock()
{
    if (still_valid)
    {
        m_lock->unlock();
    }
}
```
That's the whole reason for both `lock_guard` and `unique_lock` exist. Don't pay for what you don't use

oldnewthing: You can return a non-movable type from a function, provided you use RVO. [Example](https://gcc.godbolt.org/z/xrjqP4). The class `NonMovable` is not movable (as evidenced by our inability to move it on line 17), but the function `g` happily returns it on line 11. Our usage pattern here would be RVO-friendly.

strings/base_collections_base.h
kennykerr: It does nothing. In theory it was meant to split the container for parallel search algorithms. That wasn't a well thought out plan. Fortunately a "valid" implementation can do nothing.

strings/base_collections_base.h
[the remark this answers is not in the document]
dunhor: The latest iteration is a first stab at addressing this. I'm mostly looking for feedback/input on the approach, though it is fully functional. The most important thing is probably delaying destruction of the "wrapped" value vs. delaying destruction of the "unwrapped" value. Of course, delaying destruction of the "wrapped" value implies that the destruction of the "unwrapped"/"underlying" value is also delayed. The approach here is to delay destruction of the "wrapped" value. This seemed to be the safest approach, but is not without its pros/cons.

* This covers the scenario where the "wrapped" type's destructor does something potentially unsafe. I'm unsure how realistic this scenario is, however I figured it's best to be safe.
* We can make fewer assumptions about the "wrapped" type, at least compared to WinRT types. In particular how/if it's constructible. The approach here is to use `std::optional` so that we can delay (move) construct the value, and that is the only requirement on the type put in place. If the type is not move constructible, we fall back to running the destructor under the lock (a con, but unlikely to ever be a problem). There's also a potential issue if the moved-from value's destructor does something unsafe, however that's pretty much unavoidable no matter what we do.
* As-is, we assume that anything that's not trivially destructible needs to be destructed outside of the lock. This is a bit more restrictive than necessary. E.g. I would assume that `hstring` should be safe to destroy under the lock (not necessarily a con; doing less under the lock is generally good). Similarly, I would assume that structs with `IReference` members should be safe to destroy under the lock (but I'm not 100% sure, correct me if I'm wrong).
* This doesn't take into account the "threadedness" of the collection. I.e. the single threaded collections still try and order destruction. This is mostly noteworthy for operations like `Clear` and `ReplaceAll` where we may otherwise be able to re-use the buffer. This is mostly a simplicity decision.

Mostly just wanted to get this out to get thoughts/feedback.

strings/base_collections_base.h
dunhor: Fair enough. I'm too used to assuming backwards compatibility with previous standards.

PR #704, opened 2020-07-29
Set Private to false for specific projects instead of everything.

jlaanstra:
Set Private explicitly to false for certain projects instead of setting it to false for everything and then setting it true for certain projects and allow projects to opt-out.

TODO:

- [x] Add docs for the new property.

### DIFF
--- a/nuget/CppWinrtRules.Project.xml
+++ b/nuget/CppWinrtRules.Project.xml
@@ -76,4 +76,9 @@
                 Description="Enables or disables the generation of Windows Metadata"
                 Category="General" />
 
+  <BoolProperty Name="CppWinRTEnableDefaultPrivateFalse"
+                DisplayName="Enable Copy Local Defaults"
+                Description="Enables or disables the default for copying bineries to the output folder"
+                Category="General" />
+
 </Rule>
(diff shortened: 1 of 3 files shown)


kennykerr:
@jlaanstra "/azp run" is no longer required.

kennykerr:
Maybe add documentation for what this is for here: https://github.com/microsoft/cppwinrt/tree/master/nuget

kennykerr:
Is this PR still active?

jlaanstra:
@kennykerr yes it is. I still need to add docs for the property.

nuget/CppWinrtRules.Project.xml
```diff
@@ -76,4 +76,9 @@
                 Description="Enables or disables the generation of Windows Metadata"
                 Category="General" />
 
+  <BoolProperty Name="CppWinRTEnableDefaultPrivateFalse"
+                DisplayName="Enable Copy Local Defaults"
+                Description="Enables or disables the default for copying bineries to the output folder"
```
jlaanstra, line 81:
```suggestion
                Description="Enables or disables the default for copying binaries to the output folder"
```

jlaanstra:
@kennykerr this is ready to merged.

kennykerr:
Thanks!

PR #726, opened 2020-08-14
coroutine_policy object to configure coroutine to terminate on unhandled exception

oldnewthing:
There is until now no way to make a coroutine fail fast on unhandled exception (a.k.a. "noexcept"), which makes debugging more frustrating.

```cpp
IAsyncOperation<bool> DoSomethingAsync(int v) try
{
    co_return TryIt(int v);
}
catch (...)
{
    winrt::terminate(); // fail fast on all unhandled exceptions
}
```

however this does not generate a good crash dump stack because the `catch (...)` handles the exception, so it unwinds. All you know is that *somebody* in your coroutine raised an exception. The stowed exception gives you a stack trace, but you really want to be able to inspect local variables, but those have already unwound and are no longer available.

You can also try to fake it with

```cpp
IAsyncOperation<bool> DoSomethingAsync(int v)
{
   co_return [&]() noexcept
   {
       return TryIt(v);
   }();
}
```

The `noexcept` on the lambda will terminate on unhandled exception. However, this doesn't interact well with `co_await`. You have to wrap every non-async piece separately, which is quite cumbersome. If the original function were

```cpp
IAsyncOperation<bool> DoSomethingAsync(int v)
{
    co_return co_await Part1(v) && co_await Part2(v);
}
```

the naive wrapper would be

```cpp
IAsyncOperation<bool> DoSomethingAsync(int v)
{
   co_return [](auto v) -> IAsyncOperation<bool> noexcept // ???
   {
       co_return co_await Part1(v) && co_await Part2(v);
   }(v);
}
```

(Note that the immediately-invoked lambda cannot capture anything. Anything that needs to survive the inner `co_await` must be captured into the frame by passing it as an explicit parameter.)

This doesn't work because you're basically back where you started. If an exception occurs in `Part1(v)` or `Part2(v)`, it will be stowed into the lambda's `IAsyncOperation<bool>` and the original context is lost.

You have to wrap it like this:

```cpp
IAsyncOperation<bool> DoSomethingAsync()
{
    auto task1 = [&]() noexcept
    {
        return Part1(); // fails fast if Part1() throws an exception without returning an IAsyncOperation
    }();
    bool result = false;
    try
    {
        result = co_await task1;
    }
    catch (...)
    {
        winrt::terminate(); // fails fast if Part1() completed with an exception
    }
    if (result)
    {
        auto task2 = [&]() noexcept
        {
            return Part2();  // fails fast if Part2() throws an exception without returning an IAsyncOperation
        }();
    try
    {
        result = co_await task2;
    }
    catch (...)
    {
        winrt::terminate(); // fails fast if Part2() completed with an exception
    }
   }
   co_return result;
}
```

Even with just two `co_await` operations in the original, the transformed code is already unreadable.

We introduce the "error token" which lets you configure the coroutine's error behavior.

```cpp
IAsyncOperation<bool> DoSomethingAsync()
{
    auto policy = co_await get_coroutine_policy();
    policy.terminate_on_unhandled_exception(); // unhandled exceptions are now fatal

    co_return co_await Part1() && co_await Part2(); // if anything throws an exception, we will fail fast
}
```

This time, an exception from `TryIt()` will fail fast with the exception-throwing frame still on the stack and therefore captured in the dump for post-mortem analysis.

The `terminate_on_unhandled_exception()` method takes a boolean parameter (default true) that specifies whether the coroutine should terminate. It returns the previous value, in case you want to restore it later.

**Note** that cancellation is not considered an unhandled exception. This means that if the coroutine is cancelled, or if it `co_await`s another coroutine that is cancelled, the coroutine transitions to the cancelled state as usual. (Basically, this says "terminate on transition to error state".)

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -243,6 +243,13 @@ WINRT_EXPORT namespace winrt
     {
         return{};
     }
+
+    struct get_error_token_t {};
+
+    inline get_error_token_t get_error_token() noexcept
+    {
+        return{};
+    }
 }
 
 namespace winrt::impl
@@ -312,6 +319,36 @@ namespace winrt::impl
 
     private:
 
+        Promise* m_promise;
+    };
+
+    template <typename Promise>
+    struct error_token
+    {
+        error_token(Promise* promise) noexcept : m_promise(promise)
+        {
+        }
+
+        bool await_ready() const noexcept
+        {
+            return true;
+        }
+
+        void await_suspend(std::experimental::coroutine_handle<>) const noexcept
+        {
+        }
+
+        error_token<Promise> await_resume() const noexcept
+        {
+            return *this;
+        }
+
+        bool set_noexcept(bool is_noexcept = true) const noexcept
+        {
+            return m_promise->set_noexcept(is_noexcept);
+        }
+
+    private:
         Promise* m_promise;
     };
 
@@ -512,6 +549,12 @@ namespace winrt::impl
         {
             slim_lock_guard const guard(m_lock);
             WINRT_ASSERT(m_status.load(std::memory_order_relaxed) == AsyncStatus::Started || m_status.load(std::memory_order_relaxed) == AsyncStatus::Canceled);
+
+            if (m_is_noexcept)
+            {
+                winrt::terminate();
+            }
+
             m_exception = std::current_exception();
 
             try
@@ -549,6 +592,11 @@ namespace winrt::impl
             return{ static_cast<Derived*>(this) };
         }
 
+        error_token<Derived> await_transform(get_error_token_t) noexcept
+        {
+            return{ static_cast<Derived*>(this) };
+        }
+
         void cancellation_callback(winrt::delegate<>&& cancel) noexcept
         {
             {
@@ -565,6 +613,11 @@ namespace winrt::impl
             {
                 cancel();
             }
+        }
+
+        bool set_noexcept(bool is_noexcept) noexcept
+        {
+            return std::exchange(m_is_noexcept, is_noexcept);
         }
 
 #if defined(_DEBUG) && !defined(WINRT_NO_MAKE_DETECTION)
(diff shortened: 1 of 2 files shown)


oldnewthing:
Better names for these objects and methods are cheerfully welcomed!

strings/base_coroutine_foundation.h
```diff
@@ -512,6 +549,12 @@ namespace winrt::impl
         {
             slim_lock_guard const guard(m_lock);
             WINRT_ASSERT(m_status.load(std::memory_order_relaxed) == AsyncStatus::Started || m_status.load(std::memory_order_relaxed) == AsyncStatus::Canceled);
+
+            if (m_is_noexcept)
+            {
+                winrt::terminate();
```
ChrisGuzak, line 555: maybe a comment here to explain how to diagnose the problem

oldnewthing: I think the new code makes it self-evident why we are terminating:

```cpp
if (m_terminate_on_unhandled_exception)
{
    winrt::terminate();
}
```

oldnewthing:
Sadly this doesn't help. The exception is already caught by the C++ language as part of coroutine exception management, and unwinding has already occurred.

```cpp
IAsyncAction Example()
{
    struct S
    {
        ~S() { printf("Destructed S\n"); }
    } s;
    throw hresult_access_denied();
}
```

The `s` is destructed before `unhandled_exception` is called, so by the time `unhandled_exception` runs, it's already too late.

The behavior is required by the language specification, which says that the coroutine transformation goes

```cpp
IAsyncAction Example
{
   Promise P;
   try {
      co_await P.initial_suspend();
      /* coroutine body goes here */
   } catch (...) { /// <<<<<<<<<< already caught
      P.unhandled_exception();
   }
   co_await P.final_suspend();
}
```

So getting better fail-fast dumps will have to be done some other way, outside the C++ language itself.

kennykerr:
Thanks Raymond.

jlaanstra:
@oldnewthing is there a way to configure the default to be true?

oldnewthing:
It doesn't matter because it doesn't help. The exception is already unwound by the time it hits the promise.

PR #734, opened 2020-08-21
Fix some visualizer crashes and misbehavior

DefaultRyan:
Users trying to debug Xaml apps were having issues that rendered the debugging experience unusable. Namely:

- Generic types in the property list would throw an exception, closing down the visualizer pane.
- Properties of type "System.Object" (aka IInspectable) weren't handled properly, causing undefined behavior via uninitialized variables. This was causing crashes and other bizarre behavior in the visualizer.
- Types that had more than one page of properties would repeatedly recreate and append the properties to the list when the debugger requested more of the list, causing the list to grow endlessly.

In this change:
- Generics changed from throwing to omitting, with a diagnostic message.
- IInspectable properties changed from bad behavior to omitting, with a diagnostic message. (If anybody has pointers on how to enable them, I'm listening).
- We only create the property list once for a type/object.

I tested this by excercising scenarios that were previously blowing up the visualizer. Specifically, setting a breakpoint in a Xaml Button click handler, and trying to inspect the "sender" object. This sender is a Xaml Button, and so its property list satisfies all three conditions: it has generics, IInspectables, and a long list of properties that requires the debugger to make multiple requests.

### DIFF
--- a/natvis/cppwinrtvisualizer.vcxproj
+++ b/natvis/cppwinrtvisualizer.vcxproj
@@ -205,10 +205,10 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClInclude Include="$(NugetPackagesDirectory)Microsoft.VSSDK.Debugger.VSDebugEng.*\inc\VSDebugEng.h" />
-    <ClInclude Include="$(NugetPackagesDirectory)Microsoft.VSSDK.Debugger.VSDebugEng.*\inc\vsdebugeng.templates.h" />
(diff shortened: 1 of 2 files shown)


arkrumbe:
I think it's very likely. The whole reason this bug came up is when I was using the locals window in VS and hit a breakpoint, and attempted to continue when the visualizer failed. I'm very much looking forward to this update. @Ryan<mailto:ryansh@microsoft.com>, let us know when there is an updated extension for us to try.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #737, opened 2020-09-02
Multithreading tests: Add dummy definition to keep compiler happy

oldnewthing:
This type is instantiated by the deleted constructor whose original purpose is to allow an implicit conversion from non-const iterator to const iterator. When applied to the const iterator itself, it enables an implicit conversion from garbage to const iterator. The conversion is deleted, but the compiler still instantiates it, so we define the type to keep everybody happy.

### DIFF
--- a/test/test/multi_threaded_common.h
+++ b/test/test/multi_threaded_common.h
@@ -237,6 +237,17 @@ namespace concurrent_collections
     {
         return it += offset;
     }
+
+    // Dummy definition to keep compiler happy. Instantiated by the
+    // constructor whose purpose is to offer an implicit conversion from
+    // non-const iterator to const iterator. When applied to a const
+    // iterator, it offers implicit construction from this garbage type
+    // to const iterator (triggering its instantiation), but then the constructor
+    // is deleted by SFINAE, so the type is never actually created at runtime.
+    template<typename Container>
+    struct concurrency_checked_random_access_iterator<Container, void, void>
+    {
+    };
 #pragma endregion
 
     struct concurrency_guard
--- a/test/test/multi_threaded_common.h
+++ b/test/test/multi_threaded_common.h
@@ -238,12 +238,12 @@ namespace concurrent_collections
         return it += offset;
     }
 
-    // Dummy definition to keep compiler happy. Instantiated by the
-    // constructor whose purpose is to offer an implicit conversion from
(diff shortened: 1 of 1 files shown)


test/test/multi_threaded_common.h
```diff
@@ -237,6 +237,17 @@ namespace concurrent_collections
...
         return it += offset;
     }
+
+    // Dummy definition to keep compiler happy. Instantiated by the
+    // constructor whose purpose is to offer an implicit conversion from
+    // non-const iterator to const iterator. When applied to a const
+    // iterator, it offers implicit construction from this garbage type
+    // to const iterator (triggering its instantiation), but then the constructor
+    // is deleted by SFINAE, so the type is never actually created at runtime.
+    template<typename Container>
```
oldnewthing, line 247: I sort of forgot to mention why we need this: If we didn't specialize, then the regular version above would be instantiated, but it doesn't work when `Iterator` is void!

kennykerr: Let's update the comment - that helps. 😉

kennykerr:
Do we know why x64 Release CI builds are failing?

oldnewthing:
> Do we know why x64 Release CI builds are failing?

According to [the logs](https://microsoft.visualstudio.com/Dart/_build/results?buildId=26764166&view=logs&j=9d906afe-3eb5-5d79-316d-86f3f3fa360c&t=9e4318d3-5f72-5045-a68e-f107ae8db597), the [build completed](https://microsoft.visualstudio.com/Dart/_build/results?buildId=26764166&view=logs&j=9d906afe-3eb5-5d79-316d-86f3f3fa360c&t=9e4318d3-5f72-5045-a68e-f107ae8db597&l=7174) and the [tests ran and passed](https://microsoft.visualstudio.com/Dart/_build/results?buildId=26764166&view=logs&j=9d906afe-3eb5-5d79-316d-86f3f3fa360c&t=9e4318d3-5f72-5045-a68e-f107ae8db597&l=7205). What failed was [a web request](https://microsoft.visualstudio.com/Dart/_build/results?buildId=26764166&view=logs&j=9d906afe-3eb5-5d79-316d-86f3f3fa360c&t=9e4318d3-5f72-5045-a68e-f107ae8db597&l=12) at the start of the build. The full URL isn't in the logs, so I don't know what the web request was trying to do.

PR #742, opened 2020-09-12
Targets that add compiler options must run before FixupCLCompileOptions

### LINKED ISSUE #741: Build breaks when adding to ComputeCompileInputsTargets because FixupCLCompileOptions doesn't run at the end
cppwinrt injects targets in `ComputeCompileInputsTargets` to add some logic to the build. Because cppwinrt _appends_ to this variable, the cppwinrt target will run after `FixupCLCompileOptions`, which breaks the build in some cases: `FixupCLCompileOptions` is supposed to convert /MP to /FS

Related: https://developercommunity.visualstudio.com/content/problem/1167617/build-breaks-when-adding-to-computecompileinputsta.html

asklar:
Fixes #741 
This is hit when forcing proc count in msbuild since /MP should also imply /FS (force synchronous writes) but that enforcement is done by `FixupCLCompileOptions` after all other `ComputeCompileInputsTargets` runs. If something runs after the fix up has already happened, we break the build (`Error C1041: cannot open program database`)

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -82,9 +82,12 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <BeforeClCompileTargets>
             $(BeforeClCompileTargets);CppWinRTAddXamlMetaDataProviderCpp;CppWinRTMakeProjections;
         </BeforeClCompileTargets>
+
+        <!-- Ensure ComputeCompileInputsTargets runs at the end so that FixupCLCompileOptions is the last to run -->
         <ComputeCompileInputsTargets>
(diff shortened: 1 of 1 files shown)


asklar:
CC @kennykerr. Looks like the C++ for WinRT Xlang internal Build task is stuck, mind kicking it?

kennykerr:
I just need @Scottj1s or @jlaanstra to review the msbuild goo.

asklar:
CC @yuehuang010 @olgaark to double check me

jlaanstra (approved):

jlaanstra:
This change breaks the razzle build with ```error MSB8027: Two or more files with the name of XamlMetaDataProvider.cpp will produce outputs to the same location. This can lead to an incorrect build result.```. Need to revert.

jlaanstra:
Turns out this is actually a bug in razzle. I'll send mail with details since this is a problem with an internal tool.

Issue #736, opened 2020-08-27
single_threaded_vector().GetView() causes crash when used with .NET Native.


jlaanstra:
When passing the `IVectorView<T>` to C# code, a crash happens in NET Native when trying to enumerate the collection.

The reason is that the `GetView()` method returns the same object back, just with a different interface. This means that the view's `GetRuntimeClassName` reports that it is an `IVector<T>` instead of `IVectorView<T>`, which messes up COM interop with .NET Native.

kennykerr:
I've been meaning to add simple `vector_view` and `map_view` functions that don't implement the mutable interfaces at all and can rely on much simpler internal locking for all scenarios. Such dedicated view implementations would solve this problem and more.

benstevens48:
Are you able to elaborate on exactly when this bug occurs? I noticed the title of the issue and realized it would be very bad for my project if this were indeed the case, but I can't reproduce the bug.

kennykerr:
I haven't seen the repro, and for the most part C++/WinRT collections work just fine with C# consumers. I don't actually consider this a C++/WinRT bug. Relying on what `GetRuntimeClassName` returns for non-classes is rather dubious to begin with. The fact that some language projection relies on this rather than `QueryInterface` is the real bug. Anyway, this is mostly about hardening C++/WinRT to improve compat.

benstevens48:
Having said that I couldn't reproduce the issue, I've now hit the crash in my app. I'll give the repro (which is bizarre) below, but first I have a few comments on the suggested workaround in #743. 

I don't understand the discussion about threading. Surely since everything is read-only, no locking/synchronization is needed. It's true that the iterator position changes, but surely a new iterator object is returned whenever the caller starts an iteration over the list, so I don't see why this requires locking (unless you expect the caller to share the iterator with another thread and use it simultaneously, but that would be quite weird). Therefore the workaround that I've come up with (I needed an immediate fix for my app), based on @jlaanstra's very helpful code in #743 and also the [docs](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/collections#base-classes-for-collections) is the following (`CreateCustomVectorView`).

```C++
    namespace WinrtHelpersImpl {
        namespace wfc = Windows::Foundation::Collections;

        template <typename T, typename Container>
[...]

PR #743, opened 2020-09-14
Add simple vector_view and map_view functions.

jlaanstra:
Add simple vector_view and map_view functions. This helps with #736 but doesn't actually fix the underlying issue. The IVectorView returned from GetView() still causes issues with .NET Native.


strings/base_collections_map.h
```diff
@@ -43,6 +43,18 @@ namespace winrt::impl
 
 WINRT_EXPORT namespace winrt
 {
+    template <typename K, typename V, typename Compare = std::less<K>, typename Allocator = std::allocator<std::pair<K const, V>>>
+    Windows::Foundation::Collections::IMap<K, V> map_view()
```
DefaultRyan, line 47: When the function's name is `map_view`, I'd expect it to return `IMapView`, not `IMap`.

strings/base_collections_map.h
```diff
@@ -43,6 +43,18 @@ namespace winrt::impl
...
+    {
+        return make<impl::input_map_view<K, V, std::map<K, V, Compare, Allocator>>>(std::map<K, V, Compare, Allocator>{});
+    }
+
+    template <typename K, typename V, typename Compare = std::less<K>, typename Allocator = std::allocator<std::pair<K const, V>>>
```
DefaultRyan, line 52: Why `map_view` have two overloads, while `vector_view` has a default param value?

strings/base_collections_map.h
jlaanstra: Oops. Will fix.

strings/base_collections_map.h
jlaanstra: Similarly to how single_threaded_vector has only one overload and single_threaded_map has two. I followed that.

kennykerr:
This should be thread-safe for its `IIterator` implementation (e.g. `MoveNext`). Paging @dunhor and @oldnewthing for thoughts on how we might reuse (or not) the thread-safe implementation in a simple way. 

Also, the implementations is not agile. It should be.

kennykerr:
Also, tests are needed that confirm that it basically works and specifically what interfaces it supports e.g. `IVectorView`, `IAgileObject, and `IIterable` and what it does not e.g. `IVector`. Also that it returns the correct values from `GetRuntimeClassName`.

dunhor:
From a naive first glance, it seems like it should be rather simple to do the same `ThreadingBase` trick with `input_vector_view` as is done with `vector_impl`. That wouldn't address the agility problem, so a new type would likely be necessary. Would also probably want `single_threaded_*` and `multi_threaded_*` variants of the functions as well then.

kennykerr:
@dunhor "Would also probably want single_threaded_* and multi_threaded_* variants of the functions as well then."

The only mutating state is the iterator position. I'd rather just make that thread-safe and avoid having more confusing options.

oldnewthing:
Would we even use this if we had a proper fix for 736? I.e., does this have useful life on its own, or is it just a stopgap? If it's just a stopgap, what's our wind-down story?

kennykerr:
I thought this was a fix for #736. I'm not interested in stopgap solutions. We can either "fix" it by having `GetView` return a new object or by minting new implementations (this PR). This seems simpler and less problematic for compat, but I'm happy to entertain other ideas. 

As an aside, I don't actually consider this a C++/WinRT bug. Relying on what `GetRuntimeClassName` returns for non-classes is rather dubious to begin with. The fact that some language projection relies on this rather than `QueryInterface` is the real bug.

jlaanstra:
This isn't a complete fix as GetView still has the same problems as before. However this does help for the cases where you essentially return siongle_threaded_vector().GetView(). However this class could be used for GetView() by making a copy of the underlying container to simplify ownership. In that case the view would have a copy of the data and wouldn't see any changes to the items in the IVector. Not sure if that's desirable.

kennykerr:
`GetView` does not suggest or imply a copy of the container. That would be a serious performance regression.

jlaanstra:
> @dunhor "Would also probably want single_threaded_* and multi_threaded_* variants of the functions as well then."
>
> The only mutating state is the iterator position. I'd rather just make that thread-safe and avoid having more confusing options.

I would probably needs some help with getting the threading sorted here. There clearly is more to it than I originally anticipated.

dunhor:
All of the `*_base` collections types are already set up for synchronization, however the default is to no-op for the synchronization step (i.e. the single threaded behavior). This behavior is overridden by providing implementations of the `acquire_shared`/`acquire_exclusive` functions on the derived type. Generally this would be done by deriving from `multi_threaded_collection_base` and then `using` the two names. E.g. see the definition of `vector_impl` for an example.

kennykerr:
Rethinking this, it's probably best just to update the existing GetView implementations to return a new object every time. I may just do that and then this problem goes away without needing to introduce new functions.

DefaultRyan:
> Rethinking this, it's probably best just to update the existing GetView implementations to return a new object every time. I may just do that and then this problem goes away without needing to introduce new functions.

Yep, this. 100%

strings/base_collections_map.h
```diff
@@ -43,6 +43,18 @@ namespace winrt::impl
 
 WINRT_EXPORT namespace winrt
 {
+    template <typename K, typename V, typename Compare = std::less<K>, typename Allocator = std::allocator<std::pair<K const, V>>>
+    Windows::Foundation::Collections::IMapView<K, V> map_view()
+    {
```
BENJAMINmlp, line 48: close vectors

PR #756, opened 2020-10-01
Remove /await flag for compat with C++20

### LINKED ISSUE #751: NuGet: 2.0.200921.6 release that introduced "C++20 coroutine support", still specifies /await in Microsoft.Windows.CppWinRT.targets file
Latest release that introduced "C++20 coroutine support" still forces old behavior by having `/await` option in `Microsoft.Windows.CppWinRT.targets` file in NuGet package.

kennykerr:
Fixes #751

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -829,6 +829,7 @@ $(XamlMetaDataProviderPch)
     <ItemDefinitionGroup>
       <ClCompile>
         <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
+        <AdditionalOptions Condition="'%(ClCompile.LanguageStandard)' == 'stdcpp17'">%(AdditionalOptions) /await</AdditionalOptions>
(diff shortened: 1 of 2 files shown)


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -828,7 +828,7 @@ $(XamlMetaDataProviderPch)
     <!--Append any additional item metadata after all default and project settings have been applied-->
     <ItemDefinitionGroup>
       <ClCompile>
-        <AdditionalOptions>%(AdditionalOptions) /bigobj /await</AdditionalOptions>
```
jlaanstra, line 831: I was about to submit a PR that adds this below here:

```
        <AdditionalOptions Condition="'%(ClCompile.LanguageStandard)' == 'stdcpp17'">%(AdditionalOptions) /await</AdditionalOptions>
```

That would keep the behavior the same for folks who still target stdcpp17 and anyone who kept the defaults from the project templates.

kennykerr: Feel free to add it to this branch or just submit your branch for PR and we can scrap this PR.

kennykerr: Let me know if I got that right.

PR #757, opened 2020-10-02
VS 2017 compatibility for enable_await_cancellation

oldnewthing:
VS 2017 requires an accessible copy constructor even though it is never used. Turns out we're okay with `enable_await_cancellation` being copied, as long as it's not copied while awaiting (which we don't do). So go ahead and let it be copyable.

VS 2017 also reports an unused variable error when the use of the `promise` is removed via `if constexpr`. Mark it as `[[maybe_unused]]` to help the compiler.


kennykerr (approved):
Thanks!

Issue #760, opened 2020-10-06
2.0.200921.6 build fail on VS 2017: The "IgnoreCase" parameter is not supported by "Hash" task.


EmmettTsai:
2.0.200917.4 and 2.0.200921.6 build fail on  VS 2017 15.9.27

ex: create Windows Desktop Application (C++/WinRT) project and build x86 got error messge:
`Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB4064	The "IgnoreCase" parameter is not supported by the "Hash" task. Verify the parameter exists on the task, and it is a settable public instance property.	WindowsApplication3	C:\Emmett\Project\WindowsApplication3\packages\Microsoft.Windows.CppWinRT.2.0.200921.6\build\native\Microsoft.Windows.CppWinRT.targets	473	
`
`Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB4063	The "Hash" task could not be initialized with its input parameters. 	WindowsApplication3	C:\Emmett\Project\WindowsApplication3\packages\Microsoft.Windows.CppWinRT.2.0.200921.6\build\native\Microsoft.Windows.CppWinRT.targets	471	
`

kennykerr:
@jlaanstra care to comment about the msbuild error? 

We don't really do much testing with Visual Studio 2017 any longer. If there's a simple fix I'd be happy to accept a PR, but otherwise I suggest you upgrade or stick to an older nuget package version until you can do so.

jlaanstra:
I think we'll just turn off the use of the has task for VS 2017.

kennykerr:
@EmmettTsai I published a new build - hopefully that helps.

PR #762, opened 2020-10-07
Fix build error IgnoreCase parameter is not supported by the Hash task on VS 2017

EmmettTsai:
PR for [https://github.com/microsoft/cppwinrt/issues/760](https://github.com/microsoft/cppwinrt/issues/760)
I test build for my winrt UWP project via IDE and cmdline on VS 2017 15.9.27/ VS 2019 16.7.5.


jlaanstra:
@EmmettTsai Thanks for the PR. I think I prefer to just not use the Hash task on VS 2017. I submitted a small change for that.

PR #765, opened 2020-10-13
Define co_await operators for C++20 coroutines

### LINKED ISSUE #764: Windows.Foundation await adapters do not compile with C++20 standard coroutines
With cppwinrt 2.0.201008.2, the following does not compile with Visual Studio 2019 16.8 Preview 3.2 with /std:c++latest:
```c++
#include <coroutine>
#include <winrt/base.h>
#include <winrt/Windows.Foundation.h>
#include <winrt/Windows.Devices.Bluetooth.h>

using namespace winrt;

using winrt::Windows::Foundation::IAsyncOperation;

IAsyncOperation<Windows::Devices::Bluetooth::BluetoothLEDevice> foo()
{
    auto bluetoothAdapter = co_await Windows::Devices::Bluetooth::BluetoothAdapter::GetDefaultAsync();
    co_return nullptr;
}
```
with the error:
```
...Generated Files\winrt\base.h(8415,46): error C2338: Not an awaitable type
...Generated Files\winrt\base.h(8460,30): error C2039: 'await_resume': is not a member of 'winrt::Windows::Foundation::IAsyncOperation<winrt::Windows::Devices::Bluetooth::BluetoothAdapter>'
```

If the build is reverted to legacy coroutines by passing `/await` and including `<experimental/coroutine>` instead, the code compiles successfully.

The problem appears to be that the various `co_await` operators in Windows.Foundation.h have the following conditional:
```c++
#if defined(__cpp_coroutines)
```
which is the feature-test macro for the legacy Coroutines TS, rather than also including `__cpp_lib_coroutine`. Furthermore, the adapters themselves may need some adjustments, although I have not looked into that.
There appear to be a few other places that still only test for the Coroutines TS macro.

kobykahane:
Fixes #764.

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -223,7 +223,7 @@ namespace winrt::impl
     }
 }
 
-#ifdef __cpp_coroutines
+#ifdef WINRT_IMPL_COROUTINES
 WINRT_EXPORT namespace winrt::Windows::Foundation
 {
     inline impl::await_adapter<IAsyncAction> operator co_await(IAsyncAction const& async)
--- a/strings/base_coroutine_system.h
+++ b/strings/base_coroutine_system.h
@@ -41,7 +41,7 @@ WINRT_EXPORT namespace winrt
         return awaitable{ dispatcher, priority };
     };
 
-#ifdef __cpp_coroutines
+#ifdef WINRT_IMPL_COROUTINES
     inline auto operator co_await(Windows::System::DispatcherQueue const& dispatcher)
     {
         return resume_foreground(dispatcher);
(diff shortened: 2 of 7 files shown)


sylveon:
Couldn't that mislead other libraries which use the macro for feature detection as well?

kennykerr:
Of course, if we can fix base_macros.h to not define a feature macro and then update the rest, as Koby originally did, to rely only on those feature macros that would save introducing yet another macro.

kobykahane:
@kennykerr Any additional comments?

kennykerr:
That looks great - I've spun a build. If all looks good I'll merge this in - thanks!

kennykerr (approved):
Looks great - thanks Koby!

PR #769, opened 2020-10-17
add try_get_activation_factory overloads that take the runtimeclass name

ChrisGuzak:
We have a design pattern "extension classes" that uses runtimeclasses as extensions. The class names are discovered at runtime, so we need to be able to specify them to the activation functions. this adds that support.

### DIFF
--- a/strings/base_activation.h
+++ b/strings/base_activation.h
@@ -401,9 +401,8 @@ namespace winrt::impl
     }
 
     template <typename Interface = Windows::Foundation::IActivationFactory>
-    com_ref<Interface> try_get_activation_factory(std::wstring_view className, hresult_error* exception = nullptr) noexcept
+    com_ref<Interface> try_get_activation_factory(param::hstring const& name, hresult_error* exception = nullptr) noexcept
     {
-        param::hstring const name{ className };
         void* result{};
         hresult const hr = get_runtime_activation_factory<Interface>(name, &result);
 
@@ -463,7 +462,7 @@ WINRT_EXPORT namespace winrt
     {
         // Normally, the callback avoids having to return a ref-counted object and the resulting AddRef/Release bump.
         // In this case we do want a unique reference, so we use the lambda to return one and thus produce an
-        // AddRef'd object that is returned to the caller. 
+        // AddRef'd object that is returned to the caller.
         return impl::call_factory<Class, Interface>([](auto&& factory)
         {
             return factory;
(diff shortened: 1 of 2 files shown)


Scottj1s (approved):

strings/base_activation.h
```diff
@@ -400,10 +400,10 @@ namespace winrt::impl
         return factory.call(static_cast<CastType>(callback));
     }
 
-    template <typename Class, typename Interface = Windows::Foundation::IActivationFactory>
-    com_ref<Interface> try_get_activation_factory(hresult_error* exception = nullptr) noexcept
+    template <typename Interface = Windows::Foundation::IActivationFactory>
+    com_ref<Interface> try_get_activation_factory(std::wstring_view className, hresult_error* exception = nullptr) noexcept
     {
-        param::hstring const name{ name_of<Class>() };
+        param::hstring const name{ className };
```
kennykerr, line 406: This named temporary isn't needed - may just want to pass the `className` directly to the `get_runtime_activation_factory` function as it can create the `param::hstring` automatically.

strings/base_activation.h
```diff
@@ -473,13 +473,25 @@ WINRT_EXPORT namespace winrt
     template <typename Class, typename Interface = Windows::Foundation::IActivationFactory>
     auto try_get_activation_factory() noexcept
     {
-        return impl::try_get_activation_factory<Class, Interface>();
+        return impl::try_get_activation_factory<Interface>(winrt::name_of<Class>());
```
kennykerr, line 476: Remove `winrt::` as it is implied.

strings/base_activation.h
```diff
@@ -473,13 +473,25 @@ WINRT_EXPORT namespace winrt
...
     }
 
     template <typename Class, typename Interface = Windows::Foundation::IActivationFactory>
     auto try_get_activation_factory(hresult_error& exception) noexcept
     {
-        return impl::try_get_activation_factory<Class, Interface>(&exception);
+        return impl::try_get_activation_factory<Interface>(winrt::name_of<Class>(), &exception);
```
kennykerr, line 482: Remove `winrt::` as it is implied.

strings/base_activation.h
```diff
@@ -400,10 +400,10 @@ namespace winrt::impl
         return factory.call(static_cast<CastType>(callback));
     }
 
-    template <typename Class, typename Interface = Windows::Foundation::IActivationFactory>
-    com_ref<Interface> try_get_activation_factory(hresult_error* exception = nullptr) noexcept
+    template <typename Interface = Windows::Foundation::IActivationFactory>
+    com_ref<Interface> try_get_activation_factory(std::wstring_view className, hresult_error* exception = nullptr) noexcept
```
kennykerr, line 404: This should be a `param::hstring const& name` parameter - this is the "string view" type we use internally for improved performance.

strings/base_activation.h
```diff
@@ -473,13 +473,25 @@ WINRT_EXPORT namespace winrt
...
 
     template <typename Class, typename Interface = Windows::Foundation::IActivationFactory>
     auto try_get_activation_factory(hresult_error& exception) noexcept
     {
-        return impl::try_get_activation_factory<Class, Interface>(&exception);
+        return impl::try_get_activation_factory<Interface>(winrt::name_of<Class>(), &exception);
+    }
+
+    template <typename Interface = Windows::Foundation::IActivationFactory>
+    auto try_get_activation_factory(std::wstring_view name) noexcept
```
kennykerr, line 486: This should be a `param::hstring const& name` parameter - this is the "string view" type we use internally for improved performance.

strings/base_activation.h
```diff
@@ -473,13 +473,25 @@ WINRT_EXPORT namespace winrt
...
+    {
+        return impl::try_get_activation_factory<Interface>(name);
+    }
+
+    template <typename Interface = Windows::Foundation::IActivationFactory>
+    auto try_get_activation_factory(std::wstring_view name, hresult_error& exception) noexcept
```
kennykerr, line 492: This should be a `param::hstring const& name` parameter - this is the "string view" type we use internally for improved performance.

kennykerr (changes requested):
Looks good - just a few comments. Generally, the `std::string_view` should be avoided for systems programming. We do support it, but we should avoid using it any more than necessary as it is not strictly safe and `param::hstring` gives us more freedom to distance ourselves from it in future.

test/old_tests/UnitTests/get_activation_factory.cpp
```diff
@@ -67,3 +68,37 @@ TEST_CASE("try_get_activation_factory")
...
+    {
+        auto factory = try_get_activation_factory(winrt::name_of<Component::IErrors>());
+        REQUIRE(factory == nullptr);
+        REQUIRE(get_error_info() == nullptr);
+    }
+
+    // Try successfully with error info
+    {
+        hresult_error e;
+        auto factory = try_get_activation_factory(winrt::name_of<Component::Errors>());
```
kennykerr, line 91: Here's you're forgetting the error info param.

kennykerr (approved):
Thanks!

PR #774, opened 2020-10-19
Use the CppWinRTPlatformWinMDReferences itemgroup as the input for the platform projection.

### LINKED ISSUE #773: Build targets pass a separate list of reference to MIDL then they generate a platform projection for.
We should use the same list of winmds to avoid build errors where we passed a WinMD to MIDL but didn't generate a projection header.

jlaanstra:
Fixes #773 

## TODO

- [x] Validate razzle build.

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -226,11 +226,11 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <!--Define platform projection WinMD inputs-->
     <Target Name="GetCppWinRTPlatformWinMDInputs"
-            DependsOnTargets="ResolveAssemblyReferences"
+            DependsOnTargets="ResolveAssemblyReferences;GetCppWinRTPlatformWinMDReferences"
             Returns="@(CppWinRTPlatformWinMDInputs)">
         <ItemGroup>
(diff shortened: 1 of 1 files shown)


jlaanstra:
Validated the razzle build after reverting #742

kennykerr (approved):

DHowett:
[terminal](https://github.com/microsoft/terminal) is seeing some build failures due to platform types missing when transitively used in projects that depend on winmds produced by other components with the latest version; could this be related?

> MDMERGE : error MDM2002: The type "Windows.Foundation.IReference`1" defined in file Microsoft.Terminal.TerminalControl.winmd was referenced by type Microsoft.Terminal.TerminalControl.ICoreSettings but could not be found. [src\cascadia\UnitTests_TerminalCore\UnitTests.vcxproj]

```
cppwinrt : error Type 'Windows.Foundation.EventRegistrationToken' could not be found
 method: TerminalOutput
 type: Microsoft.Terminal.TerminalConnection.ITerminalConnection
 database: C:\Users\Dustin\src\terminal\x64\Release\TerminalConnection\Microsoft.Terminal.TerminalConnection.w
inmd
```

DHowett:
(Rolling back one version to 2.0.201017.1 fixes our build. 2.0.201026 is broken; this seems to be the only changeset that addresses platform winmds in the new version.)

jlaanstra:
@DHowett can you share a binlog from your build? Happy to take a look.

PR #781, opened 2020-10-27
Add default WinMD references to the desktop templates.

### LINKED ISSUE #780: No generated code for console app
@jlaanstra could this PR https://github.com/microsoft/cppwinrt/pull/774 have broken the "Windows Console Application" project template? It no longer generates any headers. The "Windows Runtime Component" project template still works. 

It's the only msbuild-related change since 2.0.201017.1 where this project template still works.

jlaanstra:
For Universal Windows templates the xaml build targets and C++ build targets add the platform WinMDs to the reference list. However this doesn't happen for desktop templates , because the TargetPlatformIdentifier is not set to UAP.

We have two options:
- Set TargetPlatform to UAP
- Explicitly add the references.

Since these are desktop templates, option 2 seems preferable.

A possibly even better solution would be the WinRT targets to add the references. However we would have to fight the default C++ build targets to have them stop automatically adding the references and that would be quite a bit of work and possibly break xaml.

Fixes #780


kennykerr:
Will this also fix https://github.com/microsoft/cppwinrt/pull/774#issuecomment-716960570? cc @DHowett

DHowett:
We've been told by the VC project system folks in no uncertain terms that wildcards are explicitly disallowed in vcxprojs (outside of a Target that generates them!) and they'll go out of their way to break them. This solution may conflict with that.

jlaanstra:
I am personally not a fan of this solution either. Problem is that there is no good way to add them one by one explicitly as the folder location differs per version and VS doesn't add and the C++ targets don't add any WinMD for non-UAP projects.

kennykerr:
The project system isn't exactly abounding in support for C++ developers. Much of this is making the best of some bad choices.

jlaanstra:
Let's hold off on merging this for now. I'm gonna add a target that adds the platform winmds and then have desktop templates opt-in. That's probably a better solution.

jlaanstra:
> isn't exactly abounding in support for C++ developers. Much of this is making the best of some bad choices.

I'll try to find a good (or at least much better) choice here.

jlaanstra:
> We've been told by the VC project system folks in no uncertain terms that wildcards are explicitly disallowed in vcxprojs (outside of a Target that generates them!) and they'll go out of their way to break them. This solution may conflict with that.

Thanks for calling this out @DHowett

jlaanstra:
Updated with a target that mimics what the xaml targets do, including keeping the support for the existing overrides. I don't think we have to document this as we already have CppWinRTOverrideSDKReferences.

kennykerr (approved):
Thanks!

PR #792, opened 2020-11-13
Remove visualizer from vsix

kennykerr:
This is blocking inclusion in Visual Studio.


BenJKuhn:
I would still prefer to build the visualizer, and only make the changes necessary to exclude it from packaging, so that we can keep it "fresh" and revive it later. I know that there are still people that value this. Is that reasonable to do?

kennykerr:
Sure, I can keep it building until we find it a new home.

PR #793, opened 2020-11-13
Revert nuget to 2.0.200609.3

kennykerr:
This is the last stable version that was ingested into the OS.

### DIFF
--- a/nuget/CppWinrtRules.Project.xml
+++ b/nuget/CppWinrtRules.Project.xml
@@ -76,9 +76,4 @@
                 Description="Enables or disables the generation of Windows Metadata"
                 Category="General" />
 
-  <BoolProperty Name="CppWinRTEnableDefaultPrivateFalse"
-                DisplayName="Enable Copy Local Defaults"
-                Description="Enables or disables the default for copying binaries to the output folder"
-                Category="General" />
-
 </Rule>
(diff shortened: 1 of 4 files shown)


BenJKuhn (approved):

jlaanstra:
Nothing needed to be reverted in the targets?

jlaanstra:
This effectively blocks my team from taking newer versions as this reverts a few necessary fixes /cc @asklar. If we no longer want to maintain the build support, can we ship the build support separately so I can fork build support without needing to fork the binaries?

kennykerr:
I don't think we can easily make further packaging changes as VS is intending to consume the package as-is, but you are free to fork the build support without needing any change from cppwinrt.  You can also just grab `cppwinrt.exe` (ignore everything else) and use your own build support. Then you're entirely independent and only rely on cppwinrt for code gen (as it should be).

asklar:
I'm lacking some context here... What's the issue that this is unblocking/what is broken for the OS?

jlaanstra:
@asklar this reverts also your fix for the target ordering related to accessing pdb's.

kennykerr:
This is about reverting to a stable baseline. Since then, a few people have contributed tweaks but that resulted in a stream of regressions. I'll add you to an internal thread.

PR #803, opened 2020-11-28
Fix x64 builds of cppwinrt to properly locate the SDK

asklar:
https://task.ms/29349404 - The SDK sometimes stores the 64 bit location into KitsRoot10 which is wrong,
This breaks 64-bit cppwinrt.exe, so work around this by forcing to use the WoW64 hive.

### DIFF
--- a/cppwinrt/cmd_reader.h
+++ b/cppwinrt/cmd_reader.h
@@ -137,7 +137,9 @@ namespace cppwinrt
             HKEY_LOCAL_MACHINE,
             L"SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots",
             0,
-            KEY_READ,
+            // https://task.ms/29349404 - The SDK sometimes stores the 64 bit location into KitsRoot10 which is wrong,
+            // this breaks 64-bit cppwinrt.exe, so work around this by forcing to use the WoW64 hive.
+            KEY_READ | KEY_WOW64_32KEY, 
             &key))
         {
             throw std::invalid_argument("Could not find the Windows SDK in the registry");


kennykerr (approved):
Thanks! Did you manually verify that cppwinrt still works when the SDK is installed in the default arch location? Unfortunately, the build won't verify this.

asklar:
> Thanks! Did you manually verify that cppwinrt still works when the SDK is installed in the default arch location? Unfortunately, the build won't verify this.

The default location is x86. The problem only occurs when trying to use **x64** cppwinrt (which is what the readme says to do). An x86 cppwinrt.exe will always use the value in the virtualized reg key under wow6432node.

Building with the cppwinrt nuget will always pick up the x86 key since the nuget package only contains the x86 cppwinrt.exe:
```
nuget pack nuget\Microsoft.Windows.CppWinRT.nuspec -Properties cppwinrt_exe=%cd%\_build\x86\Release\cppwinrt.exe;cppwinrt_fast_fwd_x86=%cd%\_build\x86\Release\cppwinrt_fast_forwarder.lib;cppwinrt_fast_fwd_x64=%cd%\_build\x64\Release\cppwinrt_fast_forwarder.lib;cppwinrt_fast_fwd_arm=%cd%\_build\arm\Release\cppwinrt_fast_forwarder.lib;cppwinrt_fast_fwd_arm64=%cd%\_build\arm64\Release\cppwinrt_fast_forwarder.lib
```

Verified that installing an old standalone SDK puts the wrong key in the native registry (and the right key in the x86 registry).

kennykerr:
Got it. We only ship the x86 version of cppwinrt. We tend to use the x64 version for development.

PR #805, opened 2020-11-29
Mention prebuild in readme as it is needed by the cppwinrt project

asklar:


kennykerr (approved):
I didn't mention it because it should be a project build dependency and thus should build automatically, but I don't mind mentioning it here as well.

asklar:
@kennykerr the instructions in the readme tell folks to first build the cppwinrt project by itself only (right click / build cppwinrt only), which builds it with /p:BuildProjectReferences=false, which is why prebuild isn't built unless you try to build everything in the solution.

kennykerr:
That's weird. When I right click and build `cppwinrt` it always builds `prebuild` first. Because of this:

image

asklar:
@kennykerr  My repro:
- `git clean -xdf`
- Open cppwinrt.sln, right click on the cppwinrt project -> Project only -> build only cppwinrt
image
- Get this error:
image

kennykerr:
I never even noticed that menu option! I didn't literally mean only build `cppwinrt` but I can see how that can be confusing. 😄

PR #806, opened 2020-11-29
Produce IntelliSense comments from winmd custom attributes

### LINKED ISSUE #804: Produce IntelliSense comments from winmd custom attributes
Implements proposal to leverage any attribute named HelpStringAttribute (in any namespace).
This PR will produce comments in the projected code of the form `/// <summary>comment</summary>`, where `comment` is a string stored in a custom attribute named HelpStringAttribute

Assume this IDL:
```idl
namespace Windows.Foundation.Metadata
{
    [attributeusage(target_runtimeclass, target_interface, target_struct, target_enum, target_delegate, target_field, target_property, target_method, target_event)]
    [attributename("helpstring")]
    attribute HelpStringAttribute {
      String Content;
    }
}

namespace test_component
{
    [helpstring("this is an interface")]
    interface InterfaceWithDoc
    {
      [helpstring("this is a function with docs")]
      void f();

      [helpstring("this is a property")]
      Int32 InterfacePropertyWithDoc;
    };

    [helpstring("this is an enum")]
    enum EnumWithDoc
    {
      Dog,
      Cat,
      [helpstring("this is a gecko")] // xmldoc in enum values aren't properly displayed by VS yet
      Gecko
    };

    [helpstring("this is a struct")]
    struct StructWithDoc
    {
      [helpstring("this is the value")]
      Int32 val;
    };

    [helpstring("this is a class")]
    runtimeclass ClassWithDoc
    {
      [helpstring("this is a property")]
      Int32 SomeValue;

      [helpstring("this is a method")]
      void g();
    }
```
With this you get these kind of suggestions:

## on an interface
image

## on a method in an interface
image

## on a property in an interface
image

## on enums
image

## on enum values
image

## on structs
image

## on struct fields
image

## on classes
image

## on a method in a class
image

## on a property in a class
image

asklar:
Fixes #804 
# Description

Implements proposal to leverage any attribute named `HelpStringAttribute` (in any namespace).
This PR will produce comments in the projected code of the form `/// <summary>comment</summary>`, where `comment` is a string stored in a custom attribute named `HelpStringAttribute`.

# NYI
- cppwinrt has some text-replacement logic (`^`, `%`, `@`) which the doc strings need to be exempted from.
- XML comments seem to work in VS but they're not officially supported ([feature request filed](https://developercommunity.visualstudio.com/idea/1271536/c-intellisense-doc-comments-should-support-templat.html))
- test which should check that when building with /doc, the intellisense XML produced has the right entries.

# Demo
Assume this IDL:
```idl
namespace Windows.Foundation.Metadata
{
    [attributeusage(target_runtimeclass, target_interface, target_struct, target_enum, target_delegate, target_field, target_property, target_method, target_event)]
    [attributename("helpstring")]
    attribute HelpStringAttribute {
      String Content;
    }
}

namespace test_component
{
    [helpstring("this is an interface")]
    interface InterfaceWithDoc
    {
      [helpstring("this is a function with docs")]
      void f();

      [helpstring("this is a property")]
      Int32 InterfacePropertyWithDoc;
    };

    [helpstring("this is an enum")]
    enum EnumWithDoc
    {
      Dog,
      Cat,
      [helpstring("this is a gecko")] // xmldoc in enum values aren't properly displayed by VS yet
      Gecko
    };

    [helpstring("this is a struct")]
    struct StructWithDoc
    {
      [helpstring("this is the value")]
      Int32 val;
    };

    [helpstring("this is a class")]
    runtimeclass ClassWithDoc
    {
      [helpstring("this is a property")]
      Int32 SomeValue;

      [helpstring("this is a method")]
      void g();
    }
```
With this you get these kind of suggestions:

## on an interface
image

## on a method in an interface
image

## on a property in an interface
image

## on enums
image

## on enum values
image

## on structs
image

## on struct fields
image

## on classes
image

## on a method in a class
image

## on a property in a class
image

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1,5 +1,23 @@
 #pragma once
 
+namespace winmd::reader {
+  template <typename T>
+  CustomAttribute get_attribute(T const& row, std::string_view const& type_name)
+  {
+    for (auto&& attribute : row.CustomAttribute())
+    {
+      auto pair = attribute.TypeNamespaceAndName();
+
+      if (pair.second == type_name)
+      {
+        return attribute;
+      }
+    }
+
+    return {};
+  }
+}
+
 namespace cppwinrt
 {
     struct finish_with
@@ -185,6 +203,38 @@ namespace cppwinrt
         return { w, write_close_namespace };
     }
 
+
+    static std::string getDocXml(std::string_view doc_string)
+    {
+      std::string ret{};
+      if (!doc_string.empty()) {
+        ret = "/// <summary>" + std::string(doc_string) + "</summary>\r\n";
+      }
+      return ret;
+    }
+
+    template<typename T>
+    static std::string getDocXml(T const& type)
+    {
+      if (auto const& attr = get_attribute(type, "HelpStringAttribute"))
+      {
+        auto sig = attr.Value();
+        auto fixed = sig.FixedArgs();
+        if (fixed.size() == 1) {
+          return getDocXml(std::get<std::string_view>(std::get<ElemSig>(fixed[0].value).value));
+        }
+        else {
+          auto named = sig.NamedArgs();
+          auto content = std::find_if(named.begin(), named.end(), [](NamedArgSig const& namedsig) { return namedsig.name == "Content"; });
+          if (content != named.end()) {
+            auto value = content->value;
+            return getDocXml(std::get<std::string_view>(std::get<ElemSig>(value.value).value));
+          }
+        }
+      }
+      return {};
+    }
+
     static void write_enum_field(writer& w, Field const& field)
     {
         auto format = R"(        % = %,
@@ -192,7 +242,8 @@ namespace cppwinrt
 
         if (auto constant = field.Constant())
         {
-            w.write(format, field.Name(), *constant);
+            auto formatWithDoc = getDocXml(field) + format;
+            w.write(formatWithDoc, field.Name(), *constant);
         }
     }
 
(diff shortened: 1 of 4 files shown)


kennykerr:
This is very cool. Just want to set expectations. My main reservation is that this is a feature that has zero adoption currently. We added a few features over the years that were meant to be used by various key customers and gain adoption but ended up not being used at all. Such features add considerable liability to the code base and supportability of cppwinrt. Let's see what we can do about getting more support for this idea before committing this change. Happy to discuss further.

cppwinrt/code_writers.h
```diff
@@ -185,14 +203,47 @@ namespace cppwinrt
         return { w, write_close_namespace };
     }
 
+
+    static std::string getDocXml(std::string_view doc_string)
```
kennykerr, line 207: Casing should be `get_doc_xml`.

cppwinrt/code_writers.h
```diff
@@ -185,14 +203,47 @@ namespace cppwinrt
...
+    {
+      std::string ret{};
+      if (!doc_string.empty()) {
```
kennykerr, line 210: Braces should be:

```C++
if (!doc_string.empty())
{
    // ...
}
```

asklar:
thanks @kennykerr! We discussed this in the api review meeting and folks seemed supportive. Would that count as enough, or is there something else you'd like to see to make this official? :)

kennykerr:
@AArnott raised some questions about scalability.

There's also the question of who or how we are going to get doc comments into the IDL files for APIs and when that will happen. 

This would also have to go through @sotteson1 as he ships out platform metadata. 

We'd also have to work with DevDiv to improve the visualization as we should not be showing tooltips that refer to the implementation type within C++/WinRT.

Finally, we are moving away from any tools that are specific to Visual Studio as we don't have the resources to manage that. I would rather see a solution to this problem that is independent of the cppwinrt compiler itself. Perhaps a Visual Studio or VS Code extension that offers this capability without having to inject code into the already very large headers.

kennykerr:
Closing the PR for now. Feel free to continue the discussion on the issue: https://github.com/microsoft/cppwinrt/issues/804

PR #808, opened 2020-11-30
Support _USE_32BIT_TIME_T

kennykerr:
When `_USE_32BIT_TIME_T` is defined, `time_t` is 32-bit rather than 64-bit. This is not recommended and seldom used, but does come up from time to time. This change simply avoids a build error if this occurs by forcing the conversion.

### DIFF
--- a/strings/base_chrono.h
+++ b/strings/base_chrono.h
@@ -42,7 +42,7 @@ WINRT_EXPORT namespace winrt
 
         static time_t to_time_t(time_point const& time) noexcept
         {
-            return std::chrono::system_clock::to_time_t(to_sys(time));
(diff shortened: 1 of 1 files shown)


MikeGitb:
Is there a difference between `time_t` and `std::time_t`? I'm asking, because the return type of `std::chrono::system_clock::to_time_t` should already be `std::time_t`, so I don't see, what a cast would change otherwise.

kennykerr:
That's certainly how it's documented, but the Visual C++ libraries define it as follows:

```C++
_NODISCARD static __time64_t to_time_t(const time_point& _Time) noexcept { // convert to __time64_t
    return static_cast<__time64_t>(_Time.time_since_epoch().count() / _XTIME_TICKS_PER_TIME_T);
}
```

DefaultRyan (approved):

PR #820, opened 2020-12-13
Fix IVector<T>.GetView (and map equivalent) for .Net Native

### LINKED ISSUE #736: single_threaded_vector().GetView() causes crash when used with .NET Native.

benstevens48:
Fixes #736. Note that there are now 4 tests in old_test which fail (in produce_vector and produce_map) because they require the instance returned by GetView() to be the original instance. Should I delete these tests?

The fix works by returning a new wrapper object from GetView().

Note that my C++ is not amazingly strong and I'm unfamiliar with the structure of the C++/WinRT project, so please do check this contribution carefully.

I also had a lot of problems with the dev 'inner loop' - will raise another issue for that.

### DIFF
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -301,7 +301,7 @@ WINRT_EXPORT namespace winrt
     {
         Windows::Foundation::Collections::IVectorView<T> GetView() const noexcept
         {
-            return static_cast<D const&>(*this);
+            return make<vector_view_wrapper>(static_cast<D const&>(*this)); //Return a new wrapping object to ensure correct GetRuntimeClassName for .Net Native
         }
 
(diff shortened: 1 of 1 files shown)


kennykerr (changes requested):
While simple, this solution doubles the cost of each method call by wrapping one virtual call with another. The solution should implement the interfaces directly against the container state.

kennykerr:
I should have a PR out for this today.

PR #822, opened 2020-12-14
Musings on GetView

### LINKED ISSUE #736: single_threaded_vector().GetView() causes crash when used with .NET Native.

kennykerr:
Fixes #736 but also illustrates that its becoming increasingly difficult to make changes to C++/WinRT's stock implementations. I was always a little hesitant to add collection implementations to C++/WinRT because its so very difficult to write them in a general purpose way. But we needed something... Now that we have both single- and multi-threaded implementations and helpers for creating your own implementations, its almost impossible to make improvements without breaking something or other. 

This PR attempt manages to tear apart the implementations of `IVectorView` and `IMapView` so that they're distinct objects so that `GetRuntimeClassName` can return different values to satisfy .NET but at considerable expense in complexity and potential to destabilize. 

It also requires us to retire support for `abi_guard`. I don't actually think that's a bad thing - that feature was never used as far as I can tell. It's removal makes it easier to write generic implementations that forward their implementation efficiently. So I do like this change, but it does illustrate again how difficult it is to change the internals of C++/WinRT. A consequence of success. 😉 

Then there's the convertible `single_threaded_observable_vector` and `multi_threaded_observable_vector` implementations that implement collection interfaces for both `T` and `IInspectable` to satisfy Xaml. I haven't managed to tear off their `GetView` implementations. 

Ultimately, I don't know that we want to take this change at all. Better for anyone bothered by #736 to provide their own dedicated implementation. 

@benstevens48 also offered a simple solution here: #820. While simple, the trouble is that it impacts the performance of practically every use of the stock collections. There's also @jlaanstra's solution #743 that is also quite simple by offering alternatives. 

I also don't think that fixing #736 is going to matter in the long run. The fact that .NET crashes because `GetRuntimeClassName` returns a value it doesn't expect is a .NET bug and one that the new .NET 5 language projection doesn't (or shouldn't) suffer from.  @Scottj1s may be able to confirm. `GetRuntimeClassName` is meant to return the name of the Windows Runtime class that the object represents. Relying on what it returns in other cases is rather dubious since an implementation can naturally implement multiple interfaces and there is no hard and fast rule on what to return in those cases. Ironically, we could try to fool .NET simply by reordering the interfaces so that `IVectorView` is always listed before `IVector` but who knows whether that will work reliably.

Also adding @DefaultRyan, @dunhor, and @oldnewthing for their opinions. 

If I could turn back time I would have never introduced the helper functions for creating collections, `single_threader_vector` and all the rest, and left such things to applications and other libraries to provide.


benstevens48:
@kennykerr - should you also add
```
auto acquire_exclusive() const
{
    return m_base->acquire_exclusive();
}

auto acquire_shared() const
{
    return m_base->acquire_shared();
}
```
because although the view is read-only, the underlying collection may be updated?

I admit that #736 ought to go away in the long run, but the problem is that 'anyone bothered by #736' is potentially a large number of people. If MS are developing the Windows OS or Project Reunion using C++/WinRT, then the devs all need to be aware not to use the GetView() on an IVector created by single_threader_vector, otherwise anyone consuming the API using .Net Native is going to potentially see a crash. Unfortunately there's not going to be an alternative to .Net Native for maybe up to 2 years (which is very disappointing considering it was meant to released with .Net 5.0, and now the .Net 6.0 ambitions are quite small with statements like "Early adopters can experiment with Native AOT"). Also, it would be good to confirm that the problem does indeed go away with the new .Net 5 projections.

In conclusion, I would say that not including this fix is OK, provided it is clearly stated in the docs for single_threader_vector that there is a problem with GetView when using .Net Native, and the workaround #743 is suggested, and also that any MS developers who are producing a library using C++/WinRT that may be consumed by code compiled using .Net Native know this. On the other hand the fix looks quite clean apart from the abi_guard issues.

It's definitely a good thing that C++/WinRT contains implementations for the collections. In fact it's essential in my opinion. Otherwise it would be a significant barrier to using C++/WinRT. Not having them would mean that every single developer (not just developers of the MS libraries) would have to implement their own version, and they probably don't have the same C++ expertise as the contributors of this project.

Scottj1s:
re GetRuntimeClassName, C#/WinRT only uses it internally as an optimization - instantiating a specific projected runtime class with knowledge of all its implemented interfaces, which can be lazily QI'd and cached.  If C#/WinRT cannot resolve GetRuntimeClassName to a projected type, it simply leaves it as type-erased Object (IInspectable) and it's up to callers to cast to the appropriate interface.  The problem is with callers that are expecting a type-erased runtime class instance, and attempt to cast to that.  For better or worse, that's the precedent that was set in .NET Core WinRT support.  C#/WinRT tries to honor that where possible.

kennykerr:
That makes me think that perhaps a better solution is for GRCN to return a per-interface value instead of a per-instance value.

oldnewthing:
Issue #736 has broader impact than just .NET. Consider the case of code that wants to grant read-only access to an internal vector.

```cpp
// Things.idl
runtimeclass Thing { ... }

runtimeclass ThingManager
{
   IVectorView<Thing> Things { get; };
}

// Things.h
struct ThingManager
{
    // Return a read-only view of our internal vector.
    IVectorView<Thing> Things() const { return m_things.GetView(); }

    IVector<Thing> m_things = multi_threaded_vector<Thing>();
};
```

The current implementation allows the client to cast the `IVectorView<Thing>` back to an `IVector<Thing>` and gain access to what was supposed to be a read-only collection.

Many Windows Runtime objects use the pattern of "return a read-only view to an internal collection", with the expectation that the view cannot be promoted to read-write access.

This has serious security implications for out-of-process objects.

kennykerr:
Yes, there's merit in adding `vector_view` and `map_view` functions as #736 does for the purpose of not leaking writable access to a view. That would indirectly solve the .NET crash, so perhaps that's enough.

benstevens48:
@oldnewthing - ah I see, the security implication is probably why the GetView function exists in the first place, rather than having the IVector interface implement IVectorView. @kennykerr - that doesn't completely solve the case @oldnewthing is taking about because you might want to be able to modify the collection internally, and the solution of #743 would require returning a complete copy of the collection very time.

kennykerr:
#734 prevents writes by not implementing the `IVector` and `IMap` interfaces. This is sufficient.

oldnewthing:
743 doesn't help in the common case where the vector is dynamic, so the container needs to maintain read-write access, but it also wants to hand out read-only views. 734 lets you create a read-only view on an existing vector, but you lose the ability to mutate that vector.

For example, CoreApplication.Views returns an `IVectorView<CoreApplicationView>`. Apps can get a read-only list of views, but if the app calls `CreateNewView()` or destroys an existing view, then the `CoreApplication` needs to mutate that vector to include the new view / remove the old one. Similarly, `WebViewControl.DeferredPermissionRequests` gives read-only access to the list of active permission requests. New entries are added to the list when the web page requests a permission (and the PermissionRequested event is raised); entries are removed when the app calls the `Allow()` or `Deny()` method on a permission request.

kennykerr:
Seems like a need for a "cppwinrt collections" library - I'm just not sure we can retrofit that into cppwinrt itself nor do I think we should. I'd rather see a rich library of collections for C++/WinRT come from a dedicate library + repo.

benstevens48:
Sounds reasonable. I've no idea what the best approach is, but I'd be happy if it was a header-only library and it just meant I had to add another NuGet package reference in order to use the collections.

Issue #359, opened 2019-10-09
Delete operators that allow assignment to rvalues


kennykerr:
From internal bug report:

Rookie mistake that I’ve made when converting large chunks of C++/CX to C++/WinRT.

```
TextBlock textBlock = …;

hstring newValue = …;

textBlock.Text() = newValue; // oops, intended textBlock.Text(newValue);
```

This successfully compiles but does nothing. You are obtaining an hstring from the Text property and overwriting your local copy of the hstring with newValue. The original Text property remains unchanged.

Similarly for other smart types like com_ptr.

This can be converted to a compiler error by making the existing operator= explicitly for the lvalue reference (adding “&”) and deleting the rvalue reference version.

```
struct hstring
{
    hstring& operator=(hstring const& value) & // <- add ampersand here
    {
        m_handle.attach(impl::duplicate_string(value.m_handle.get()));
        return*this;
    }

    hstring& operator=(hstring const&) && = delete; // delete the rvalue version
}
```

Assigning to an rvalue is not particular useful. Since it isn’t named, you can’t take thee modified rvalue and do anything with it.

Okay, so maybe you could do something nuts like this

```
(textBlock.Text() = newValue).c_str()
```

but there’s no real value there. you can just remove the assignment

```
newValue.c_str()
```

or if the method you’re calling is a mutating method, you can copy the object so the method takes place on the copy.

```
hstring(newValue).c_str()
```

kennykerr:
Closing old issue as not urgent. Feel free to contribute a fix if so desired. @oldnewthing

ChrisGuzak:
someone reported a bug due to this today, so +1 to fixing this

wenyizou:
> someone reported a bug due to this today, so +1 to fixing this

Thanks for moving forward to fixing this :)

ackrumb:
Upvote. This would remove a bunch of pain from people migrating from CX. If possible, is there a way to add a static_assert in that definition that will tell them what went wrong (e.g. "this is the right way to do what you were trying to do")

kennykerr:
Feel free to contribute a PR. 😉

PR #825, opened 2020-12-16
Prevent inadvertent assignment to temporary object

oldnewthing:
Common porting error coming from languages that have properties as native concepts is trying to set a C++/WinRT property by doing

```cpp
o.Property() = value;
```

due to overzealous application of the rule "In C++/WinRT, you access a property by doing `o.Property()`."

Fix this by making all assignment operators require the assigned-to variable to be an lvalue, rendering the above mistake a compile-time error.

Assigning to an rvalue is not interesting because the rvalue has no name, so you have no way of using the assigned-to object. There are some wacky fringe cases where you might be doing this on purpose:

```cpp
get_abi(hstring() = L"Hello")
```

gives you an ABI string handle for a temporary. But this is decidedly non-idiomatic and I'm fairly confident nobody does this on purpose. The idiomatic way of writing this is simply

```cpp
get_abi(hstring(L"Hello"))
```

This fixes issue #359 

## Implementation notes

Assignment operators default to non-ref-qualified, which makes them applicable to both lvalue and rvalue references. We must explicitly redeclare them as lvalue ref-qualified in order to remove the rvalue ref-qualified version.

Once you have an explicit assignment operator, the implicit copy and move constructors disappear, so we have to bring them back explicitly.

Once you add explicit copy and move constructors, then the implicit default constructor disappears. Fortunately, we never relied upon the implicit default constructor, so we don't have to to make explicit versions.

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -2319,6 +2319,10 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
     {
         %(std::nullptr_t = nullptr) noexcept {}
         %(void* ptr, take_ownership_from_abi_t) noexcept : Windows::Foundation::IInspectable(ptr, take_ownership_from_abi) {}
+        %(% const&) noexcept = default;
+        %(%&&) noexcept = default;
+        %& operator=(% const&) & noexcept = default;
+        %& operator=(%&&) & noexcept = default;
 %%    };
 )";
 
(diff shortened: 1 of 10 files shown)


ChrisGuzak (approved):

kennykerr (approved):
Thanks!

kennykerr:
@DefaultRyan see any issues with this? It looks good to me.

DefaultRyan (approved):

DefaultRyan:
> @DefaultRyan see any issues with this? It looks good to me.

Looks good to me. Approved.

PR #829, opened 2020-12-31
Define make_guid to parse string_view to guid

tusharpm:
- Migrate logic from `generic_types.h` (test projects) to `winrt/base.h`.
- Add `_guid` string literal wrappers.
- Support for `std::wstring_view`.
- Address `std::string_view` ->`winrt::guid` parsing in #797. A version of formatting is available in `winrt::impl::to_array`.

Issue #797, opened 2020-11-19
Add winrt::guid<->string parsing/formatting


DrusTheAxe:
Would be nice to add this to C++/WinRT
https://gist.github.com/kennykerr/6c948882de395c25b3218ad8d4daf362#gistcomment-3533293

plus wchar_t/std::wstring flavors

PR #833, opened 2021-01-05
Define guid constructor to parse string_view

tusharpm:
Supersedes #829.
- Migrate logic from `generic_types.h` (test projects) to `winrt/base.h`.
- Support for `std::wstring_view`.
- Address std::string_view ->winrt::guid parsing in #797.


kennykerr (approved):
Thanks Tushar!

PR #839, opened 2021-01-12
Experimental support for design-time projection generation.

jlaanstra:
Currently devs are required to build a C++/WinRT project before intellisense works as none of the projection headers are generated until then. 

This PR hooks into the C++ design time build to generation projection headers without requiring devs to build the project as documented here: https://docs.microsoft.com/en-us/visualstudio/extensibility/visual-cpp-project-extensibility?view=vs-2019#design-time-build

Design-time builds need to be fast as to not slow down project load. It's unclear if we can achieve reasonable performance with current mdmerge and midl tooling.

The design-time build should generate the WinMD for the project as well as generate the projection headers and component source files.


kennykerr:
One month check in 😉 feel free to reopen if you intend to move forward with this.

Issue #841, opened 2021-01-22
winrt::throw_hresult should have an extension point to allow interop with WIL logging


dmachaj:
There is currently a gap when combining cppwinrt usage with WIL and it's logging functionality.  The gap is that cppwinrt exceptions that come out of winrt::throw_hresult are invisible to WIL.  This is problematic in a few ways so it would be best to fill this gap so that these libraries interop better.

To achieve that an extension point should be added to winrt::throw_hresult.  This extension point will be nullptr in normal usage.  When <wil/cppwinrt.h> is included it will wire the extension point with a function pointer that can take the exception information and log it appropriately.  (This will also require a change to WIL).

PR #842, opened 2021-01-22
Add a hook to throw_hresult for integration with WIL

dmachaj:
#841

This change is being made to improve the interop between cppwinrt and WIL (particularly WIL's logging abilities for errors).  It introduces a hook to the `winrt::throw_hresult` method that <wil/cppwinrt.h> will be able to fill in with a valid function pointer.  That function pointer will allow WIL's logging abilities to gain knowledge of the exceptions that are originated within cppwinrt.

An extern function pointer was created which is empty by default.  If it is set then `winrt::throw_hresult` will call it before continuing with actually throwing the exception.

One interesting note is that several of the parameters to `winrt_throw_hresult_handler` are not actually filled in.  There is an expectation that [C++20 source_location](https://en.cppreference.com/w/cpp/utility/source_location) will allow correct file/function/line annotation to these calls for optimum debuggability.  That hopefully isn't too far off in the future so the contract is being established now to avoid the need to change it soon in a backwards-compatible way.  Once source_location does come online there will be no changes needed on the WIL side to log them properly.

## How was the change tested?
`build_test_all.cmd` was used for this.  A new test case was added to confirm that throw_hresult does call the handler if set.  I also copied my private build of cppwinrt.exe to a test project to confirm that the generated base.h matches expectations.

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -426,6 +426,11 @@ WINRT_EXPORT namespace winrt
 
     [[noreturn]] inline __declspec(noinline) void throw_hresult(hresult const result)
     {
+        if (winrt_throw_hresult_handler)
+        {
+            winrt_throw_hresult_handler(0, nullptr, nullptr, _ReturnAddress(), result);
+        }
+
         if (result == impl::error_bad_alloc)
         {
             throw std::bad_alloc();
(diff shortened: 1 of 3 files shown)


kennykerr:
Looks good, do we have a commitment form the WIL folks that they will take this change as is? Just want to make sure we're not adding complexity without a taker. It has been well over a year since the last WIL hook was added and it is still in the pipeline for WIL integration. 

Also, in order to allow the CI build to start you need to submit the PR from a local branch. I've added you to the repo so you have permission to create a branch here. Sorry about that.

dmachaj:
I'm going to be making the change on both sides so there shouldn't need to be a worry about that. I'm staging the WIL changes now that the cppwinrt half is out for review.  I'm going to need the cppwinrt version number with this change in place so that the WIL side can constexpr check against it.

PR #857, opened 2021-02-15
to_message() function extracts exception text from catch(...)

oldnewthing:
Handy for logging. This complements to_hresult().

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -546,6 +546,31 @@ WINRT_EXPORT namespace winrt
         }
     }
 
+    inline __declspec(noinline) hstring to_message() noexcept
+    {
+        if (winrt_to_message_handler)
+        {
+            return winrt_to_message_handler(WINRT_IMPL_RETURNADDRESS());
+        }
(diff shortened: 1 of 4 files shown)


kennykerr:
Is the intention to log an error message and then proceed in cases where an exception is otherwise ignored?

oldnewthing:
Yeah, it's common to catch all errors, log it, and then move on. Whether it's advisable is another matter, but it's a common pattern, and too often people just catch `hresult_error` and forget to catch `std::exception` so they end up missing standard exceptions, most notably `std::bad_alloc` (which is what `E_OUTOFMEMORY` projects to).

kennykerr:
Cool. And the motivation to add another WIL hook?

kennykerr (approved):
Oh to support custom exception types.

kennykerr:
So for some reason, the CI build won't build anything other than local branches. I trust you've fully tested this, but next time please use a local branch on https://github.com/microsoft/cppwinrt.

PR #858, opened 2021-02-22
Add async version of winrt::event

### LINKED ISSUE #724: Can winrt::event deliver invocations asynchronously?
Currently, `winrt::delegate<T>::operator()` delivers its invocations on the calling thread. This can be problematic if the delegate _implementation_ is not reentrant, or if the invocation is advisory-only.  There's an internal implementation of a WIL type that creates a separate thread to deliver the invoke calls.  While this can lead to unnecessary thread creation, supporting a mode of "deliver this invoke async" as part of the delegate's traits directly would be useful.

Implementations can do this themselves, however:

```c++
template<typename Q> fire_and_forget do_async(Q&& action) 
{
    typename remove_reference<Q>::type stored { std::move(action) };
    co_await resume_background();
    stored();
}
void MyType::SetValue(hstring const& value) 
{
    {
        auto lock = lock_self();
        m_value = value;
    }
    do_async([self = get_strong()] {
        m_valueUpdatedEvent(self, m_value);
    });
}
```

kennykerr:
Fixes #724 

Unlike `winrt::event` which is completely serial and synchronous, `windows::async_event` invokes its handlers in parallel and awaits them all, and you can await the aggregate or proceed without blocking. Calling in parallel means that one stuck handler does not prevent the other handlers from being called.

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -801,4 +801,43 @@ WINRT_EXPORT namespace winrt
         impl::check_status_canceled(shared->status);
         co_return shared->result.GetResults();
     }
+
+    template <typename Delegate>
+    struct async_event : impl::event_base<Delegate>
+    {
+        async_event() = default;
+
+        template<typename...Arg>
+        Windows::Foundation::IAsyncAction operator()(Arg const... args)
+        {
+            auto targets = this->make_temp_targets();
+
+            if (!targets)
+            {
+                co_return;
+            }
+
+            std::vector<Windows::Foundation::IAsyncAction> workers;
+            workers.reserve(targets->size());
+
+            for (auto&& element : *targets)
+            {
+                workers.push_back(invoke(element, args...));
+            }
+
+            for (auto&& element : workers)
+            {
+                co_await element;
+            }
+        }
+
+    private:
+
+        template <typename... Args>
+        Windows::Foundation::IAsyncAction invoke(Delegate const& element, Args const&... args)
+        {
+            co_await resume_background();
+            impl::invoke(element, args...);
+        }
+    };
 }
--- a/test/test/event.cpp
+++ b/test/test/event.cpp
@@ -1,0 +1,98 @@
+#include "pch.h"
+
+using namespace winrt;
+using namespace Windows::Foundation;
+
+namespace
+{
+    struct Sender : implements<Sender, IStringable>
+    {
+        hstring m_value{ L"Sender" };
+
+        hstring ToString()
+        {
+            return m_value;
+        }
+    };
+}
+
+TEST_CASE("event")
+{
+    handle start{ CreateEventW(nullptr, true, false, nullptr) };
+    handle completed{ CreateEventW(nullptr, true, false, nullptr) };
+
+    auto sender = make<Sender>();
+    async_event<TypedEventHandler<IStringable, int>> event;
+
+    // No handlers and the resulting IAsyncAction completes synchronously (without blocking).
+    REQUIRE(event(sender, 123).Status() == AsyncStatus::Completed);
+
+    auto token = event.add([&](auto&& sender, auto&& args)
+        {
+            // Lengthy timeout allowing test thread to signal start.
+            REQUIRE(WaitForSingleObject(start.get(), 1000) == WAIT_OBJECT_0);
+            REQUIRE(sender.ToString() == L"Sender");
+            REQUIRE(args == 123);
+            REQUIRE(SetEvent(completed.get()));
+        });
+
+    auto result = event(sender, 123);
+
+    // Short sleep allowing buggy code (if any) to race ahead.
+    Sleep(100);
+
+    // Result should not Completed yet as the start event is not yet set.
+    REQUIRE(result.Status() == AsyncStatus::Started);
+
+    // Race to completion.
+    REQUIRE(SetEvent(start.get()));
+
+    // Can now wait for result (and throw on failure).
+    result.get();
+
+    // Zero timeout checks that the event is already signaled.
+    REQUIRE(WaitForSingleObject(completed.get(), 0) == WAIT_OBJECT_0);
+
+    event.remove(token);
+
+    // A single event is raied with a number of handlers.
+    std::atomic<int> counter;
+
+    for (int i = 0; i != 10; ++i)
+    {
+        event.add([=, &counter](auto&&...)
+            {
+                counter += i;
+            });
+    }
+
+    event(sender, 0).get();
+    REQUIRE(counter == 45);
+
+    // A series of events are raised with a growing set of handlers.
+    async_event<TypedEventHandler<IStringable, int>> stress;
+    counter = 0;
+    std::vector<IAsyncAction> results;
+
+    for (int j = 0; j != 10; ++j)
+    {
+        for (int i = 0; i != 10; ++i)
+        {
+            stress.add([=, &counter](auto&&...)
+                {
+                    counter += i;
+                });
+        }
+
+        // Don't wait for results before raising next event.
+        results.push_back(stress(sender, 0));
+    }
+
+    // Finally wait for results and check.
+    for (auto&& async : results)
+    {
+        async.get();
+    }
+
+    REQUIRE(counter == 2475);
+}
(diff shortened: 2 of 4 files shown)


strings/base_coroutine_foundation.h
```diff
@@ -801,4 +801,43 @@ WINRT_EXPORT namespace winrt
...
+            }
+        }
+
+    private:
+
+        template <typename... Args>
+        Windows::Foundation::IAsyncAction invoke(Delegate const& element, Args const&... args)
+        {
+            co_await resume_background();
+            impl::invoke(element, args...);
```
oldnewthing, line 840: But this doesn't have the "auto-revoke on disconnected" behavior that the sync event supports. Shouldn't it return `IAsyncOperation<bool>` and `co_return impl::invoke(element, args...)` and then the `co_await` loop up above would check if the `invoke` failed and if so, revoke the token for the element?

strings/base_coroutine_foundation.h
```diff
@@ -801,4 +801,43 @@ WINRT_EXPORT namespace winrt
         impl::check_status_canceled(shared->status);
         co_return shared->result.GetResults();
     }
+
+    template <typename Delegate>
+    struct async_event : impl::event_base<Delegate>
```
oldnewthing, line 806: The name `async_event` may be too ambiguous. I can think of six possible meanings!

1. This is a an object that you `co_await` and wait for it to be signaled. Sort of like an "awaitable event" (where "event" is used in the kernel sense).
2. This is a WinRT event which invokes its callbacks sequentially on a background thread. Used if you want to protect the main thread from hanging on a stuck handler. You can await the initiation of the handlers (but cannot await their completions).
3. This is a WinRT event which invokes its callbacks in parallel on background threads. Used if you want to protect one stuck handler from interfering with all the other handlers.  You can await the initiation of the handlers (but cannot await their completions).
4. This is a WinRT event which invokes its callbacks sequentially on the main thread, then awaits them all, and you can await the aggregate.
5. This is a WinRT event which invokes its callbacks sequentially on a background thread, then awaits them all, and you can await the aggregate.
6. This is a WinRT event which invokes its callbacks in parallel on a background thread, then awaits them all, and you can await the aggregate. Calling in parallel means that one stuck handler does not prevent the other handlers from receiving the callback.

It looks like this is option 3.

strings/base_coroutine_foundation.h
oldnewthing, line 840: This smuggles the delegate, which will not work with JavaScript or any other caller who passed a non-agile delegate.

kennykerr: The trouble is I can't easily guarantee the event source still lives.

kennykerr: All delegates are agile at this point either directly or wrapped.

strings/base_coroutine_foundation.h
kennykerr: @jonwis asked for an async version of `winrt::event` - this is what I came up with. It allows handlers to execute in parallel to avoid any single handler blocking progress on the rest. It also allows the event source to continue without waiting for all the handlers to execute, although it does allow the event source to wait for their completion.

kennykerr: See https://github.com/microsoft/cppwinrt/issues/724#issuecomment-783562037 for an example.

oldnewthing, line 806: Trying to think of a clearer name. Maybe `background_invoked_event`? The point is that the event handlers are invoked on a (new) background thread.

oldnewthing (approved):

strings/base_coroutine_foundation.h
```diff
@@ -801,4 +801,43 @@ WINRT_EXPORT namespace winrt
...
+                co_await element;
+            }
+        }
+
+    private:
+
+        template <typename... Args>
+        Windows::Foundation::IAsyncAction invoke(Delegate const& element, Args const&... args)
+        {
+            co_await resume_background();
```
jonwis, line 839: I still think you need to let the user of the async_event determine whether there's per-delegate threads launched, or if all the delegates are fired on a single thread. My recollection is that WRL's async-event does all-on-single-new-thread, so someone coming from that space might be surprised when there's delegates competing on calling back during their operations.

test/test/event.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    auto sender = make<Sender>();
+    async_event<TypedEventHandler<IStringable, int>> event;
+
+    // No handlers and the resulting IAsyncAction completes synchronously (without blocking).
+    REQUIRE(event(sender, 123).Status() == AsyncStatus::Completed);
+
+    auto token = event.add([&](auto&& sender, auto&& args)
+        {
+            // Lengthy timeout allowing test thread to signal start.
```
jonwis, line 32: Why bother waiting at all?

jonwis (approved):
Thanks! Still a little concerned about where the parallelism goes, but looks great!

strings/base_coroutine_foundation.h
kennykerr: This can be achieved with the existing synchronous `winrt::event` - it executes everything in series but is fully thread-safe so you can simply call it in parallel.

test/test/event.cpp
kennykerr: This ensures that the test can observe the the status of the async transitions from the Started to Completed states.

strings/base_coroutine_foundation.h
oldnewthing: So for "execute in series on background thread", the pattern is
```cpp
auto lifetime = get_strong(); // prevent premature destruction
co_await resume_background();
m_event.invoke(*this, args);
```
?

kennykerr: Yes, that would work.

ChrisGuzak: so glad to see this. we can finally stop using wrlasyncevent.h

strings/base_coroutine_foundation.h
kennykerr: It's probably safe because the "sender" is likely the strong reference we need to make it so, but there's no way to guarantee this.

strings/base_coroutine_foundation.h
kennykerr: If you guys are happy with "execute in series on background thread" then I can close this PR and avoid introducing a new event type.

strings/base_coroutine_foundation.h
```diff
@@ -801,4 +801,43 @@ WINRT_EXPORT namespace winrt
...
         co_return shared->result.GetResults();
     }
+
+    template <typename Delegate>
+    struct async_event : impl::event_base<Delegate>
+    {
+        async_event() = default;
+
+        template<typename...Arg>
+        Windows::Foundation::IAsyncAction operator()(Arg const... args)
```
kennykerr, line 811: There's also a usability issue here. Because the args are necessarily `const` rather than `const&`, the callsite cannot simply use `m_event(*this, args)` but must explicitly cast as in `m_event(static_cast<SenderType>(*this), args)`.

oldnewthing: (sad face)

kennykerr: Another approach is to have `async_event`'s `operator()` be a blocking call that runs the handlers in parallel. If the caller wants to avoid blocking the call itself they can simply run it on a background thread like you can with `event`. That would solve both this usability problem and the issue of removing dead handlers.

oldnewthing: Maybe at the end of the day we don't need this at all. Callers can do

```cpp
[](auto lifetime, auto args) -> winrt::fire_and_forget
{
    winrt::resume_background();
    lifetime->m_event(*lifetime, args);
}(get_strong(), args);
```

kennykerr: Right, but that would still allow one long-running handler to delay other handlers from running. I just don't know whether that's a scenario anyone cares about.

strings/base_coroutine_foundation.h
oldnewthing: WRL-based AsyncEvent raised all events serialized on a single background thread, so I don't think we need the parallel invoke.

kennykerr: Then my work here is done. 😉

PR #860, opened 2021-02-25
To support local servers, use .exe files from dependent projects as Implementation in Microsoft.Windows.CppWinRT.targets

ChrisGuzak:
I found that if I create a WinRT local server project, a dependent app project won't build due to a missing file referenced in the .appxmanifest (the WinRT local server registration information that I added). The error lead me to this line in the build system that supports `.dll` dependents, but not `.exe`s. Adding the `.exe` test fixes this problem and seems appropriate (experts please advise).

With this change a warning is still produced, it seems to to missing support in the .appxmanifest generation support.
```
warning APPX1708: The executable 'WinRtLocalServer.exe' is specified as the implementation for the .winmd file 'C:\Users\chris\source\repos\ChrisGuzakWork\WinRtLocalServerClient\Debug\WinRtLocalServer.winmd'. 
Only in-process servers are supported for generating registration information in the app manifest. You must specify the out-of-process server registration information in the app manifest.
```
This can be ignored.

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -253,7 +253,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <WinMDFullPath Include="@(_CppWinRTProjectWinMDItems->FullPath()->Distinct()->ClearMetadata())" Condition="'$(CppWinRTGenerateWindowsMetadata)' == 'true'">
                 <TargetPath>$([System.IO.Path]::GetFileName('$(CppWinRTProjectWinMD)'))</TargetPath>
                 <Primary>true</Primary>
-                <Implementation Condition="'$(TargetExt)' == '.dll'">$(WinMDImplementationPath)$(TargetName)$(TargetExt)</Implementation>
+                <Implementation Condition="'$(TargetExt)' == '.dll' or '$(TargetExt)' == '.exe'">$(WinMDImplementationPath)$(TargetName)$(TargetExt)</Implementation>
                 <FileType>winmd</FileType>
                 <WinMDFile>true</WinMDFile>
                 <ProjectName>$(MSBuildProjectName)</ProjectName>


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -253,7 +253,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <WinMDFullPath Include="@(_CppWinRTProjectWinMDItems->FullPath()->Distinct()->ClearMetadata())" Condition="'$(CppWinRTGenerateWindowsMetadata)' == 'true'">
                 <TargetPath>$([System.IO.Path]::GetFileName('$(CppWinRTProjectWinMD)'))</TargetPath>
                 <Primary>true</Primary>
-                <Implementation Condition="'$(TargetExt)' == '.dll'">$(WinMDImplementationPath)$(TargetName)$(TargetExt)</Implementation>
+                <Implementation Condition="'$(TargetExt)' == '.dll' or '$(TargetExt)' == '.exe'">$(WinMDImplementationPath)$(TargetName)$(TargetExt)</Implementation>
```
jlaanstra, line 256: VS never supported this for CX and so when support for WinRT was added we added the same limitation. We could possibly remove the TargetExt check alltogether. Not sure if you would run into any other roadblocks with that.

jlaanstra (commented):
Kenny has indicated that he wants build changes to be validated on the OS, WinUI and Terminal before taking them.

nuget/Microsoft.Windows.CppWinRT.targets
ChrisGuzak: local servers are commonly implemented as a .exe (Win32 service is the other option).

ChrisGuzak: @jlaanstra and @kennykerr do you have advice on doing an OS and terminal build? Given this did not work before perhaps this is safe enough to not require that (me trying to shirk this work), let me know if you think so.

kennykerr: 😉 I do like the fact that it is a small and targeted change. We've just had a lot of problems with seemingly innocuous changes to the msbuild support that have had surprising compatibility issues in different environments. @jlaanstra has done this validation a few times - he may have some tips to quickly smoke out any issues.

jlaanstra: @ChrisGuzak the way I generally go about this is to replace the CppWinRT props/targets used by a project and then do a full build.  For the OS this means replacing the files under tools\cppwinrt (I can share a script I use to build a few known WinRT directories) and for terminal it means replacing the files in the expanded NuGet package under the packages folder and building terminal per instructions in their repo.

DHowett:
Terminal confirms success.

nuget/Microsoft.Windows.CppWinRT.targets
ChrisGuzak: Terminal has been validated. working on os.2020 next.

ChrisGuzak (commented):
merging into this branch so I can create a C++ WinRT package for testing

PR #863, opened 2021-02-26
Add ProjectCapability for VS telemetry

DefaultRyan:


kennykerr:
Can you explain what this does? This will also have to be validated more carefully. https://github.com/microsoft/cppwinrt/pull/860#pullrequestreview-598064436

jlaanstra:
My understanding is that this actually doesn't affect the build at all but can be used by the project system to detect that a project has the CppWinRT capability. VS could potentially customize the project system based on this and we would have no idea, but it seems like this is just for telemetry based on the title.

BenJKuhn (approved):

jlaanstra (approved):

PR #867, opened 2021-03-04
Clang workarounds

kennykerr:
* Versions of STL typically used with Clang don't corrently attribute `std::terminate` with `[[noreturn]]`.

* Clang incorrectly fails to see the static constexpr member.

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -179,7 +179,7 @@ namespace winrt::impl
 
     [[noreturn]] inline void __stdcall fallback_RoFailFastWithErrorContext(int32_t) noexcept
     {
-        std::terminate();
+        abort();
     }
 }
 
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -82,6 +82,50 @@ WINRT_EXPORT namespace winrt
 
     struct guid
     {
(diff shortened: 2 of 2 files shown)


strings/base_error.h
```diff
@@ -179,7 +179,7 @@ namespace winrt::impl
 
     [[noreturn]] inline void __stdcall fallback_RoFailFastWithErrorContext(int32_t) noexcept
     {
-        std::terminate();
+        abort();
```
oldnewthing, line 182: Also at line 570 (`to_message`) of `base_error.h` and line 39 (`hex_to_uint`) of `base_types.h`? Otherwise, clang may complain "Not all code paths return a value."

kennykerr: I didn't change those as Clang figured them out and didn't complain. We could switch everything over to `abort` or just wait for Clang to get with the standard.

oldnewthing (approved):

strings/base_error.h
oldnewthing, line 182: I'm somewhat surprised that we don't have to say `std::abort()` here, but obviously it's working somehow.

PR #868, opened 2021-03-04
abort

kennykerr:
Switched remaining `std::terminate` calls to `abort` - more exhaustive testing with Clang still runs into issues in some dev environments.


ChrisGuzak:
Why did you change from `std::terminate()` -> `abort()`, is the later more diagnosable?

kennykerr:
No, terminate just calls abort. The latter has better support from the clang tool chain.

PR #870, opened 2021-03-05
Allow IAsync...WithProgress to report intermediate results

oldnewthing:
Clients of IAsync...WithProgress are permitted to call GetResults() before the operation has completed, allowing them to observe partial results. This behavior is necessary because the Progress type must be a value type, but the operation may want to report reference types in their progress reports. The way to do that is to report the reference type as the GetResults().

The coroutine itself can use the `set_result()` method on the progress token to report a result prior to completion, and can set a new intermediate result as often as desired prior to the completion of the coroutine. The value passed to `co_return` acts as the final result of the coroutine. If `set_result()` is never called, then the intermediate result is an empty result (null for reference types, default-initialized for value types).

In practice, the result of the coroutine is usually a reference type, and the common pattern for coroutines that want to report a reference type as progress is to create the result object, pass it to `set_result()` at the start of the coroutine, update the result object as the coroutine progresses, and (redundantly) pass the same result object to `co_return` when the coroutine completes.

```cpp
IAsyncOperationWithProgress<Muffin, double> BakeMuffinAsync()
{
    auto progress = co_await get_progress_token();

    Muffin muffin;
    Muffin.Diameter(5);
    progress.set_result(muffin);
    progress(0.0);

    muffin.Diameter(5.2);
    progress(0.5);

    muffin.Diameter(6);
    progress(1.0);

    co_return muffin;
}
```

Since clients of ...WithProgress are permitted to call GetResults() prior to completion, this introduces a few new wrinkles.

1. We need to use the lock to ensure a client doesn't try to read a partial result at the same time we are writing a new one.
2. The GetResults() method must return a copy of the result. This is true even if the coroutine has completed, because a progress handler might go off and do some asynchronous work, and then come back and ask for the intermediate result. If GetResults() on a completed coroutine consumes the result, then the async progress handler and the co_await will compete to retrieve the completed results, and somebody will lose.

As a result, ...WithProgress coroutines are slightly more expensive than non-progress coroutines due to the extra locking and copying. Fortunately, it's mostly pay-for-play. If a coroutine never calls `set_result()`, and the client never calls `GetResults()` from its progress handler, then there is no new lock contention. The final copy is unavoidable, but fortunately, the result is usually a reference type, so you just pay an extra AddRef/Release pair.

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -338,6 +338,13 @@ namespace winrt::impl
             m_promise->set_progress(result);
         }
 
+        template<typename T>
+        void set_result(T&& value) const
+        {
+            static_assert(!std::is_same_v<Progress, void>, "Setting preliminary results requires IAsync...WithProgress");
+            m_promise->return_value(std::forward<T>(value));
+        }
+
     private:
 
         Promise* m_promise;
@@ -454,9 +461,20 @@ namespace winrt::impl
         {
             slim_lock_guard const guard(m_lock);
 
-            if (m_status.load(std::memory_order_relaxed) == AsyncStatus::Completed)
+            if constexpr (std::is_same_v<TProgress, void>)
             {
-                return static_cast<Derived*>(this)->get_return_value();
+                if (m_status.load(std::memory_order_relaxed) == AsyncStatus::Completed)
+                {
+                    return static_cast<Derived*>(this)->get_return_value();
+                }
+            }
+            else
+            {
+                if (m_status.load(std::memory_order_relaxed) == AsyncStatus::Completed ||
+                    m_status.load(std::memory_order_relaxed) == AsyncStatus::Started)
+                {
+                    return static_cast<Derived*>(this)->copy_return_value();
+                }
             }
 
             rethrow_if_failed();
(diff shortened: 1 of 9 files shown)


kennykerr (approved):
Looks good - thanks!

strings/base_coroutine_foundation.h
```diff
@@ -454,9 +461,20 @@ namespace winrt::impl
...
+            {
+                if (m_status.load(std::memory_order_relaxed) == AsyncStatus::Completed)
+                {
+                    return static_cast<Derived*>(this)->get_return_value();
+                }
+            }
+            else
             {
-                return static_cast<Derived*>(this)->get_return_value();
+                if (m_status.load(std::memory_order_relaxed) == AsyncStatus::Completed ||
```
jonwis, line 473: Is it OK to load twice here?

jonwis (approved):

strings/base_coroutine_foundation.h
kennykerr: Ah, I didn't spot that. It's safe but I think it would be cleaner and cheaper to load once and compare the result.

oldnewthing: Ooh, good catch. Should load once and test the cached copy. (std::atomic.load() forces a reach back to memory each you load.)

oldnewthing: Turns out there are a lot of places where I perform redundant loads from the atomic.

PR #873, opened 2021-03-07
Ranges support and begin/end members

### LINKED ISSUE #807: No ranges support for collection types?
Hi!

Just to confirm. For the collection types, e.g. IVector, IMap. Currently there is no c++20 ranges support, right?

sylveon:
This enables WinRT collections to be used with C++20 ranges.

I added member `begin`/`end` because the existing solution (ADL-only) doesn't work with `std::ranges::begin`'s ADL resolution - it defines a deleted template `begin`, and this template is a better match than C++/WinRT's ADL functions.

Fixes #807


sylveon:
Example:
```cpp
template<typename T>
class winrt_view {
private:
    const T& collection;

public:
    winrt_view(const T& coll) noexcept : collection(coll) { }

    auto begin() const noexcept(noexcept(winrt::impl::begin(collection)))
    {
        return winrt::impl::begin(collection);
    }

    auto end() const noexcept(noexcept(winrt::impl::end(collection)))
    {
        return winrt::impl::end(collection);
    }
};

template<typename T>
inline constexpr bool std::ranges::enable_borrowed_range<winrt_view<T>> = true;

int main()
{
    auto a = winrt::single_threaded_vector<int>({ 1, 2, 3 });

    winrt_view(a) | std::views::reverse;
}
```

strings/base_iterator.h
```diff
@@ -0,0 +1,183 @@
...
+    template <typename T>
+    struct fast_iterator
+    {
+        using iterator_concept = std::random_access_iterator_tag;
+        using iterator_category = std::input_iterator_tag;
+        using value_type = decltype(std::declval<T>().GetAt(0));
+        using difference_type = ptrdiff_t;
+        using reference = value_type;
+
+        fast_iterator() noexcept : m_collection(nullptr), m_index(0) {}
```
miscco, line 13: Could this use member initializer and then default the constructor?

strings/base_iterator.h
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        reference operator[](difference_type n) const
+        {
+            return m_collection->GetAt(m_index + static_cast<uint32_t>(n));
+        }
+
+        bool operator==(fast_iterator const& other) const noexcept
+        {
+            WINRT_ASSERT(m_collection == other.m_collection);
```
miscco, line 84: Should this assert be applied to `operator-` and friends?

strings/base_iterator.h
```diff
@@ -0,0 +1,183 @@
...
+            return m_index < other.m_index;
+        }
+
+        bool operator!=(fast_iterator const& other) const noexcept
+        {
+            return !(*this == other);
+        }
+
+        bool operator>(fast_iterator const& other) const noexcept
+        {
```
miscco, line 100: The fact that `operator>` is identical to `operator>=` suggests a Bug with missing swap of arguments

sylveon:
I'm actually thinking it might be possible to enable ranges for `IIterable` as well - they do require post-fix ++ but put no constraint on its return type, so it would be possible to write a post-fix ++ returning `void`. Thoughts?

sylveon:
With my latest commit, every `IIterable` can also be used with ranges:
```cpp
// prints "1 3 5 7"
auto v = winrt::single_threaded_vector<int>({ 1, 2, 3, 4, 5, 6, 7 });
winrt::Windows::Foundation::Collections::IIterable<int> c(v);

for (int i : c | std::views::filter([](int i) { return i % 2 != 0; }))
{
    std::cout << i << ' ';
}
```

strings/base_iterator.h
sylveon: That was indeed a bug copied over from the existing implementation, resolved.

strings/base_iterator.h
sylveon: I see no reason not to, applied.

strings/base_iterator.h
sylveon: Indeed, done

kennykerr:
Thanks for the contribution - @DefaultRyan can you take a look?

DefaultRyan:
I should be able to get to it in a day or two. I haven't been doing my homework reading up on the details of these newfangled ranges, but from what I've heard, this should be exciting.

DefaultRyan:
> I'm actually thinking it might be possible to enable ranges for `IIterable` as well - they do require post-fix ++ but put no constraint on its return type, so it would be possible to write a post-fix ++ returning `void`. Thoughts?

I'm not quite convinced. Which concept is it that needs to be satisfied?

sylveon:
> Which concept is it that needs to be satisfied?

[std::weakly_incrementable](http://en.cppreference.com/w/cpp/iterator/weakly_incrementable)

sylveon:
To be clear, it's `std::incrementable` which requires `i++` to return `decltype(i)`. `std::weakly_incrementable` just wants `i++` to be a valid expression and it can return nothing for all it cares.

`std::input_iterator` only requires `std::weakly_incrementable`, while `std::forward_iterator` wants `std::incrementable`, so this makes `IIterator<T>` only comply to `std::input_iterator` (currently, it also only presents itself as an input iterator for the same post-fix ++ reason, nothing changes here). As explained in cppreference, algorithms over `std::input_iterator` should be single-pass algorithms only as well.

Making `i++` return void will also emit a compiler error in case someone tries to do something like `IIterator<T> last = i++`, so I think it's a fair trade to enable ranges: it doesn't break any code, since post-fix ++ on `IIterator<T>` doesn't exist currently, and makes sure that nobody accidentally writes wrong code using this added operator.

sylveon:
Any update on this? I'd like to start using ranges in my projects but this is a pretty big blocker.

kennykerr:
@DefaultRyan will hopefully get a chance to take a look soon. 

I had a quick look and the obvious missing piece here is good test coverage.

sylveon:
I was thinking the same but I'm not sure how to proceed: do I create a brand new test project with C++20 or upgrade the existing one to C++20?

kennykerr:
Yes, we don't want to destabilize the existing tests. Probably best to create a new test project by copying "test", removing the existing tests, and then flipping it to C++20.

sylveon:
Just added a few unit tests, and made sure they pass locally. The IDE says some constraints aren't satisfied when editing the ranges.cpp file but it builds and runs, so it seems like some EDG bugs that should be reported.

kennykerr:
Thanks Charles, I had a quick look and this is sufficiently disruptive that I'll need to run a CI build to validate the changes. Unfortunately I can only do so against branches on https://github.com/microsoft/cppwinrt - so if you don't mind please push a branch directly and reopen a PR. I have given you Write access in order to do that. I would also do a complete build and test before pushing as I did notice a few issues with some of the other test projects indicating regressions.

sylveon:
I pushed my current changes to `user/sylveon/ranges`, will do a full test run and fix anything that crops up this evening. I'll open a new PR once it's done :)

PR #881, opened 2021-03-11
Use app manifest to enable long path support

asklar:
In #880 we had a hacky way to enable long path support; here we do it via manifest
CC @oldnewthing


asklar:
oops wrong remote

PR #882, opened 2021-03-11
use app manifest

asklar:
In #880 we had a hacky way to enable long path support; here we do it via manifest
CC @oldnewthing


kennykerr:
I don't see where you've actually backed out the undocumented API call (and #include).

asklar:
thanks, good catch; fixed, I had messed up my clone when I migrated from my fork to the main trunk and this file hadn't gotten added.

oldnewthing (commented):
Belated 👍

PR #893, opened 2021-03-17
Use compiler-specific WINRT_IMPL_NOINLINE macro for noinline functions

oldnewthing:
Different toolchains use different ways of marking a function as not inlinable. Wrap it inside a macro so we can accommodate the differences.

Fortunately, all of them insert the marker at the same place (between the "inline" keyword and the return type).

### DIFF
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -45,6 +45,14 @@
 #undef _WINDOWS_NUMERICS_END_NAMESPACE_
 #endif
 
+#if defined(_MSC_VER)
+#define WINRT_IMPL_NOINLINE __declspec(noinline)
+#elif defined(__GNUC__)
+#define WINRT_IMPL_NOINLINE __attribute__((noinline))
+#else
+#define WINRT_IMPL_NOINLINE
+#endif
+
 #ifdef __IUnknown_INTERFACE_DEFINED__
 #define WINRT_IMPL_IUNKNOWN_DEFINED
 #endif
--- a/strings/base_activation.h
+++ b/strings/base_activation.h
@@ -26,7 +26,7 @@ namespace winrt::impl
 
 
     template <bool isSameInterfaceAsIActivationFactory>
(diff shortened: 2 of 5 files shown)


strings/base_macros.h
```diff
@@ -45,6 +45,14 @@
 #undef _WINDOWS_NUMERICS_END_NAMESPACE_
 #endif
 
+#if defined(_MSC_VER)
+#define WINRT_IMPL_NOINLINE __declspec(noinline)
+#elif defined(__GNUC__)
+#define WINRT_IMPL_NOINLINE __attribute__((noinline))
+#else
+#define WINRT_IMPL_NOINLINE
+#endif
```
kennykerr, line 54: How does this cover Clang? I believe clang already know how to interpret `__declspec(noinline)`. I wouldn't want to degrade Clang inadvertently.

oldnewthing: Clang defines `__GNUC__` and supports GNU-style `__attribute__((noinline))`. Clang has a (non-default) option to support MSVC-style `__declspec(noinline)` as well. This version removes the need to add the `-fdeclspec` flag when compiling with clang.

kennykerr (approved):
Thanks!

strings/base_macros.h
oldnewthing: The breakdown is

| Toolchain | Special macros | `__declspec(noinline)` | `__attribute__((noinline))`
|-|-|-|-|
| MSVC | `_MSC_VER` | ☑ | | |
| Clang (default) | `__GNUC__`<br>`__clang__` | | ☑ |
| Clang (`-fdeclspec`) | `__GNUC__`<br>`__clang__` | ☑ | ☑ |
| gcc | `__GNUC__` |  | ☑ |
| icc | `__GNUC__`<br>`__INTEL_COMPILER` |  | ☑ |
| icx | `__GNUC__`<br>`__INTEL_LLVM_COMPILER` | ☑ | ☑ |

djgpp, ellcc and zapcc are based on one of the above toolchains.

So sniffing for `_MSC_VER` and `__GNUC__` covers all the major toolchains.

kennykerr: Great!

PR #895, opened 2021-03-20
Restore Win7 compat by using old versions of threadpool functions

### LINKED ISSUE #894: SetThreadpoolWaitEx and SetThreadpoolTimerEx are not available in Windows 7
IIRC C++/WinRT guarantee support Windows 7, but it uses `SetThreadpoolWaitEx` and `SetThreadpoolTimerEx`, which are available since Windows 8. Is it a bug?

https://github.com/microsoft/cppwinrt/blob/d61cb159be7b0a31ae8e82970aabedf9fcb2fda9/strings/base_coroutine_threadpool.h#L436-L443

oldnewthing:
Fortunately, the Ex versions behave the same as the Win7 versions, so we can just switch everybody to the Win7 versions.

Fixes #894

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -414,7 +414,7 @@ WINRT_EXPORT namespace winrt
                 m_handle = handle;
                 m_timer.attach(check_pointer(WINRT_IMPL_CreateThreadpoolTimer(callback, this, nullptr)));
                 int64_t relative_count = -m_duration.count();
-                WINRT_IMPL_SetThreadpoolTimerEx(m_timer.get(), &relative_count, 0, 0);
+                WINRT_IMPL_SetThreadpoolTimer(m_timer.get(), &relative_count, 0, 0);
 
                 state expected = state::idle;
                 if (!m_state.compare_exchange_strong(expected, state::pending, std::memory_order_release))
@@ -435,10 +435,10 @@ WINRT_EXPORT namespace winrt
 
             void fire_immediately() noexcept
             {
-                if (WINRT_IMPL_SetThreadpoolTimerEx(m_timer.get(), nullptr, 0, 0))
+                if (WINRT_IMPL_SetThreadpoolTimer(m_timer.get(), nullptr, 0, 0))
                 {
                     int64_t now = 0;
-                    WINRT_IMPL_SetThreadpoolTimerEx(m_timer.get(), &now, 0, 0);
+                    WINRT_IMPL_SetThreadpoolTimer(m_timer.get(), &now, 0, 0);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


kennykerr (approved):
Sweet!

strings/base_coroutine_threadpool.h
```diff
@@ -435,10 +435,10 @@ WINRT_EXPORT namespace winrt
 
             void fire_immediately() noexcept
             {
-                if (WINRT_IMPL_SetThreadpoolTimerEx(m_timer.get(), nullptr, 0, 0))
+                if (WINRT_IMPL_SetThreadpoolTimer(m_timer.get(), nullptr, 0, 0))
```
ysc3839, line 438: According to [SetThreadpoolTimer docs](https://docs.microsoft.com/en-us/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-setthreadpooltimer), it returns void instead of BOOL, is it OK to use this return value?

kennykerr: Well spotted - @oldnewthing I'm guessing this means we don't have test coverage for this one. 😟

oldnewthing: Strange, because [there's a cancellation test](https://github.com/microsoft/cppwinrt/blob/1502e293d6e8f0a2c017431f6cf186d1551202fe/test/test/async_propagate_cancel.cpp#L76). Will have to figure out how this slipped through. Probably because I messed up.

oldnewthing: Okay, the deal is that this works on Win8+ due to how SetThreadpoolTimer is implemented (it calls SetThreadpoolTimerEx and throws away the result, but the result is still lying around in eax so we can still act on it). But this won't work on Win7, where eax does not contain anything meaningful. It looks like timer cancellation propagation will not work on Win7. Should I do GetProcAddress/dynamic light-up?

kennykerr: Yes, we'll have to do that. You can (almost) use `load_runtime_function` except that its currently only used for combase functions and the thread pool is implemented by kernel32.

https://github.com/microsoft/cppwinrt/blob/1502e293d6e8f0a2c017431f6cf186d1551202fe/strings/base_agile_ref.h#L124-L140

This is the way Win7 fallback is handled, for example:

https://github.com/microsoft/cppwinrt/blob/1502e293d6e8f0a2c017431f6cf186d1551202fe/strings/base_activation.h#L21-L38

oldnewthing: Can't quite do it that way because I want to disable a feature if the GetProcAddress fails. There is no "fallback" available. (I guess I could see if result == fallback_Blah.)

kennykerr: Whatever you think is best.

oldnewthing: #898

PR #898, opened 2021-03-23
Restore Win7 functionality (more correctly this time)

oldnewthing:
Win7 cannot cancel threadpool timers or threadpool waits atomically, so we simply disable cancellation of resume_after and resume_on_signal when running downlevel. Sorry, Win7.

Most of the work in this PR is updating `load_runtime_function` so is no longer hard-coded to combase.


kennykerr (approved):
Thanks!

PR #903, opened 2021-03-26
Add support for boxing and unboxing most kinds of arrays

oldnewthing:
The boxing and unboxing functions now support most kinds of arrays. The notable exception is arrays of enumerations. Nothing technically preventing it. Just lazy.

You can box and unbox arrays nearly anywhere you could box and unbox scalars, with one exception: You cannot unbox arrays with `unbox_value_or`. The reason is that `com_array` is not copyable, and `unbox_value_or` returns a copy of its second parameter on failure. We could try to address this by having `unbox_value_or` accept its second parameter as an `array_view` when asked to unbox an array type. But I didn't try. Just lazy.

One annoyance is the treatment of `com_array<GUID>`. This is identical to `com_array<guid>` but I had to play games to convert between them efficiently. Maybe it wasn't worth it. Perhaps I should have been lazy.

### DIFF
--- a/strings/base_reference_produce.h
+++ b/strings/base_reference_produce.h
@@ -111,90 +111,105 @@ namespace winrt::impl
     struct reference_traits
     {
         static auto make(T const& value) { return winrt::make<impl::reference<T>>(value); }
+        using itf = Windows::Foundation::IReference<T>;
     };
 
     template <>
     struct reference_traits<uint8_t>
     {
         static auto make(uint8_t value) { return Windows::Foundation::PropertyValue::CreateUInt8(value); }
+        using itf = Windows::Foundation::IReference<uint8_t>;
     };
 
     template <>
     struct reference_traits<uint16_t>
     {
         static auto make(uint16_t value) { return Windows::Foundation::PropertyValue::CreateUInt16(value); }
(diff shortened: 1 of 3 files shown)


strings/base_reference_produce.h
```diff
@@ -111,128 +111,291 @@ namespace winrt::impl
     struct reference_traits
     {
         static auto make(T const& value) { return winrt::make<impl::reference<T>>(value); }
+        using itf = Windows::Foundation::IReference<T>;
```
kennykerr, line 114:
```suggestion
        using interface = Windows::Foundation::IReference<T>;
```

We try to avoid abbreviations in C++/WinRT, where practical, for clarity.

kennykerr (approved):
Thanks!

strings/base_reference_produce.h
oldnewthing: Can't do that thanks to these macros in combaseapi.h:

```cpp
#define __STRUCT__ struct
#define interface __STRUCT__
```

Guess how I discovered those macros.

oldnewthing: I could call it `reference_interface`?

kennykerr: Sigh 😜 - nevermind then.

PR #904, opened 2021-03-26
Support nested Windows namespace

### LINKED ISSUE #892: Code generation fails when creating a component with a nested Windows namespaces
Description: Introducing a new namespace: Microsoft.Windows.<ComponentName> leads to compilation issues in ProjectReunion. Apparently, the compiler is trying to find ::Windows stuff inside Microsoft::Windows::*.

kennykerr:
Fixes #892

Adds a simple test class to ensure this works.

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1619,7 +1619,7 @@ namespace cppwinrt
         else if (optional)
         {
             auto format = R"(            if (%) *% = nullptr;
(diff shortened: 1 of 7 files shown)


DefaultRyan:
The test case covers the type name appearing in isolation, but I was expecting to also see issues with parameter and field types. Your test class also needs to have methods/properties that use things from a Windows namespace. `Windows.Foundation.Uri` comes to mind.

At least that's the approach I was planning as I started looking at this.

PR #905, opened 2021-03-26
Windows test

kennykerr:
@DefaultRyan is this what you had in mind?


DefaultRyan:
Pretty much.  While we're here, since structs have different codegen patterns than interfaces and runtime classes, it would also be a good idea to have a struct in the nested namespace that has a field coming from the main Windows namespace, like `Rect`.

kennykerr:
Bingo, structs tripped up the type writer.

kennykerr:
That smoked out a few more issues. Thanks @DefaultRyan!

PR #906, opened 2021-03-28
IReference operators need to go into the level-1 header

### LINKED ISSUE #801: The generated 'operator==' of a struct with an 'IReference<T>' member is sensitive to inclusion/include order
E.g. consider the following:
```
namespace Component
{
    struct Test
    {
        IReference<Int32> member;
    };
}
```
Then the generated file `winrt/impl/Component.2.h` will include an `operator==` definition that looks something like:
```c++
    inline bool operator==(Test const& left, Test const& right) noexcept
    {
        return left.member == right.member;
    }
```
It seems that the intended functionality is that the above uses the `operator==` defined for `IReference<T>`:
```c++
    template <typename T>
    bool operator==(IReference<T> const& left, IReference<T> const& right)
    {
        if (get_abi(left) == get_abi(right))
        {
            return true;
        }

        if (!left || !right)
        {
            return false;
        }

        return left.Value() == right.Value();
    }
```
This works fine, however the issue lies in the fact that the above `operator==` for `IReference<T>` is defined in `winrt/Windows.Foundation.h`. This file is included by neither `winrt/Component.h` nor `winrt/impl/Component.2.h`. Both - curiously enough - include `winrt/impl/Windows.Foundation.2.h`, however the desired function definition is not present in this file.

The end result is that `winrt/Windows.Foundation.h` must be explicitly included _before_ `winrt/Component.h` for it to work properly.

oldnewthing:
That way they are defined whenever the interface is defined.

This brings `IReference<T>` in line with all other types with custom `operator==`, which had already defined those operators at the point the type is defined.

Note that `operator co_await` is still defined in the top header even though the type being co-awaited is defined in the level-1 or level-2 header. This isn't a problem in practice because if you try to `co_await` a `DispatcherQueue` (say), without having included the top-level header, you will be told, "I don't know how to co_await a DispatcherQueue" because the DispatcherQueue doesn't implement await_ready.

Fixes #801


kennykerr (approved):
Thanks for the fix!

oldnewthing:
@kennykerr Note that `write_namespace_special` generated `base_reference_produce.h` conditionally upon the presence of `PropertyValue`. But `write_namespace_special_1` generates `base_reference_produce_1.h` unconditionally. Was there an old version of Windows.Foundation that didn't have IReference support?

kennykerr:
That's a remnant of the xlang side project that aimed to support C++/WinRT on non-Windows metadata. That's been canned so we can remove those conditionals.

PR #909, opened 2021-03-30
Use separate vsixmanifest files for Component and Standalone deployments

DefaultRyan:
We need to remove the dependency on `Microsoft.VisualStudio.Component.CoreEditor` for the "Component" VSIX, but want to keep this dependency for the "Standalone" VSIX.

This change forks the `source.extension.vsixmanifest` file into "Component" and "Standalone" flavors.


Scottj1s (approved):
thanks for taking this on!

vsix/Component/source.extension.vsixmanifest
```diff
@@ -0,0 +1,31 @@
...
+    <MoreInfo>https://go.microsoft.com/fwlink/?linkid=869449</MoreInfo>
+    <License>LICENSE</License>
+    <GettingStartedGuide>https://docs.microsoft.com/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt</GettingStartedGuide>
+    <ReleaseNotes>https://docs.microsoft.com/windows/uwp/cpp-and-winrt-apis/faq</ReleaseNotes>
+    <Icon>cppwinrt.ico</Icon>
+    <PreviewImage>cppwinrt.png</PreviewImage>
+    <Tags>WinRT, C++, cppwinrt, native</Tags>
+  </Metadata>
+  <Installation AllUsers="true">
+    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[15.0,17.0)" />
```
BenJKuhn, line 17: just curious, will this line, as written, block on 17.1, 17.2, etc.? If so, is that intentional?

DefaultRyan: This line will actually block starting on 17.0 and up. I didn't write this line originally, but I imagine it was intentional, with the thinking being that we'd want to actually test this extension on 17.0 first. This seems like a good decision, especially in light of some upcoming changes to VS 17.

PR #913, opened 2021-04-02
Factories no longer default to no_weak_ref

oldnewthing:
Originally, factories defaulted to no_weak_ref because there wasn't believed to be any use case for weak references to factories, and the thought was that the weak reference support would be a performance penalty.

In practice, the weak reference support does not have a significant performance impact, so it's okay to leave it on even for factories.

Allowing weak references to factories is important because factories implement the static members of a runtime class, and those members may require taking weak references:

* Event handlers need weak references so that the event handler doesn't create a circular reference back to the factory.
* Queueing work to a background thread takes a weak reference so that the work can be abandoned if the factory has been destroyed. (We don't want to extend the lifetime of COM objects beyond the lifetime of COM itself.)

Furthermore, the ability to take a weak reference to a factory is required in order to use event_revoker for static events, since event_revoker requires that the event source support weak references.

### DIFF
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -1041,8 +1041,6 @@ namespace winrt::impl
             return Windows::Foundation::TrustLevel::BaseTrust;
         }
 
-        using is_factory = std::disjunction<std::is_same<Windows::Foundation::IActivationFactory, I>...>;
-
     private:
 
         class has_final_release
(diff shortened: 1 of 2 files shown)


strings/base_implements.h
```diff
@@ -1041,8 +1041,6 @@ namespace winrt::impl
             return Windows::Foundation::TrustLevel::BaseTrust;
         }
 
-        using is_factory = std::disjunction<std::is_same<Windows::Foundation::IActivationFactory, I>...>;
-
```
oldnewthing, line 1045: Deleting this public type is a potential breaking change. Do we care?

kennykerr: I don't think it matters too much - developers can easily sniff this out themselves - and its not likely to be something a project would depend on, I think.

kennykerr (approved):
Thanks!

PR #914, opened 2021-04-02
Natvis support for hresult_error and array_view

oldnewthing:
hresult_error displays the wrapped hresult.

array_view displays the size and the first few elements. It expands to the elements in the view, in the form of an array.

### DIFF
--- a/natvis/cppwinrt.natvis
+++ b/natvis/cppwinrt.natvis
@@ -20,6 +20,15 @@
     <DisplayString Condition="this == 0">null</DisplayString>
   </Type>
   <!--Primitive type visualizers-->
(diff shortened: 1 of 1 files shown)


sylveon:
Might be harder but it would be nice if `hresult_error`'s natvis could display the restricted error information, if any, since sometimes you have a useless error code ("Catastrophic failure") but a good description of the issue in the stashed `IRestrictedErrorInfo*`

oldnewthing:
Yes, it would be nice. But it's also harder.

kennykerr:
That would require a debugger extension for runtime execution of code, not something the natvis format itself supports.

kennykerr (approved):

Scottj1s (approved):

PR #917, opened 2021-04-04
Fix few typos in comments

jbrianceau:
Note that this change only impacts comments and metadata, not actual code.

### DIFF
--- a/test/catch.hpp
+++ b/test/catch.hpp
@@ -10545,7 +10545,7 @@ namespace Catch {
             // Extracts the actual name part of an enum instance
             // In other words, it returns the Blue part of Bikeshed::Colour::Blue
             StringRef extractInstanceName(StringRef enumInstance) {
-                // Find last occurrence of ":"
+                // Find last occurence of ":"
                 size_t name_start = enumInstance.size();
                 while (name_start > 0 && enumInstance[name_start - 1] != ':') {
                     --name_start;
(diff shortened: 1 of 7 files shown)


test/catch.hpp
```diff
@@ -10545,7 +10545,7 @@ namespace Catch {
             // Extracts the actual name part of an enum instance
             // In other words, it returns the Blue part of Bikeshed::Colour::Blue
             StringRef extractInstanceName(StringRef enumInstance) {
-                // Find last occurence of ":"
+                // Find last occurrence of ":"
```
kennykerr, line 10548: This file is generated by https://github.com/catchorg/Catch2 - please make the contribution there and then I'll be happy to ingest a new version from that project.

kennykerr (changes requested):
Thanks for the fixes! Happy to accept them, but please revert the change to catch.hpp.

jbrianceau:
> Thanks for the fixes! Happy to accept them, but please revert the change to catch.hpp.

Sure, I've just updated the PR accordingly.

test/catch.hpp
jbrianceau: Pull request is out: https://github.com/catchorg/Catch2/pull/2206
I don't think that it's worth ingesting a new version just for such a small cosmetic change and I definitely won't be offended if you wait for a more valuable payload before making a new ingestion :)

kennykerr (approved):
Thanks!

PR #918, opened 2021-04-05
Support non top-level root namespaces in item templates

sylveon:
This fixes the following issue reported on dev community: https://developercommunity.visualstudio.com/t/vsix-does-not-correctly-provide-rootnamespace-and/1153008

To achieve this, I had to write a custom template wizard so that I'm able to inject a custom root namespace replacement string with the dots replaced by `::`. The code was basically 1:1 taken from https://github.com/NuGet/NuGet.Client/blob/7e3b0e445edecbdb6c46895ee4375cf834f45d87/src/NuGet.Clients/NuGet.VisualStudio.Interop/TemplateWizard.cs, so the automatic package installation that happens when first using the template should still work.

The custom template wizard could enable better support for custom scenarios in the future as well, for example nested folders vs dotted prefix conventions for namespace hierarchy.

To make this correctly light up in the VSIX, some infrastructure work on the MS side will be required to strongly sign the `Microsoft.Windows.CppWinRT.dll` file, since otherwise VS refuses to load it.

The assembly version is `0.0.0.0` because it's carried around by the VSIX so there's no need to bind specific versions (the NuGet template wizard DLL already doesn't for example), but if desired this could most likely be updated to use proper versions.

Some changes where made to the `.csproj` file to enable a better inner loop (now running `build_vsix Debug`, opening the solution in VS and hitting F5 should "just work" to get the extension installed in the experimental instance).

### DIFF
--- a/vsix/ItemTemplates/BlankPage/BlankPage.cpp
+++ b/vsix/ItemTemplates/BlankPage/BlankPage.cpp
@@ -7,7 +7,7 @@
 using namespace winrt;
 using namespace Windows::UI::Xaml;
 
-namespace winrt::$rootnamespace$::implementation
+namespace winrt::$cpprootnamespace$::implementation
 {
     $safeitemname$::$safeitemname$()
     {
--- a/vsix/ItemTemplates/BlankPage/BlankPage.h
+++ b/vsix/ItemTemplates/BlankPage/BlankPage.h
@@ -2,7 +2,7 @@
 
 #include "$safeitemname$.g.h"
 
-namespace winrt::$rootnamespace$::implementation
+namespace winrt::$cpprootnamespace$::implementation
 {
     struct $safeitemname$ : $safeitemname$T<$safeitemname$>
     {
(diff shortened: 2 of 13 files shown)


kennykerr:
Unfortunately, we have nobody on staff to maintain or test the project templates so they are largely frozen except for critical fixes. Everything under nuget/vsix/natvis is unfortunately very brittle and requires manual testing. Feel free to fork them and provide superior project templates but these will have to stay as they are.

sylveon:
That's a shame.

I don't have the bandwidth to maintain a fork either, so I'll just leave the branch here in hope msbuild work gets picked up again.

kennykerr:
Many thanks, I wish I had a better answer.

sylveon:
@kennykerr do you think it might be worthwhile to revive this change? Since the cppwinrt addon is part of VS 2022 now, it probably has some amount of resources for testing, simply as part of standard VS process (plus it should have time to cook in preview and find issues before it hits stable).

WASDK templates also have this issue, so I could fix both at once, or try a fix in one before the other.

kennykerr:
I don't think the resourcing situation has changed, but I've tagged a few project maintainers. If one of them is available to offer guidance and mentorship for this issue, they will reach out according to the [contributing guide](https://github.com/microsoft/cppwinrt/blob/master/docs/contributing.md) to discuss and agree on an approach.

https://github.com/orgs/microsoft/teams/cppwinrt-maintainers

sylveon:
I believe a link doesn't ping, one has to directly write @microsoft/cppwinrt-maintainers (but my message won't ping since I am not a member of the MS org and the team is not public).

kennykerr:
Let's try again: @microsoft/cppwinrt-maintainers

PR #922, opened 2021-04-08
Inline internal error codes

### LINKED ISSUE #921: winrt::impl::error_ok and friends are duplicated in each translation unit
I was running a binary size analysis tool on our optimized binary and the "wasteful duplication" analysis flagged `winrt::impl::error_ok` and all of the other errors in that same block.  These are `constexpr` so they are in the readonly segment, as expected, but they are duplicated.  According to this tool they are 4 bytes each, duplicated 42 times, and add up to 168 bytes _each_.  That adds up to roughly 3KB of our final binary being copies of these error codes.

This can be fixed by changing `constexpr` to `inline constexpr`.  That syntax will result in them being de-duped to a single copy.  It requires C++17, but cppwinrt already requires that so it shouldn't be a problem.

kennykerr:
Fixes #921 - thanks for spotting these @dmachaj - we already inlined most of the heavy hitters but these ones didn't get noticed.

### DIFF
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -222,22 +222,22 @@ namespace winrt::impl
     using trust_level_type = Windows::Foundation::TrustLevel;
 #endif
 
-    constexpr hresult error_ok{ 0 }; // S_OK
-    constexpr hresult error_fail{ static_cast<hresult>(0x80004005) }; // E_FAIL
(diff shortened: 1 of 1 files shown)


dmachaj:
Wow, that was really fast :).  Thanks @kennykerr.

dmachaj (approved):

PR #930, opened 2021-04-23
Runtime class name for bare IInspectable should not be "Object"

oldnewthing:
This tickles an edge case in the metadata-based marshaller. Use an empty string instead.

### DIFF
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -430,6 +430,15 @@ namespace winrt::impl
         }
     };
 
+    template <>
+    struct runtime_class_name<Windows::Foundation::IInspectable>
+    {
+        static hstring get()
+        {
+            return {};
+        }
+    };
+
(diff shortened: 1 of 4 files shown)


strings/base_implements.h
```diff
@@ -430,6 +430,15 @@ namespace winrt::impl
         }
     };
 
+    template <>
+    struct runtime_class_name<Windows::Foundation::IInspectable>
```
oldnewthing, line 434: Note that `name_of` still returns `Object`, which is necessary for proper autogeneration of parameterized interfaces.

ChrisGuzak (approved):

strings/base_implements.h
kennykerr: Yep, that's kinda important. 😉

kennykerr (approved):
Thanks!

PR #934, opened 2021-04-30
Update .targets file to support local servers, support .exe files from dependent projects just like .dll's

ChrisGuzak:
(#860)


ChrisGuzak:
To verify it I hand patched tools\cppwinrt\Microsoft.Cpp.CppWinRTEnabled.targets and built the projects listed below. There were some link failures due to missing static  lib dependencies (incomplete dependency info in the LDG) but these are unrelated to the change. The Terminal project was built successfully with this change.

Is this sufficient validation?

kennykerr:
Kicked off a build.

PR #936, opened 2021-05-04
Add switch to disable merge validation to support certain test scenarios

Scottj1s:


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -561,7 +561,8 @@ $(XamlMetaDataProviderPch)
         </PropertyGroup>
         <PropertyGroup>
             <!-- mdmerge.exe wants the folders to not have a trailing \ -->
-            <_MdMergeParameters>-v @(CppWinRTMdMergeMetadataDirectories->'-metadata_dir &quot;%(RelativeDir).&quot;', '&#x0d;&#x0a;')</_MdMergeParameters>
+            <_MdMergeParameters Condition="'$(CppWinRTMergeNoValidate)'==''">-v</_MdMergeParameters>
```
jlaanstra, line 564: Right now this adds -v for both true and false. Do you want to assign a default value and check for true?

Scottj1s: sure, I can change to !='false'

kennykerr (approved):

PR #946, opened 2021-05-19
Make the v142 toolset the default.

jlaanstra:
Instead of making the oldest toolset the default, make the newest toolset the default. This ensures that projects load correctly in future VS versions with a higher version number.


kennykerr (approved):
Kicked off a build - `/azp run` doesn't work any more for some reason.

PR #948, opened 2021-05-20
Add #pragma once in addition to header guard.

### LINKED ISSUE #945: Generated headers should use #pragma once in addition to header guard
Headers that have a header guard are still processed in full by the preprocessor.

Since C++/WinRT doesn't use any tricks that require the header files to be multiply-included, it could emit a `#pragma once` at the beginning of each generated header.

This will allow the compiler (or preprocessor, as it were) to use at least one clever optimization:

A compiler that supports `#pragma once` could keep an inode cache of the files that have been included.
This would allow it to completely skip preprocessing the file to find the header guard at the point of `#include`

jlaanstra:
Fixes #945. Given that the component headers already use this I don't think this would break C++/WinRT compatibility with compilers that don't support this (technically it's non-standard). 

Question is if we need to keep the open file guard?

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -69,6 +69,8 @@ namespace cppwinrt
 
     static void write_open_file_guard(writer& w, std::string_view const& file_name, char impl = 0)
     {
+        write_include_guard(w);
+            
         std::string mangled_name;
 
         for (auto&& c : file_name)


kennykerr:
I vaguely recall having to use `ifdef` guards to avoid compatibility problems and using `pragma once` caused other problems for some compilers, but perhaps those issues have since been dealt with by the compilers.

kennykerr (approved):
I'm happy to accept this change, but its possible this will break in some obscure way in the OS build. If that happens, its simple enough to revert.

jlaanstra:
Let me send mail to see whatI can find out, before merging.

kennykerr (changes requested):
It seems like we don't have a compelling reason to take this change. There's no data that this improves performance and some risk that this breaks the OS build.

jlaanstra:
There is at least the argument of consistency that makes it worth it, even if there is no large perf benefit? Especially given that the stl follows the same pattern?

kennykerr:
I don't mind the change itself - I'm just concerned about the fallout from the OS build - ChrisG's comment reminded me that I'd seen that very issue in the OS build before when using pragma once with C++/WinRT. I don't know if it still exists but it would burn a Windows build finding out. I'm happy if you want to do that finding out though. 😉

jlaanstra:
Let's find out :). Worst case I learn something new.

kennykerr (approved):

PR #956, opened 2021-06-12
Add IMemoryBufferByteAccess accessor

### LINKED ISSUE #944: IMemoryBufferReference should have a data method
IBuffer has a data method in the C++/WinRT projected version but IMemoryBufferReference doesn't. I think IMemoryBufferReference should also have a data method so it will be easier to get the data than casting to IMemoryBufferByteAccess.

dlech:
This adds a new `data()` method to get the data pointer returned by the `IMemoryBufferByteAccess::GetBuffer()` method for `Windows.Foundation.IMemoryBufferReference`.

Fixes #944.

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1310,6 +1310,18 @@ namespace cppwinrt
         }
 )");
         }
+        else if (type_name == "Windows.Foundation.IMemoryBufferReference")
+        {
+            w.write(R"(
+        auto data() const
+        {
+            uint8_t* data{};
+            uint32_t capacity{};
+            check_hresult(static_cast<D const&>(*this).template as<IMemoryBufferByteAccess>()->GetBuffer(&data, &capacity));
+            return data;
+        }
+)");
+        }
         else if (type_name == "Windows.Foundation.Collections.IIterator`1")
         {
             w.write(R"(
--- a/strings/base_abi.h
+++ b/strings/base_abi.h
@@ -127,6 +127,11 @@ namespace winrt::impl
         virtual int32_t __stdcall Buffer(uint8_t** value) noexcept = 0;
     };
 
+    struct __declspec(novtable) IMemoryBufferByteAccess : unknown_abi
+    {
+        virtual int32_t __stdcall GetBuffer(uint8_t** value, uint32_t* capacity) noexcept = 0;
+    };
+
     template <> struct abi<Windows::Foundation::TimeSpan>
     {
         using type = int64_t;
@@ -155,4 +160,5 @@ namespace winrt::impl
     template <> inline constexpr guid guid_v<IContextCallback>{ 0x000001da, 0x0000, 0x0000, { 0xC0,0x00,0x00,0x00,0x00,0x00,0x00,0x46 } };
     template <> inline constexpr guid guid_v<IServerSecurity>{ 0x0000013E, 0x0000, 0x0000, { 0xC0,0x00,0x00,0x00,0x00,0x00,0x00,0x46 } };
     template <> inline constexpr guid guid_v<IBufferByteAccess>{ 0x905a0fef, 0xbc53, 0x11df, { 0x8c,0x49,0x00,0x1e,0x4f,0xc6,0x86,0xda } };
+    template <> inline constexpr guid guid_v<IMemoryBufferByteAccess>{ 0x5b0d3235, 0x4dba, 0x4d44, { 0x86,0x5e,0x8f,0x1d,0x0e,0x4f,0xd0,0x4d } };
 }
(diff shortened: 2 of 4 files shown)


dlech:
FYI, I have tested this with the Python projection in the xlang repository.

cppwinrt/code_writers.h
```diff
@@ -1308,6 +1308,18 @@ namespace cppwinrt
             static_cast<D const&>(*this).template as<IBufferByteAccess>()->Buffer(&data);
             return data;
         }
+)");
+        }
+        else if (type_name == "Windows.Foundation.IMemoryBufferReference")
```
jonwis, line 1313: Should this return both the length and the data? I guess a caller would use `::Capacity()` but that costs another call.

jonwis (approved):

cppwinrt/code_writers.h
sylveon: This would violate the convention of `data()` returning only the pointer.

I think a better idea would be adding an implicit conversion to `std::span`, if someone wants to retrieve both at once they can then use this.

Jaiganeshkumaran: The problem with an implicit conversion is that if you only have a single reference to the object and then assign it directly from a temporary then the returned std::span would be invalid and you'd not notice. Instead, it's better to have a method named as_span.

sylveon: `as_span()` wouldn't make that detail immediately obvious either.

dlech: I'm not a C++ programmer, so I don't have an opinion here about spans. Having a `data()` method is continent for the WinRT Python bindings since `IBuffer` already has the same.

kennykerr: Unfortunately, `IMemoryBufferByteAccess` and `IBufferByteAccess` are quite different - in hindsight the same interface should probably have been used for both classes. The original `data` method was modeled after `std::vector`'s `data` method, providing "direct access to the underlying array". Let's stick with that as that is a well-known convention. We can add explicit conversions to `std::span` to both later on if that ends up being valuable. That way they remain largely consistent.

kennykerr (changes requested):
Looks good - please add a basic test.

cppwinrt/code_writers.h
```diff
@@ -1308,6 +1308,18 @@ namespace cppwinrt
...
+        {
+            w.write(R"(
+        auto data() const
+        {
+            uint8_t* data{};
+            uint32_t capacity{};
+            check_hresult(static_cast<D const&>(*this).template as<IMemoryBufferByteAccess>()->GetBuffer(&data, &capacity));
+            return data;
```
oldnewthing, line 1321: Why not call the method `view()` and have it return

```cpp
    return array_view(data, capacity);
```

?

cppwinrt/code_writers.h
kennykerr: I'd prefer not to add more uses of array_view as it should be "deprecated" in favor of std::span - the trouble is span is new to C++20 and we haven't yet made the break with C++17.

oldnewthing: That's fair. We can keep this in our back pocket once we break with C++17.

kennykerr:
So we can either go with the proposed `data` method for now or add a `span` method to both that returns a [std::span](https://en.cppreference.com/w/cpp/container/span) wrapped in a `__cpp_lib_span` [feature test](https://en.cppreference.com/w/cpp/feature_test). Either way, we need a basic test for the additions and then we can wrap this up.

Jaiganeshkumaran:
> So we can either go with the proposed `data` method for now or add a `span` method to both that returns a [std::span](https://en.cppreference.com/w/cpp/container/span) wrapped in a `__cpp_lib_span` [feature test](https://en.cppreference.com/w/cpp/feature_test). Either way, we need a basic test for the additions and then we can wrap this up.

It can just use winrt::array_view and later an implicit conversion to span can be added separately.

kennykerr:
> It can just use winrt::array_view and later an implicit conversion to span can be added separately.

As I [mentioned here](https://github.com/microsoft/cppwinrt/pull/956#discussion_r651144349), the array_view should not be used for new code. It was originally only included as a param type (not to be used directly) and even that as a stopgap while we waited for std::span to be standardized.

cppwinrt/code_writers.h
kennykerr: On the other hand, I just remembered that we've already been using [feature testing](https://en.cppreference.com/w/cpp/feature_test) to enable C++20 features for coroutines and ranges. We can just do the same for std::span.

dlech:
Hi @kennykerr, I have added a basic test (based on https://github.com/microsoft/cppwinrt/blob/master/test/old_tests/UnitTests/buffer.cpp).

kennykerr (approved):
Thanks!

PR #965, opened 2021-06-24
IReference operators in level-1 header should be forward declarations

oldnewthing:
Implementation relies on methods in the top-level header, so defer implementation to there.

This reverts the change to base_reference_produce.h from #906 and reduces base_reference_produce_1.h to merely forward references. This is enough to get structures to compile, although you cannot meaningfully use any IReferences in them until you include Windows.Foundation.h (which is consistent with existing rules for namespace headers).


kennykerr (approved):
Thanks! Kicked off a build.

PR #968, opened 2021-06-28
Update VSIX to work with Dev17

DefaultRyan:
This updates the VSIX to work with Dev17 (VS 2022) Preview.

One thing I learned is that the VCTargets install path was deprecated starting in VS2019, so I had to update it to an explicit path under MSBuild.

See:
- https://docs.microsoft.com/en-us/visualstudio/extensibility/set-install-root?view=vs-2019
- https://docs.microsoft.com/en-us/cpp/build/reference/msbuild-visual-cpp-overview?view=msvc-160

The second link suggests that the new path won't work in VS2017, and I thought we didn't still support VS2017, so I've moved up the minimum VS version from 2017 to 2019.

### DIFF
--- a/vsix/Standalone/source.extension.vsixmanifest
+++ b/vsix/Standalone/source.extension.vsixmanifest
@@ -14,10 +14,24 @@
     <Tags>WinRT, C++, cppwinrt, native</Tags>
   </Metadata>
   <Installation AllUsers="true">
-    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[15.0,17.0)" />
-    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[15.0,17.0)" />
-    <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[15.0,17.0)" />
-  </Installation>
+    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[16.0, 17.0)">
+      <ProductArchitecture>x86</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[16.0, 17.0)">
+      <ProductArchitecture>x86</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[16.0, 17.0)">
+      <ProductArchitecture>x86</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>  </Installation>
   <Dependencies>
     <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
   </Dependencies>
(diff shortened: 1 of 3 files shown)


vsix/Standalone/source.extension.vsixmanifest
```diff
@@ -14,10 +14,24 @@
...
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[16.0, 17.0)">
+      <ProductArchitecture>x86</ProductArchitecture>
+    </InstallationTarget>
+    <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[17.0, 18.0)">
+      <ProductArchitecture>amd64</ProductArchitecture>
+    </InstallationTarget>  </Installation>
```
sylveon, line 34: Was `</Installation>` meant to go on a new line here?

sylveon:
Could you consider taking a look to #918 to see if we can roll it in at the same time?

vsix/Standalone/source.extension.vsixmanifest
walbourn: In the VSIX manifest for VS 2017+, there's a block like this (in this example it's supposed to install on VS 2017 or VS 2019):

```
    <Installation>
        <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[15.0,17.0)" />
        <InstallationTarget Version="[15.0,17.0)" Id="Microsoft.VisualStudio.Enterprise" />
        <InstallationTarget Version="[15.0,17.0)" Id="Microsoft.VisualStudio.Community" />
    </Installation>
```

To make it support VS 2022 as well, my example needs to be (as I understand it)

```
    <Installation>
        <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[15.0,17.0)" />
        <InstallationTarget Version="[15.0,17.0)" Id="Microsoft.VisualStudio.Enterprise" />
        <InstallationTarget Version="[15.0,17.0)" Id="Microsoft.VisualStudio.Community" />
        <InstallationTarget Id="Microsoft.VisualStudio.Pro" Version="[17.0, 18.0)">
            <ProductArchitecture>amd64</ProductArchitecture>
        </InstallationTarget>
        <InstallationTarget Id="Microsoft.VisualStudio.Enterprise" Version="[17.0, 18.0)">
            <ProductArchitecture>amd64</ProductArchitecture>
        </InstallationTarget>
        <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[17.0, 18.0)">
            <ProductArchitecture>amd64</ProductArchitecture>
        </InstallationTarget>
    </Installation>
```

sylveon: I meant stylistically, right now both `</InstallationTarget>` and `</Installation>` are on the same line

jlaanstra (approved):

kennykerr (approved):
If @jlaanstra is happy with this that's good enough for me.

asklar (approved):

tim-weis:
Dropping VS2017 support is somewhat unfortunate. VS2017 has one distinct feature that all successors are missing: The ability to deploy an application from layout to a USB-attached Windows 10 Mobile device, allowing immediate on-device debugging.

Unless I'm the Last Man Standing that still supports Windows 10 Mobile for several applications, could you re-evaluate the cost/risk/effort of continuing VS2017 support, and ideally re-enable VS2017 that support?

kennykerr:
@tim-weis can you use an older version the VSIX for your VS 2017 install? Sadly as it is, we hardly have the resources to keep the VSIX running with the latest version.

s-wtnb:
@kennykerr Where can I find older versions of VSIX for VS 2017?
Do I need to build from C++/WinRT source code(contains VSIX) prior to the [2.0.210505.3 version](https://github.com/microsoft/cppwinrt/releases/tag/2.0.210707.1) that has this fix?

kennykerr:
I don't actually know. I assumed the marketplace lets you download older versions, but there doesn't appear to be anything obvious there for finding older versions. Perhaps you can fiddle with the download link.

https://aka.ms/cppwinrt/vsix

s-wtnb:
thx!
I was able to get the old VSIX by changing the version part of the URI in the download link.

ex)
[2.0.210505.3](https://marketplace.visualstudio.com/_apis/public/gallery/publishers/CppWinRTTeam/vsextensions/cppwinrt101804264/2.0.210505.3/vspackage)

PR #971, opened 2021-06-30
check_hresult returns the hresult as a convenience

oldnewthing:
Like `check_pointer`, `check_hresult` accepts multiple successful values. The `check_pointer` function returns the successful pointer, but `check_hresult` swallows the successful result, which makes using `check_hresult` inconvenient to use with COM methods that return `S_FALSE`. For example:

```cpp
if (hresult error = root->SelectSingleNode(query, node.put()); check_hresult(error), error == S_OK)
{
    process(node);
}
```

Extending `check_hresult` to return the successful `hresult` simplifies the code:

```cpp
if (check_hresult(root->SelectSingleNode(query, node.put())) == S_OK)
{
    process(node);
}
```

I think it is unlikely that anybody has been doing `decltype(check_hresult())`, so changing the return value should be non-breaking.

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -576,12 +576,13 @@ WINRT_EXPORT namespace winrt
         throw_hresult(impl::hresult_from_win32(WINRT_IMPL_GetLastError()));
     }
 
-    inline void check_hresult(hresult const result)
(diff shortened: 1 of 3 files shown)


kennykerr (approved):
Thanks!

PR #973, opened 2021-07-05
Automatically enable App SDK 0.8 coroutine support when present

sylveon:
Fixes https://github.com/microsoft/WindowsAppSDK/issues/957 by automatically including the header from the App SDK

Keeps behavior consistent with other Microsoft-provided dispatcher types, and allows the App SDK to version the header themselves still

Testing methodology:
 - Create a new console app
 - Install CppWinRT and WinUI 3 NuGet packages
 - Replace cppwinrt.exe from NuGet with self-built binary
 - Ensure the following code compiles:
   ```cpp
   #include <winrt/Microsoft.UI.Dispatching.h>
   
   winrt::fire_and_forget test(const winrt::Microsoft::UI::Dispatching::DispatcherQueue& queue)
   {
       co_await queue;
   }
   ```

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -3300,6 +3300,10 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
         {
             w.write(strings::base_xaml_typename);
         }
+        else if (namespace_name == "Microsoft.UI.Dispatching")
+        {
+            w.write(strings::base_coroutine_dispatching_winui);
+        }
     }
 
     static void write_namespace_special_1(writer& w, std::string_view const& namespace_name)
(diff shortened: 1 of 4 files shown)


kennykerr:
We've decided against adding support for additional dispatchers. Those from outside libraries can instead ship their own helper libraries.

sylveon:
This presents several problems though:
- When the switch to C++20 modules is done, libraries won't be able to extend the `winrt` module in any way, leading to the need to continue using headers, or to ship a different module. This is counterproductive to the adoption of modules.
- Something that was expected to "just work", due to precedent with existing types, won't work anymore. And since it's a poorly documented change/requirement (it is written [absolutely nowhere](https://www.google.com/search?q=%22microsoft.ui.dispatching.co_await.h%22) that I need to `#include <microsoft.ui.dispatching.co_await.h>` for coroutines to work), it will lead to user confusion, like in the original issue linked above.

If this is the way to go, I think there should be a standard way for libraries to extend cppwinrt's codegen, as to inject their own "strings" into generated headers.

This PR just includes the header from the NuGet package if available, so it can still be updated and versioned outside of cppwinrt, and doesn't add any code specific to that dispatcher directly in the core cppwinrt codebase.

PR #976, opened 2021-07-07
Revert "Prevent inadvertent assignment to temporary object"

kennykerr:
Although a useful change (#825), it caused a few compatibility issues with the OS build.


oldnewthing (approved):
Maybe I can bring back a scaled-down version that simply blocks assignment to rvalue hstring.

PR #977, opened 2021-07-08
Fix target override as it doesn't work when the target import order changes.

jlaanstra:
Current C++/WinRT override the GetResolvedWinMD target. However if the import order changes this override can easily break, like it does when NuGet imports the targets as part of PackageReference. 

This PR removes the override.

Validation:

- [x] OS
- [x] WinUI
- [x] Terminal
- [x] YourPhone
- [ ] React-Native


sylveon:
I thought PackageReference was not supported for C++?

jlaanstra:
There seems to be ways to enable it, see https://github.com/dotnet/project-system/issues/2491

kennykerr:
Kicked off a build.

jlaanstra:
Normal validation passed. Since the react-native team is also experimenting with PackageReferences, tagging @JunielKatarn to see if he can test this for react-native. Probably don't have to block on that.

kennykerr (approved):

PR #979, opened 2021-07-09
Ensure we don't override XamlLanguage if it's already set.

jlaanstra:
When NuGet imports the props as part of PackageReference, it easy to end up in situations where the project file defines XamlLanguage before the C++/WinRT props are imported.

We shouldn't override if the value is already set.


jlaanstra:
This is a very straightforward change. I think we can skip the extensive validation for this.

kennykerr (approved):

kennykerr:
While Dale is updating the OS compiler vpack, I don't think he's updating the msbuild vpack. Did you want to do that or is there not really any value?

jlaanstra:
Yeah let's do it. Best to be in sync.

PR #986, opened 2021-07-21
Harden `put` functions against misuse

kennykerr:
The `put_abi` functions are already hardened in this way.

### DIFF
--- a/strings/base_com_ptr.h
+++ b/strings/base_com_ptr.h
@@ -111,7 +111,7 @@ WINRT_EXPORT namespace winrt
 
         type** put() noexcept
         {
-            WINRT_ASSERT(m_ptr == nullptr);
+            release_ref();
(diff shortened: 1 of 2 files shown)


ChrisGuzak (approved):

sylveon (approved):
That's actually how I would expect put() to behave.

kennykerr:
@sylveon long story... when I first built C++/WinRT and the DX library before it, the `put` functions were used heavily by the projection itself for binding ABI calls under the hood and injecting the cleanup code into every callsite was prohibitive. Today, C++/WinRT doesn't use those interop functions internally very much and its mainly used by app-level interop code where developers expect more from such helpers.

kennykerr:
The build actually succeeded - seems to be a mix of Azure build and GitHub reporting issues.

PR #989, opened 2021-07-26
Fix cyclic dependency error when upgrading VS 2022

DefaultRyan:
This will help prevent [this error](https://developercommunity.visualstudio.com/t/Unable-to-update-from-Preview-11-to-Pre/1476321) from happening in the future. #960 should be fully working after this.


kennykerr (approved):

raffaeler:
@DefaultRyan 
Is this already on the marketplace?
I tested it today on VS2022 Preview 3.1 and still get the error

LanceMcCarthy:
Confirmed still seeing this (attempted install yesterday) 

image

sylveon:
See https://github.com/microsoft/cppwinrt/issues/960#issuecomment-908558849

VS 2022 Preview 4 isn't out yet

Bluesman74:
The preview is out, but the Extension still doesn't work 
WinRT_BadVersion

raffaeler:
Same here.
I tested it today on VS2022 Preview 4 and still get the same error

ericjpeters:
It looks like there was a fix merged into master, but as of 09/22/2021 / VS2022 Preview 4.1, I'm still getting the same error.

Issue #991, opened 2021-07-29
guid parse errors can't be caught and handled at runtime


dfields-msft:
The `winrt::guid` constructor that accepts a string calls `abort()` on failure instead of throwing an exception. This precludes the application from handling the error, e.g. to show a message to the user or fall back to default inputs. There is also no `try_parse()` function, so the `guid` parsing functionality of the library is limited to strings which are a priori known to be valid GUIDs.

kennykerr:
Yes, these are constexpr constructors meant to be used with literals at compile time. If you want to do it at run time you can always use a function like `CLSIDFromString`.

dfields-msft:
`CLSIDFromString` is quite clumsy and "un-modern" compared with the parsing constructor for `winrt::guid`. If the `abort()` were simply changed to a `throw`, it would be just as good for `constexpr` (perhaps better, since `abort()` itself isn't strictly `constexpr`), and also better suited for runtime use, which is available either way since `constexpr` functions are exposed for runtime use. What disadvantage is there to using `throw` instead?

dfields-msft:
Yes, I've successfully used `throw` in `constexpr` with Clang and MSVC before, and the documentation suggests that this is normative (in that a `constexpr` branch which evaluates to a `throw` statement is malformed and will fail to compile, but that a constexpr function may contain a `throw` statement that is not evaluated for a given compile-time invocation without error). See:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#note-50
https://en.cppreference.com/w/cpp/language/constexpr
https://en.cppreference.com/w/cpp/language/constant_expression

PR #992, opened 2021-07-30
Use throw rather than abort() for guid parse failures

dfields-msft:
Make `winrt::guid("...")` more useful at runtime by throwing on failure instead of aborting the program. See #991

### DIFF
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -20,7 +20,7 @@ namespace winrt::impl
     };
 
     template <typename T>
-    constexpr uint8_t hex_to_uint(T const c) noexcept
+    constexpr uint8_t hex_to_uint(T const c)
     {
         if (c >= '0' && c <= '9')
         {
@@ -36,22 +36,22 @@ namespace winrt::impl
         }
         else 
         {
-            abort();
+            throw std::invalid_argument("Character is not a hexadecimal digit");
         }
     }
 
     template <typename T>
-    constexpr uint8_t hex_to_uint8(T const a, T const b) noexcept
+    constexpr uint8_t hex_to_uint8(T const a, T const b)
     {
         return (hex_to_uint(a) << 4) | hex_to_uint(b);
     }
 
-    constexpr uint16_t uint8_to_uint16(uint8_t a, uint8_t b) noexcept
+    constexpr uint16_t uint8_to_uint16(uint8_t a, uint8_t b)
     {
         return (static_cast<uint16_t>(a) << 8) | static_cast<uint16_t>(b);
     }
 
-    constexpr uint32_t uint8_to_uint32(uint8_t a, uint8_t b, uint8_t c, uint8_t d) noexcept
+    constexpr uint32_t uint8_to_uint32(uint8_t a, uint8_t b, uint8_t c, uint8_t d)
     {
         return (static_cast<uint32_t>(uint8_to_uint16(a, b)) << 16) |
                 static_cast<uint32_t>(uint8_to_uint16(c, d));
@@ -85,7 +85,7 @@ WINRT_EXPORT namespace winrt
     private:
 
         template <typename TStringView>
-        static constexpr guid parse(TStringView const value) noexcept
+        static constexpr guid parse(TStringView const value)
         {
             if (value.size() != 36 || value[8] != '-' || value[13] != '-' || value[18] != '-' || value[23] != '-')
             {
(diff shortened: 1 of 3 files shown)


dfields-msft:
@kennykerr does a successful run of "C++ for WinRT Xlang Internal Build" include the Clang validation you were asking for?

kennykerr (changes requested):
Thanks! Please add a test that verifies this works. At a glance, this won't compile.

strings/base_types.h
```diff
@@ -89,7 +89,7 @@ WINRT_EXPORT namespace winrt
         {
```
kennykerr, line 89: This function can't be `noexcept` otherwise it will just abort as before.

Also, all of the `hex_to_xxx` functions are noexcept and terminate on failure. They would need to be updated as well.

kennykerr:
Yes, you can either validate it locally or run the build once a test has been added.

kennykerr:
Oh and the CI build for some reason will only run branches local to this repo.

dfields-msft:
Does the test infrastructure have any existing support for compile-time negative tests (i.e. code that's expected to _not_ compile)? I can certainly add compile-time and runtime positive tests, as well as a runtime negative test, but I'm not going to bite off setting up test infrastructure for compile-time negative tests if that's not already available...

kennykerr:
No negative testing unfortunately.

Just need a test that validates it works in a constant expression (e.g. `static_assert`) and a runtime test that validates it works successfully (e.g. `assert`) and unsuccessfully (e.g. throw and catch).

strings/base_types.h
```diff
@@ -170,16 +170,27 @@ WINRT_EXPORT namespace winrt
...
+            (left.Data3 == right.Data3) &&
+            (left.Data4[0] == right.Data4[0]) &&
+            (left.Data4[1] == right.Data4[1]) &&
+            (left.Data4[2] == right.Data4[2]) &&
+            (left.Data4[3] == right.Data4[3]) &&
+            (left.Data4[4] == right.Data4[4]) &&
+            (left.Data4[5] == right.Data4[5]) &&
+            (left.Data4[6] == right.Data4[6]) &&
+            (left.Data4[7] == right.Data4[7]);
     }
```
kennykerr, line 187: Hmm, I'm worried this may have an impact on code size. `memcmp` was carefully picked to reduce code size as it is used repeatedly in every QueryInterface implementation. We'd have to build something big (like WinUI) with LTCG to see whether it remains an issue.

kennykerr: @DefaultRyan may have some more recent experience with this. My data is pretty old.

dfields-msft: Ah, ok; I wasn't aware of those considerations. On the one hand, I'm curious and hopeful that the compiler can be smart about this esp. since it's `constexpr`. On the other hand, the motivation is a it weak; I only changed this to make the test easier to write. So it's probably best to revert this part and hack the test together instead.

sylveon: I'd recommend `if (std::is_constant_evaluated())` for this, but then that's C++20 only.

sylveon: Just tested if the compiler was smart about it, it isn't: compiling with `/FA` gives this

image

dfields-msft: Thanks for looking into it! I'll revert this part of the change for now.

strings/base_types.h
```diff
@@ -36,22 +36,22 @@ namespace winrt::impl
         }
         else 
         {
-            abort();
+            throw std::invalid_argument("Character is not a hexadecimal digit");
```
sylveon, line 39: Can we put the character in the error message?

strings/base_types.h
```diff
@@ -85,11 +85,11 @@ WINRT_EXPORT namespace winrt
     private:
 
         template <typename TStringView>
-        static constexpr guid parse(TStringView const value) noexcept
+        static constexpr guid parse(TStringView const value)
         {
             if (value.size() != 36 || value[8] != '-' || value[13] != '-' || value[18] != '-' || value[23] != '-')
             {
-                abort();
+                throw std::invalid_argument("value is not a valid GUID string");
```
sylveon, line 92: Same here, put the actual value rather than just `value`?

strings/base_types.h
dfields-msft: The minor complication is that it could be a `char` or a `wchar_t`... likewise for the string in your other comment.

sylveon: Ah, fair enough. Basic ASCII for char and wchar_t is the same, but if random garbage ends up being passed, this might not be desirable

dfields-msft:
@kennykerr I'm not sure how to trigger the "C++ for WinRT Xlang internal build" pipeline run that's required by policy... Is this even possible, since my changes are coming from a different fork? I don't think I have permission to create a branch in the official microsoft fork.

kennykerr:
I've given you Write access so you should be able to push a branch. Then you can go here and build your branch:

https://dev.azure.com/microsoft/Dart/_build?definitionId=31784

dfields-msft:
Verified offline that Clang works with these changes too. @kennykerr any additional changes needed, or is this ready to merge?

kennykerr (approved):

dfields-msft:
Thanks for approving the changes, @kennykerr! Looks like I'm not authorized to merge, so you may need to complete this PR yourself...

kennykerr:
@DefaultRyan or @jlaanstra can you help? I'm on the beach. 😎

PR #1002, opened 2021-08-18
Some uses of `co_await` were not protected by `WINRT_IMPL_COROUTINES`

### LINKED ISSUE #1001: Windows.Foundation.h has a few unguarded uses of co_await
All other uses are guarded by `WINRT_IMPL_COROUTINES`, however there are a couple of locations that look unguarded. From what I've seen in Windows.Foundation.h:

* `await_adapter::cancel_asynchronously`
* `when_all`
* `when_any`

oldnewthing:
Fixes #1001

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -167,7 +167,9 @@ namespace winrt::impl
     private:
         static fire_and_forget cancel_asynchronously(Async async)
         {
+#ifdef WINRT_IMPL_COROUTINES
             co_await winrt::resume_background();
+#endif
             try
             {
                 async.Cancel();
@@ -793,6 +795,7 @@ namespace std::experimental
 
 WINRT_EXPORT namespace winrt
 {
+#ifdef WINRT_IMPL_COROUTINES
     template <typename... T>
     Windows::Foundation::IAsyncAction when_all(T... async)
     {
@@ -838,4 +841,5 @@ WINRT_EXPORT namespace winrt
         impl::check_status_canceled(shared->status);
         co_return shared->result.GetResults();
     }
+#endif
 }
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -127,6 +127,7 @@ namespace winrt::impl
         }
     };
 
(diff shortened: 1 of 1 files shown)


strings/base_coroutine_foundation.h
```diff
@@ -167,7 +167,9 @@ namespace winrt::impl
     private:
         static fire_and_forget cancel_asynchronously(Async async)
         {
+#ifdef WINRT_IMPL_COROUTINES
```
sylveon, line 170: Why not call `submit_threadpool_callback` instead of changing the semantics of this function?

oldnewthing: Yes, I had a note to make this fix better, but this is to unblock "cannot build because of compiler errors" which is significantly more urgent.

dunhor: Could this type just be excluded entirely? Seems only to be used by `operator co_await` further down in the file (which is also conditioned on `WINRT_IMPL_COROUTINES`).

As is, I'd expect this to still give compilation errors, but of the form "function has no return"

kennykerr (approved):

strings/base_coroutine_foundation.h
oldnewthing: Okay, I went and did it.

kennykerr: Let's try to avoid further feature creep. It's useful to have fixes stand alone in PRs so they can easily be rolled back and reapplied if needed.

oldnewthing: The previous fix was incomplete. It kept the compiler happy but didn't preserve behavior.

oldnewthing: @dunhor Ooh, good point. I missed that it's used only by `co_await`. Let me see what happens if I take it out entirely.

kennykerr: Got it 👍

oldnewthing: @dunhor Okay, it works. I had to comment out a boatload of tests that used co_await, but the rest of them build successfully when `WINRT_IMPL_COROUTINES` is disabled.

dunhor (approved):

Issue #1008, opened 2021-08-31
`IAsyncOperation::GetResults` returns an object that has been destructed.


seven-mile:
Reproduced `cpp/winrt` version: `2.0.201201.7` and `2.0.210806.1`

`IAsyncOperation::GetResults` returns an object that has been destructed. But in the same situation, I can use `IAsyncOperation::get` instead to get the correct object. Is it a bug? Or I missed something?

A minimal example to reproduce this issue:

```cpp
winrt::HttpClient client;

winrt::IAsyncOperation<winrt::JsonObject> GetAsync(winrt::Uri rest_uri) {

  auto resp { co_await client.GetAsync(rest_uri) };
  if (!resp.IsSuccessStatusCode())
    winrt::throw_hresult(MAKE_HRESULT(SEVERITY_ERROR, FACILITY_HTTP, resp.StatusCode()));
      
  auto rest_content { co_await resp.Content().ReadAsStringAsync() };

  auto res = winrt::JsonObject::Parse(rest_content);
  // at least here, res is correct
  co_return res;
}

int main()
{
  winrt::init_apartment();

  auto get_op = GetAsync(winrt::Uri([ some URI returning json data ]));
  while (get_op.Status() == winrt::AsyncStatus::Started); // just wait

  if (get_op.Status() != winrt::AsyncStatus::Completed || !get_op.GetResults())
    return get_op.Status() == winrt::AsyncStatus::Canceled ? E_ABORT : get_op.ErrorCode().value;

  auto res = get_op.GetResults();
  // here res is "null" in debugger

  wprintf(L"%s\n", get_op.GetResults().ToString().c_str()); // failed: access violation (object already deleted)
  
  return 0;
}
```

oldnewthing:
This is by design. You can call `GetResults()` only once.

PR #1009, opened 2021-09-01
Fixes regressions

kennykerr:
The following PRs introduced bugs. The CI build didn't catch them because the tests just crashed, leaving them undetected. 

#1008 had a change to consuming events but the test threw an exception on the producing side which has never been supported. This removes those tests. @sylveon

#992 never worked as intended. @dfields-msft

### DIFF
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -159,12 +159,12 @@ WINRT_EXPORT namespace winrt
 
 #endif
 
-        constexpr explicit guid(std::string_view const value) noexcept :
+        constexpr explicit guid(std::string_view const value) :
             guid(parse(value))
         {
         }
 
-        constexpr explicit guid(std::wstring_view const value) noexcept :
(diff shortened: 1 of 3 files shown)


BenJKuhn (approved):

kennykerr:
Confirmed all tests have run and passed.

dfields-msft:
What task tracks fixing the tests/CI pipeline to detect crashes as failures?

kennykerr:
There isn't anything. Unfortunately, we don't have any budget for improving the build but it would be nice to get this fixed.

dfields-msft:
I filed #1010 to track, and spent a little while this afternoon looking into it. Here's a potentially simple fix: #1011

dfields-msft:
Also sorry for introducing the silent failures in the first place... I had some confusion about how Catch2 deals with tests that validate exception throwing and mistakenly believed that running with `-e` was required, and was passing (which is how it is logged, rather than skipping which is what it actually does...).

PR #1011, opened 2021-09-01
Generate test failure output on abnormal termination

### LINKED ISSUE #1010: Unit tests that terminate abnormally are not considered as failures
The CI pipeline doesn't see any failure messages in the test output on abnormal termination (i.e. crash), so these failures occur silently and do not block PRs, etc. A specific example when this can occur is when an exception hits a `noexcept` boundary and triggers a `std::terminate()`.

dfields-msft:
Fixes #1010 by calling Catch2's `reportFatal()` in the `std::terminate` handler so that failure logging is generated before the process exits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.c
 *.nupkg
 test*.xml
+test*_results.txt
 build
 packages
 Debug
--- a/run_tests.cmd
+++ b/run_tests.cmd
@@ -20,5 +20,6 @@ goto :eof
 
 :run_test
 if not "%target_version%"=="" set args=-o %1-%target_version%.xml -r junit
-_build\%target_platform%\%target_configuration%\%1.exe %args%
+rem Buffer output and redirect to stdout/stderr depending whether the test executable exits successfully. Pipeline will fail if there's any output to stderr.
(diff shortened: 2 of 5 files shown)


kennykerr:
Thanks David! Why don't you add the `noexcept` back to the `guid` constructor that I removed here:

https://github.com/microsoft/cppwinrt/pull/1009/files

That will verify whether the CI build fails visibly or silently. Once that has been confirmed, just remove the `noexcept` again and it should be green again.

kennykerr:
The log captures the failure but it doesn't fail the build itself.

image

dfields-msft:
It's worse than that - even **normal** test failures don't cause the pipeline to fail (see my latest commit in this PR and the corresponding run). I've sent email to the team to discuss.

dfields-msft:
Added some changes to reroute output to `stderr` when the test executable does not exit cleanly, and validated that this fails the pipeline with the `noexcept` in place as expected:
image

kennykerr (approved):
Thanks!

dfields-msft:
Logged #1012 on an unrelated test failure that passed on rerun (also passed on both x86 runs); this may have been masked for some time due to the lack of failure reporting in the build pipeline until now.

PR #1015, opened 2021-09-10
Enable /await opt-out so users can manually enable /await:strict.

### LINKED ISSUE #1014: Switch to /await:strict when compiling with /stdcpp17
Since 16.10 /await:strict is available: https://docs.microsoft.com/en-us/cpp/build/reference/await-enable-coroutine-support?view=msvc-160

This makes it easier to share C++/WinRT code between code compiled with /stdcpp17 and /stdcpp20 as it makes the coroutine abis compatible.

jlaanstra:
/await:strict is binary compatible with c++20 coroutines. Instead of adding new logic to the C++/WinRT build targets, allow opt-out of the current default so users can set this manually.

For details, see: https://docs.microsoft.com/en-us/cpp/build/reference/await-enable-coroutine-support?view=msvc-160 and https://devblogs.microsoft.com/cppblog/cpp20-coroutine-improvements-in-visual-studio-2019-version-16-11/

Fixes #1014

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -871,7 +871,7 @@ $(XamlMetaDataProviderPch)
     <ItemDefinitionGroup>
         <ClCompile>
             <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
-            <AdditionalOptions Condition="'%(ClCompile.LanguageStandard)' == 'stdcpp17'">%(AdditionalOptions) /await:strict</AdditionalOptions>
+            <AdditionalOptions Condition="'%(ClCompile.LanguageStandard)' == 'stdcpp17' And '$(CppWinRTEnableLegacyCoroutines)' != 'false'">%(AdditionalOptions) /await</AdditionalOptions>
             <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(GeneratedFilesDir)</AdditionalIncludeDirectories>
         </ClCompile>
         <Midl Condition="'$(CppWinRTModernIDL)' != 'false'">
--- a/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/nuget/Microsoft.Windows.CppWinRT.targets
@@ -871,7 +871,7 @@ $(XamlMetaDataProviderPch)
     <ItemDefinitionGroup>
         <ClCompile>
             <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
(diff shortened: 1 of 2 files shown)


oldnewthing:
The downside is that it means two libraries both compiled for C++17 with the same version of cppwinrt.exe but different versions of the NuGet package (or one compiled with the NuGet package and one without) are now no longer ABI compatible. I think we need to make this opt-in rather than a change in default.

jlaanstra:
My understanding is that CX code can also use /await:strict, My primary motivation for this is that without it you can't mix a static C++/WinRT library compiled with stdcpp20 within a dll project that uses C++/CX.

jlaanstra:
I am a fan of having opt-outs though so let me add that.

sylveon:
Yeah, I am generally not a fan of things like NuGet package forcing command line arguments on the project - if I was involved in the original design, I would have had the templates include `/await` and never touch the command line args in the NuGet package, as it now causes issues such as this (and the fact you couldn't compile C++20 until cppwinrt was updated, as well)

jlaanstra:
> Yeah, I am generally not a fan of things like NuGet packages forcing command line arguments on the project - if I was involved in the original design, I would have had the templates include `/await` and never touch the command line args in the NuGet package, as it now causes issues such as this (and the fact you couldn't compile C++20 until cppwinrt was updated, as well)

Agreed, instead of adding /await:strict, I'm instead adding the option to opt-out of the current /await, so project maintainers can disable it and pass their own option.

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -871,7 +871,7 @@ $(XamlMetaDataProviderPch)
     <ItemDefinitionGroup>
         <ClCompile>
             <AdditionalOptions>%(AdditionalOptions) /bigobj</AdditionalOptions>
-            <AdditionalOptions Condition="'%(ClCompile.LanguageStandard)' == 'stdcpp17'">%(AdditionalOptions) /await</AdditionalOptions>
+            <AdditionalOptions Condition="'%(ClCompile.LanguageStandard)' == 'stdcpp17' And '$(CppWinRTEnableLegacyCoroutines)' != 'false'">%(AdditionalOptions) /await</AdditionalOptions>
```
kennykerr, line 874: Why not just remove this, or does this affect existing projects?

sylveon: Removing it entirely could affect projects which relied on the implicit addition of /await.

kennykerr: Makes sense, the logic is just throwing me but then msbuild has never made sense to me.

kennykerr (approved):

PR #1017, opened 2021-09-13
Avoid exception in observable IndexOf implementation

### LINKED ISSUE #1013: convertible_observable_vector::IndexOf slows debugging because it uses exceptions in a non-exceptional case
We have a scenario where some code in XAML repeatedly calls IObservableVector::IndexOf, where `value` is a different type than the container items. That causes hresult_no_interface to be thrown and caught internally within cppwinrt many times [in this code](https://github.com/microsoft/cppwinrt/blob/05737b0ce1e129f07f784bb9eb461ea6a3cf7bd7/strings/base_collections_vector.h#L107).

All of these exceptions cause ~1s pauses in the debugger and make it difficult for the developer to be productive if they have first-chance exceptions turned on.

It appears that since #695 was merged, IInspectable.try_as is equivalent (?) to unbox_value and could be used in the implementation of IndexOf to avoid this. (Is that true, or did I miss some nuance there?)

kennykerr:
Fixes #1013

### DIFF
--- a/strings/base_collections_vector.h
+++ b/strings/base_collections_vector.h
@@ -125,19 +125,16 @@ namespace winrt::impl
                 {
                     return base_type::IndexOf(as, index);
                 }
-                else
-                {
-                    return false;
-                }
-            }
-            else if (auto as = value.try_as<T>())
-            {
-                return base_type::IndexOf(as.value(), index);
             }
             else
             {
-                return false;
+                if (auto as = value.try_as<T>())
+                {
+                    return base_type::IndexOf(as.value(), index);
+                }
             }
+
+            return false;
         }
 
         using base_type::GetMany;
--- a/strings/base_collections_vector.h
+++ b/strings/base_collections_vector.h
@@ -115,13 +115,27 @@ namespace winrt::impl
 
         bool IndexOf(Windows::Foundation::IInspectable const& value, uint32_t& index) const
         {
-            try
+            if constexpr (is_com_interface_v<T>)
             {
-                return IndexOf(unbox_value<T>(value), index);
+                if (!value)
+                {
+                    return base_type::IndexOf(nullptr, index);
+                }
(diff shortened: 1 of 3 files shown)


strings/base_collections_vector.h
```diff
@@ -115,13 +115,27 @@ namespace winrt::impl
...
         {
-            try
+            if constexpr (is_com_interface_v<T>)
             {
-                return IndexOf(unbox_value<T>(value), index);
+                if (!value)
+                {
+                    return base_type::IndexOf(nullptr, index);
+                }
+                else if (auto as = value.try_as<T>())
```
oldnewthing, line 124: I would prefer that the value-type branch be indented at the same level as the COM case, so that it's clear that we are dealing with two cases, not three.

```cpp
if constexpr (is_com_interface_v<T>)
{
    if (!value)
    {
        return base_type::IndexOf(nullptr, index);
    }
    else if (auto as = value.try_as<T>())
    {
        return base_type::IndexOf(as, index);
    }
    else
    {
        return false;
    }
}
else
{
    if (auto as = value.try_as<T>())
    {
        return base_type::IndexOf(as.value(), index);
    }
    else
    {
        return false;
    }
}
```

Could factor out the final "return false".

```cpp
if constexpr (is_com_interface_v<T>)
{
    if (!value)
    {
        return base_type::IndexOf(nullptr, index);
    }
    else if (auto as = value.try_as<T>())
    {
        return base_type::IndexOf(as, index);
    }
}
else
{
    if (auto as = value.try_as<T>())
    {
        return base_type::IndexOf(as.value(), index);
    }
}
return false;
```

oldnewthing (approved):

strings/base_collections_vector.h
kennykerr: good idea - done

PR #1019, opened 2021-09-14
Ignore MIDL files which are excluded from the build

sylveon:
Without this, cppwinrt would attempt to use the output of MIDL files which where excluded from the build, resuting in mdmerge either erroring due to missing files, or using partially outdated build output.


jlaanstra:
There are some uses on line 201/202. Probably want to fix those as well.

sylveon:
I'm not really sure how to fix those - the change I did in this PR was enough to unblock my build however.

sylveon:
@jlaanstra I fixed those two, and tested by building the project that was broken and a different project that uses XAML with these changes, just to be sure.

jlaanstra:
LGTM

jlaanstra (approved):

PR #1021, opened 2021-09-17
std::format support

### LINKED ISSUE #1018: std::format support for hstring, IStringable and more
C++ 20 introduces std::format however currently Windows Runtime types don't work directly with them. Consider providing support for directly passing hstring and IStringable to std::format and additionally I would like to see hstring work with std::format_to so you can avoid an extra copy from std::wstring.

sylveon:
Fixes #1018 

This implements support for std::formatting of WinRT types. `hstring` and direct `IStringable` is always enabled, while support for types deriving from `IStringable` is behind `WINRT_LEAN_AND_MEAN` (but they can always be used by downcasting to `IStringable` before formatting).

Added a few unit tests that validate support as well.


strings/base_stringable_format.h
```diff
@@ -0,0 +1,12 @@
...
+#ifdef __cpp_lib_format
+template<>
+struct std::formatter<winrt::Windows::Foundation::IStringable, wchar_t> : std::formatter<winrt::hstring, wchar_t>
+{
+    template<typename FormatContext>
+    auto format(winrt::Windows::Foundation::IStringable const& obj, FormatContext& fc)
+    {
+        return std::formatter<winrt::hstring, wchar_t>::format(obj.ToString(), fc);
+    }
+};
```
oldnewthing, line 11: It is unfortunate that you cannot SFINAE a partial specialization.

sylveon: Every day I wonder why I'm still using C++ 😂

kennykerr: This is why I'm enjoying Rust these days. 😎

cppwinrt/helpers.h
```diff
@@ -498,6 +498,27 @@ namespace cppwinrt
...
+        for (auto&& impl : type.InterfaceImpl())
+        {
+            const auto iface = impl.Interface();
+            if (iface.type() != TypeDefOrRef::TypeSpec && type_name(iface) == name)
+            {
+                return true;
+            }
+        }
+
+        if (auto base = get_base_class(type))
```
kennykerr, line 512: There can be a hierarchy of bases...

kennykerr: But this can get expensive (e.g. how long cppwinrt takes to run). Will the formatter not work polymorphically, with anything that is convertible to `IStringable` rather than specific types (like `std::hash`)? I never liked that about hash...

kennykerr: Why not just enable it for `hstring` and `IStringable` for now and developers can always cast if needed. The idea was to eventually remove everything inside `WINRT_LEAN_AND_MEAN` in a future (major) version anyway.

sylveon: > There can be a hierarchy of bases...

This is a recursive call, so it will check each base until it hits the end of the hierarchy. Or is there anything else I'm missing?

> But this can get expensive (e.g. how long cppwinrt takes to run).

I haven't personally seen much impact on cppwinrt's runtime when developing this. I can do measurements if wanted.

> Will the formatter not work polymorphically, with anything that is convertible to IStringable rather than specific types (like std::hash)? I never liked that about hash...

Yeah unfortunately std::formatter behaves exactly like std::hash there. But you can "enable" the desired behavior just by inheritance like I do in this PR. Since then the compiler can figure out the implicit conversion at the call site.

> Why not just enable it for hstring and IStringable for now and developers can always cast if needed.

I think this will lead to confusion, especially since the error message that would be given when called on a type that implements IStringable is deep within std::format internals and can't really give anything useful, but then it works when called on a direct IStringable. Giving a good error message would require similar codegen to what we're already doing, so why not just enable the support directly.

> The idea was to eventually remove everything inside WINRT_LEAN_AND_MEAN in a future (major) version anyway.

Seeing as there's only like a dozen of types in the system APIs that implement IStringable, we could maybe remove this from WINRT_LEAN_AND_MEAN already and enable it all the time.

kennykerr: Yes, if there's a small number of types rather just always enable - if its useful it shouldn't be in `WINRT_LEAN_AND_MEAN`.

sylveon:
No problem! I'll move the formatter specializations out of LEAN_AND_MEAN and open a new PR.

PR #1022, opened 2021-09-19
Improve classic COM support, and improve diagnostic for common error

oldnewthing:
Before this change, if you tried to use classic COM without first including <unknwn.h>, C++/WinRT spit out really confusing error messages like

```cpp
auto widget = object.as<::IClassicComInterface>();
```

> error: 'unbox_value_type': no matching overloaded function found

or worse: Just ignored you!

```cpp
struct Widget : WidgetT<Widget, ::IPersistStream>
{
    // IPersistStream is simply ignored!
};
```

This change relaxes the rules around `<unknwn.h>` and improves diagnostics.

1. If you want to use `com_ptr<I>`, `as<I>`, etc. with classic COM interfaces, you need only include `<unknwn.h>` at *some point* before using them. Doesn't have to be included before `winrt/base.h`. (And really, since you need to include `<unknwn.h>` in order to define *any* classic COM interfaces, the prerequisite is automatically satisfied in practice.)

2. If you want to use `implements<...>` with classic COM interfaces, you still need to include `<unknwn.h>` before `winrt/base.h`, but we improved the diagnostics so that if you forget, you get a compiler error instead of a runtime error:

> To implement classic COM interfaces, you must `#include <unknwn.h>` before including C++/WinRT headers.

Implementation notes
====================

The key step is forward-declaring the missing types `IUnknown` and `GUID` so that we can talk *about* them without needing to know what they are. This trick allows us to remove a large number of `#ifdef WINRT_IUNKNOWN_DEFINED` tests, which unlocks 80% of this feature.

To avoid repetition, introduced a new type trait `is_classic_com_interface<T>`, which factors out many uses of `is_base_of<::IUnknown, T> && !is_implements<T>`.

Having a forward declaration of `IUnknown` lets us detect that a type is a classic COM interface by seeing if it derives from `::IUnknown`.

Having a forward declaration of `GUID` lets us reinterpret-cast them to `guid`, which covers nearly everything `GUID`-related.

The only place where we really need to know what's in a `GUID` is the `constexpr guid(GUID const&)` constructor. For that, the constructor is templated with a dummy default parameter. The dummy default parameter delays expansion until the constructor is invoked. We then forward the `GUID` to a templated helper as a dependent type `T`, so the compiler accepts the helper despite not knowing what `T`'s members are.

This change found a bug in the unit tests, because Nested.HierarchyD used classic COM interface `IReferenceTrackerExtension` without having first included `unknwn.h`, so it was ignored!

(Amusing note: The implementation of this feature got shorter the longer I worked on it, as I gradually realized that nearly all of the places we needed to protect the use of `GUID` didn't need the duck-typing trick employed in the `guid(GUID const&)` constructor. It also took me a while to find a short-and-sweet version of the duck-typing trick. My earlier attempts were overly complicated and required new unit tests because I was changing the constructors too much. But the short version doesn't change any constructor overload matching at all, which means that we don't need fancy new tests either. Just a quick check that `constexpr` construction still works.)

### DIFF
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -44,17 +44,8 @@ namespace winrt::impl
     template <template <typename> typename Condition, typename T>
     using tuple_if = typename tuple_if_base<Condition, T>::type;
 
-#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
-
     template <typename T>
-    struct is_interface : std::disjunction<std::is_base_of<Windows::Foundation::IInspectable, T>, std::conjunction<std::is_base_of<::IUnknown, T>, std::negation<is_implements<T>>>> {};
-
-#else
-
-    template <typename T>
-    struct is_interface : std::is_base_of<Windows::Foundation::IInspectable, T> {};
-
-#endif
+    struct is_interface : std::disjunction<std::is_base_of<Windows::Foundation::IInspectable, T>, is_classic_com_interface<T>> {};
 
     template <typename T>
     struct is_marker : std::disjunction<std::is_base_of<marker, T>, std::is_void<T>> {};
@@ -485,19 +476,18 @@ namespace winrt::impl
         }
     };
 
-#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
-
     template <typename D, typename I>
-    struct producer<D, I, std::enable_if_t<std::is_base_of_v< ::IUnknown, I> && !is_implements_v<I>>> : I
+    struct producer<D, I, std::enable_if_t<is_classic_com_interface<I>::value>> : I
     {
-    };
-
-    template <typename D, typename I>
-    struct producer_convert<D, I, std::enable_if_t<std::is_base_of_v< ::IUnknown, I> && !is_implements_v<I>>> : producer<D, I>
-    {
-    };
-
+#ifndef WINRT_IMPL_IUNKNOWN_DEFINED
+        static_assert(std::is_void_v<I> /* dependent_false */, "To implement classic COM interfaces, you must #include <unknwn.h> before including C++/WinRT headers.");
 #endif
+    };
+
+    template <typename D, typename I>
+    struct producer_convert<D, I, std::enable_if_t<is_classic_com_interface<I>::value>> : producer<D, I>
+    {
+    };
 
     struct INonDelegatingInspectable : Windows::Foundation::IUnknown
     {
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -55,4 +55,8 @@
 
 #ifdef __IUnknown_INTERFACE_DEFINED__
 #define WINRT_IMPL_IUNKNOWN_DEFINED
+#else
+// Forward declare so we can talk about it.
+struct IUnknown;
+typedef struct _GUID GUID;
 #endif
--- a/strings/base_meta.h
+++ b/strings/base_meta.h
@@ -118,12 +118,14 @@ namespace winrt::impl
     };
 
     template <typename T>
-#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
-#ifdef __clang__
+#if defined(__clang__)
+#if __has_declspec_attribute(uuid)
     inline const guid guid_v{ __uuidof(T) };
 #else
-    inline constexpr guid guid_v{ __uuidof(T) };
+    inline constexpr guid guid_v{};
 #endif
+#elif defined(_MSC_VER)
+    inline constexpr guid guid_v{ __uuidof(T) };
 #else
     inline constexpr guid guid_v{};
 #endif
--- a/strings/base_reference_produce.h
+++ b/strings/base_reference_produce.h
@@ -212,14 +212,12 @@ namespace winrt::impl
         using itf = Windows::Foundation::IReference<guid>;
     };
 
-#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
     template <>
     struct reference_traits<GUID>
     {
-        static auto make(GUID const& value) { return Windows::Foundation::PropertyValue::CreateGuid(value); }
+        static auto make(GUID const& value) { return Windows::Foundation::PropertyValue::CreateGuid(reinterpret_cast<guid const&>(value)); }
         using itf = Windows::Foundation::IReference<guid>;
     };
-#endif
 
     template <>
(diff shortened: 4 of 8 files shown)


strings/base_meta.h
```diff
@@ -118,12 +118,14 @@ namespace winrt::impl
     };
 
     template <typename T>
-#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
-#ifdef __clang__
+#if defined(__clang__)
+#if __has_declspec_attribute(uuid)
```
kennykerr, line 122: Can't believe Clang still doesn't treat `__uuidof` as a `constexpr` expression. 🙄

strings/base_macros.h
```diff
@@ -55,4 +55,8 @@
 
 #ifdef __IUnknown_INTERFACE_DEFINED__
 #define WINRT_IMPL_IUNKNOWN_DEFINED
+#else
+// Forward declare so we can talk about it.
+struct IUnknown;
+typedef struct _GUID GUID;
```
kennykerr, line 61: Nice!

strings/base_implements.h
```diff
@@ -485,20 +476,19 @@ namespace winrt::impl
...
     };
 
-#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
-
     template <typename D, typename I>
-    struct producer<D, I, std::enable_if_t<std::is_base_of_v< ::IUnknown, I> && !is_implements_v<I>>> : I
+    struct producer<D, I, std::enable_if_t<is_classic_com_interface<I>::value>> : I
     {
+#ifndef WINRT_IMPL_IUNKNOWN_DEFINED
+        static_assert(std::is_void_v<I> /* dependent_false */, "To implement classic COM interfaces, you must #include <unknwn.h> before including C++/WinRT headers.");
```
kennykerr, line 483: Excellent.

kennykerr (approved):

strings/base_meta.h
sylveon: It should, I reported a bug back in 2018 and that bug was fixed in 2020: https://bugs.llvm.org/show_bug.cgi?id=38490

oldnewthing: Cool! Now to figure out how to detect this bugfix at compile time. Any clues? (Or is it okay to require everyone to have a fixed clang?)

kennykerr: Assuming this fix has been published, I would be ok with just requiring it. Older compilers can just stick with an older version of C++/WinRT.

sylveon: We can version detect or use [something to check if an expression is constexpr](https://stackoverflow.com/questions/55288555/c-check-if-statement-can-be-evaluated-constexpr), but Clang 11 is almost 1 year old so I'm not sure how much it's worth to support older. VS seems to require (or strongly suggest) an up to date clang to be able to use the LLVM platform toolset so I imagine the amount of consumers of cppwinrt with an out of date clang to be somewhat low.

oldnewthing: Sadly, [godbolt's clang hasn't been updated with the fix](https://gcc.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:16,endLineNumber:9,positionColumn:16,positionLineNumber:9,selectionStartColumn:16,selectionStartLineNumber:9,startColumn:16,startLineNumber:9),source:'typedef+struct+_GUID+%7B%0A++++unsigned+long++Data1%3B%0A++++unsigned+short+Data2%3B%0A++++unsigned+short+Data3%3B%0A++++unsigned+char++Data4%5B+8+%5D%3B%0A%7D+GUID%3B%0Astruct+__declspec(uuid(%2267f4c08e-703b-4afa-bb75-d13a3ba2c583%22))+blarg%3B%0A%0Aconstexpr+GUID++foo+%3D+__uuidof(blarg)%3B%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang_trunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-fms-extensions',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+clang+(trunk)+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',tree:'1',wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+(trunk)+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

kennykerr: Too bad - looks like its not been fixed after all.

sylveon: This is weird, it works on my local copy of Clang 12:
```
T:\Projects>type constexpr.cpp
typedef struct _GUID {
    unsigned long  Data1;
    unsigned short Data2;
    unsigned short Data3;
    unsigned char  Data4[ 8 ];
} GUID;
struct __declspec(uuid("67f4c08e-703b-4afa-bb75-d13a3ba2c583")) blarg;

constexpr GUID  foo = __uuidof(blarg);

T:\Projects>clang constexpr.cpp
LINK : fatal error LNK1561: entry point must be defined
clang: error: linker command failed with exit code 1561 (use -v to see invocation)

T:\Projects>clang --version
clang version 12.0.0
Target: i686-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\Llvm\bin
```

The version of clang on Godbolt runs on Linux - this might be related.

PR #1026, opened 2021-09-27
Properly report failure to switch to target apartment

oldnewthing:
If we were unable to switch to a target apartment (either by an explicit `co_await apartment_context` or implicitly at the completion of ` co_await IAsyncInfo`), we now report the exception in the `co_await`'ing coroutine, so it can be caught and handled.

Note that when this error occurs, we are stuck in limbo. We left the original apartment for the MTA, and now we can't get from the MTA to the target apartment. We have no choice but to raise the exception on the MTA, even though the coroutine may not have expected to be running there.

```cpp
IAsyncAction Trapped()
{
    Object o; // some object with a destructor
    co_await destroyed_ui_context; // throws an exception
}
```

The destructor for `Object` will run on the MTA, even though the `Trapped` coroutine never runs on the MTA under normal conditions.

This is making the best of a bad situation. We need to raise the exception into the coroutine so that the coroutine can enter a failure state and allow the coroutine chain to proceed.

Since the object `o` already had to be prepared for destruction on the original apartment or the switched-to apartment (in case an exception occurs in either apartment), we know that the object cannot have apartment affinity anyway. So we're probably okay to destruct it on the MTA.

The old code raised the exception on whatever thread happened to notice that something bad happened. Sometimes that's okay, for example if you do a `co_await destroyed_ui_context` from an MTA thread, in which case the exception is thrown into the coroutine and can be caught, or left uncaught and send the coroutine into an error state.

However, all of the other cases weren't quite so happy. To avoid deadlocks, the apartment switch is routed through the threadpool, and that means that the failure is raised on a threadpool thread. There's nobody around to catch that exception, so the process fails fast.

We solve the problem by having an atomic hresult in the awaiter, which is initialized to 0 (`S_OK`) and transitions to error when we are about to resume the coroutine on the wrong apartment, so that the error can be propagated during `await_resume()`, which happens in the context of the awaiting coroutine and therefore can be caught by the coroutine (or sent to `unhandled_exception`).

The hresult in which we record the failure needs to be atomic to avoid data races in case multiple threads detect failure simultaneously. However, the access to it can be relaxed, because it is set and consumed by the same thread.

The current implementation pulls a sneaky trick: To avoid having to create an `operator co_await` for the `apartment_context` object, we continue to let it be its own awaiter. This means that if the apartment switch fails, we record the failure in the `apartment_context` itself. We are making the assumption that once an apartment context goes bad, it is unrecoverably broken.

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -77,36 +77,56 @@ namespace winrt::impl
         return 0;
     };
 
-    inline void resume_apartment_sync(com_ptr<IContextCallback> const& context, coroutine_handle<> handle)
+    struct resumption_failure
+    {
+        resumption_failure() = default;
+        resumption_failure(resumption_failure const&) noexcept {}
+        resumption_failure& operator=(resumption_failure const&) noexcept { return *this; }
+
+        // Call only on failure, never on success.
+        void report_failed(int32_t result) noexcept { failure.store(result, std::memory_order_relaxed); }
+        void check() const { check_hresult(failure.load(std::memory_order_relaxed)); }
+
+        std::atomic<int32_t> failure;
+    };
+
+    inline void resume_apartment_sync(com_ptr<IContextCallback> const& context, coroutine_handle<> handle, resumption_failure* failure)
     {
         com_callback_args args{};
         args.data = handle.address();
 
-        check_hresult(context->ContextCallback(resume_apartment_callback, &args, guid_of<ICallbackWithNoReentrancyToApplicationSTA>(), 5, nullptr));
+        auto result = context->ContextCallback(resume_apartment_callback, &args, guid_of<ICallbackWithNoReentrancyToApplicationSTA>(), 5, nullptr);
+        if (result < 0)
+        {
+            // Resume the coroutine on the wrong apartment, but tell it why.
+            failure->report_failed(result);
+            handle();
+        }
     }
 
     struct threadpool_resume
     {
-        threadpool_resume(com_ptr<IContextCallback> const& context, coroutine_handle<> handle) :
-            m_context(context), m_handle(handle) { }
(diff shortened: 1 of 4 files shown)


strings/base_coroutine_threadpool.h
```diff
@@ -370,17 +390,19 @@ WINRT_EXPORT namespace winrt
...
         }
 
-        void await_resume() const noexcept
+        void await_resume() const
         {
+            failure.check();
         }
 
-        void await_suspend(impl::coroutine_handle<> handle) const
+        void await_suspend(impl::coroutine_handle<> handle)
```
oldnewthing, line 398: Nope, can't do this. It prevents you from `co_await`'ing a `const apartment_context`. I'll have to give the `apartment_context` an `operator co_await`.

antmor:
nit: the description's scope_exit have typos, `:;scope_exit` instead of `::scope_exit`

antmor:
So the only drawback here is that, if you have a STA-bound object, the `Release` and possibly the destructor of the object is not guaranteed to be freed on a STA if the co_await fails. We should at least make sure that in our documentation, we say that objects crossing a co_await need to be agile enough to not fail out if destruction doesn't happen happen on the source STA, if the thread can't be guaranteed by the developer.

oldnewthing:
@antmor If you have an STA-bound object, you're already in trouble, because it might destruct on the wrong apartment even in the absence of the MTA weirdness.

```cpp
fire_and_forget Victim()
{
   StaObject object;
   DoSomething1();
   co_await resume_foreground(dispatcher);
   DoSomething2();
}
```

If `DoSomething1()` or `resume_foreground()` raise an exception, then the `StaObject` destructs on the old apartment. But if the `co_await` or the `DoSomething2()` raise an exception, then the `StaObject` destructs on the new apartment.

So the StaObject needs to be prepared to be destructed in either the old apartment or the new one. In other words, *it must already be agile*.

kennykerr (approved):

PR #1030, opened 2021-09-30
Support binding `std::optional<T>` implementation to `IReference<T>` parameters

### LINKED ISSUE #1006: Support binding std::optional<T> implementation to IReference<T> parameters
Assume I have this IDL:

```c#
runtimeclass MyThing {
    Windows.Foundation.IReference<UInt32> GetSize();
}
```

And my implementation says:

```c++
struct MyThing : MyThingT<MyThing> {
    std::optional<uint32_t> m_sizeComputed;
    winrt::Windows::Foundation::IReference<uint32_t> GetSize() {
        if (!m_sizeComputed) {
            return nullptr;
        } else {
            return winrt::box_value(m_sizeComputed.value()).as<winrt::Windows::Foundation<IReference<uint32_t>>>();
    }
}
```

I'd like this to be easier on the implementation, where instead I can just say:

```c++
struct MyThing : MyThingT<MyThing> {
    std::optional<uint32_t> m_sizeComputed;
    const std::optional<uint32_t> GetSize() { return m_sizeComputed; }
}
```

Is there a way for C++/WinRT's shims to match up `IReference<T>`-things (properties, returned types, parameters) and auto-convert them to/from `std::optional<T>`?

kennykerr:
Fixes #1006

### DIFF
--- a/test/test/optional.cpp
+++ b/test/test/optional.cpp
@@ -1,0 +1,22 @@
+#include "pch.h"
+#include "winrt/test_component.h"
+
+using namespace winrt;
+
+TEST_CASE("optional")
+{
+    test_component::Optional object;
+
+    object.Property(Windows::Foundation::IReference(123));
+    REQUIRE(object.Property().Value() == 123);
+
+    object.Property(nullptr);
+    REQUIRE(object.Property() == nullptr);
+
+    object.Property(456);
+    REQUIRE(object.Property().Value() == 456);
+
+    object.Property(std::optional(789));
+    std::optional<int32_t> value = object.Property();
+    REQUIRE(value == std::optional(789));
+}
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1507,13 +1507,25 @@ namespace cppwinrt
         }
         else if (type_name == "Windows.Foundation.IReference`1")
         {
-            w.write(R"(        IReference(T const& value) : IReference<T>(impl::reference_traits<T>::make(value))
+            w.write(R"(        IReference(T const& value) : IReference(impl::reference_traits<T>::make(value))
         {
         }
-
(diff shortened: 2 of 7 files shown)


kennykerr:
@jonwis @sylveon looks good?

test/test/optional.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+TEST_CASE("optional")
+{
+    test_component::Optional object;
+
+    object.Property(Windows::Foundation::IReference(123));
+    REQUIRE(object.Property().Value() == 123);
+
+    object.Property(nullptr);
+    REQUIRE(object.Property() == nullptr);
```
oldnewthing, line 14: Also

```cpp
REQUIRE(object.Property() == std::optional<int32_t>());
REQUIRE(object.Property() == std::nullopt);
```

to verify that implicit conversion to `std::optional<int32_t>` can successfully compare with std::nullopt.

oldnewthing (approved):

test/test/optional.cpp
kennykerr: I don't think C++ supports this. The C++ compiler would have to perform an implicit `operator T` followed by an `operator ==` and that's more than C++ is willing to do implicitly. I would have to add `operator ==` for `std::optional` if this is desired.

sylveon:
LGTM, will help with some code that did pass around optionals but then needed to pass it to WinRT.

test/test/optional.cpp
oldnewthing: Okay well then nevermind then. May call out in a comment that those operations *don't* work.

kennykerr: Let's consider adopting C++20's spaceship operator to simplify stuff like this.

PR #1031, opened 2021-09-30
Optimize factory caching for apps and pinned DLLs

kennykerr:
C++/WinRT provides `WINRT_NO_MODULE_LOCK` that may be used by apps or pinned DLLs (that don't support unloading) to avoid the overhead of tracking of a module reference count. This update further optimizes the factory cache in such cases since unloading has an impact on caching performance. With this change, C++/WinRT is on par with [the Rust language](https://github.com/microsoft/windows-rs). Previously, Rust was just slightly faster. 

compare

Data from: https://github.com/kennykerr/lang-perf

cc @axelandrejs

### DIFF
--- a/strings/base_activation.h
+++ b/strings/base_activation.h
@@ -233,19 +233,23 @@ namespace winrt::impl
 
         explicit factory_count_guard(size_t& count) noexcept : m_count(count)
         {
+#ifndef WINRT_NO_MODULE_LOCK
 #ifdef _WIN64
             _InterlockedIncrement64((int64_t*)&m_count);
 #else
             _InterlockedIncrement((long*)&m_count);
 #endif
+#endif
         }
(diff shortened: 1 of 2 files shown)


JamesMcNellis (approved):
LGTM

jonwis:
A DLL that compiles with this should be super careful to pin itself, yes? Is there a "pin" type that should be turned on to grab and leak a DLL reference, or is that up to the DLL to do?

kennykerr:
C++/WinRT will do this automatically. If you define `WINRT_NO_MODULE_LOCK` then C++/WinRT's implementation of `DllCanUnloadNow` will always return false and you are effectively pinned. Such DLLs can simply not export `DllCanUnloadNow` as added insurance.

PR #1041, opened 2021-10-19
Allow to build a projection out of a component

sylveon:
This option comes in handy when you are writing IDL files in your projects not to produce a component but to consume some API without needing to carry the WinMD around (e.g. only consuming some parts of the API, not wanting to check in binaries to source control, third-party provides an IDL, etc). Without this change, you would face issues with trying to call static functions or constructions, because it was expecting an implementation to be present from a  `.g.cpp` file while there is none (since you are not actually implementing the API).


sylveon:
No problem, this isn't anything urgent

kennykerr:
Ignoring the specific use case, the idea of needing to generate a projection from some IDL seems sound.

Issue #1042, opened 2021-10-20
Comment added to auto-generated source files is confusing


ptorr-msft:
The comment says to `copy[...] these generated source files to your project` but that's ambiguous since a project consists only of links to files; there is no concept of literally "copying a file" to a project, only copying it to a different directory in the filesystem.

PR #1043, opened 2021-10-20
Update comments written by component_writers.h

ptorr-msft:
Clarified how developers should manage generated files. Addresses #1042


ChrisGuzak (approved):

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
...
+// add this file to your project, as any changes you make will be lost.
+//
+// If this is a new file:
+//   1. Copy this file from the 'Generated Files' directory into the
+//      directory where you store your other source files (e.g. the project root).
+//   2. Add the copied file to your project via Project -> Add Existing Item.
+//   3. Delete this comment and the 'static_assert' (below) from the copied file
+//      (do not modify the original file in the 'Generated Files' directory).
+//
+// If this is an update to an existing file:
```
oldnewthing, line 1001:
```suggestion
// To update an existing file in your project:
```

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
...
     {
         auto format = R"(
-// Note: Remove this static_assert after copying these generated source files to your project.
-// This assertion exists to avoid compiling these generated source files directly.
-static_assert(false, "Do not compile generated C++/WinRT source files directly");
+// WARNING: This file is automatically generated from IDL. Do not directly
+// add this file to your project, as any changes you make will be lost.
+//
+// If this is a new file:
+//   1. Copy this file from the 'Generated Files' directory into the
```
oldnewthing, line 995: Note that the name "Generated Files" did not come from cppwinrt.exe. It came from the Visual Studio `<GeneratedFilesDir>` property in `Directory.Build.props`. Somebody who uses a different build system (like Razzle) will have an output directory that isn't named "Generated Files".

Perhaps

```suggestion
//   1. Copy this file from its current location into the
```

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
     static void write_generated_static_assert(writer& w)
     {
         auto format = R"(
-// Note: Remove this static_assert after copying these generated source files to your project.
-// This assertion exists to avoid compiling these generated source files directly.
-static_assert(false, "Do not compile generated C++/WinRT source files directly");
+// WARNING: This file is automatically generated from IDL. Do not directly
```
oldnewthing, line 991: Incorrect. This file is generated from the winmd. Instead of saying how the file is generated, just say that it was generated.

```suggestion
// WARNING: This file is automatically generated by a tool. Do not directly
```

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
...
+// add this file to your project, as any changes you make will be lost.
```
oldnewthing, line 992: Perhaps also

```cpp
// This file is a stub you can use as a starting point for your implementation.
```

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
...
+//
+// If this is a new file:
```
oldnewthing, line 994: "If this is a new file" is unclear about what "this" is. Literally, "this" is the auto-generated file, which is naturally new. But the "this" in this sentence really means "the corresponding file in your project".

```suggestion
// To add a copy of this file to your project:
```

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
...
+//   1. Copy this file from the 'Generated Files' directory into the
+//      directory where you store your other source files (e.g. the project root).
+//   2. Add the copied file to your project via Project -> Add Existing Item.
```
oldnewthing, line 997: Be a little more build-system-agnostic. (People also use C++/WinRT with Razzle, CMake, etc.)

```suggestion
//   2. Add the copied file to your project. In Visual Studio, you
//      can use Project -> Add Existing Item.
```

cppwinrt/component_writers.h
```diff
@@ -988,9 +988,22 @@ namespace winrt::@::implementation
...
+//   3. Delete this comment and the 'static_assert' (below) from the copied file
+//      (do not modify the original file in the 'Generated Files' directory).
```
oldnewthing, line 999: Grammar / punctuation:

```suggestion
//   3. Delete this comment and the 'static_assert' (below) from the copied file.
//      (Do not modify this file.)
```

cppwinrt/component_writers.h
ptorr-msft: Interesting. The problem is that "current location" is relative, so if you've already moved / copied the file and then "forget" you already did it, the "current location" is actually the proper location. I'll change it to "original location" (which is still not super helpful, but...).

cppwinrt/component_writers.h
ChrisGuzak: that is the most important point, the rest is the machinery that support it.

PR #1060, opened 2021-11-17
Lay groundwork for localizing VSIX templates and remove Dev15 build cruft

DefaultRyan:
There are a couple big changes to the VSIX in this PR:

### Lay foundation for VSIX localization

(Copied this technique from https://github.com/microsoft/WindowsAppSDK/pull/1341)

There is an undocumented method for localizing string resources in VS project and item templates. Firstly, the extension containing the templates must provide a [VSPackage](https://docs.microsoft.com/en-us/visualstudio/extensibility/creating-an-extension-with-a-vspackage?view=vs-2019). Next, a `VSPackage.resx` (Visual Studio first looks for the `VSPackage` resource stream) is added to the extension project to hold the string resources; the IDs of these resources *must* be integers (specifically, within the range of a `unsigned int`). Finally, the [Name](https://docs.microsoft.com/en-us/visualstudio/extensibility/name-element-visual-studio-templates?view=vs-2019) and [Description](https://docs.microsoft.com/en-us/visualstudio/extensibility/description-element-visual-studio-templates?view=vs-2019) elements in each `.vstemplate`'s `<TemplateData>` section so that:
1. They do not provide a text value
2. The `Package` attribute is the GUID assigned to the extension's VSPackage
3. The `ID` attribute is the ID of the string resource as defined in the `.resx`

The convention that we will use for choosing resource IDs is as follows:
- Numbering starts at 1000. This gives us what should be a sufficient number of digits to work around the VS's `.resx` designer use of a lexical sort.

### Remove Dev15 props/targets build cruft

Since the release of Dev16 (Visual Studio 2019), the props and targets that provide build support have shipped in the NuGet package, and are no longer needed in the VSIX. This change removes these unneeded files.


evelynwu-msft (approved):
:shipit:

PR #1063, opened 2021-11-19
Allow non-void static setter methods when compiling with `-opt`

### LINKED ISSUE #1062: Allow async property setter
I have a property setter that needs to do some asynchronous operations. So I changed the return type from void to fire_and_forget however now it doesn't compile since the codegen in g.cpp attempts to return it which isn't possible in this case since fire_and_forget can't be returned in a void function.

kennykerr:
Fixes #1062

### DIFF
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -502,13 +502,14 @@ catch (...) { return winrt::to_hresult(); }
     }
 )";
 
+                        bool ignore_return = is_put_overload(method) || !signature.return_signature();
 
                         w.write(format,
                             signature.return_signature(),
                             type_name,
                             method_name,
                             bind<write_consume_params>(signature),
-                            is_put_overload(method) ? "" : "return ",
+                            ignore_return ? "" : "return ",
                             type_namespace,
                             type_name,
                             method_name,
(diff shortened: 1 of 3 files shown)


oldnewthing:
Similar problem exists with void methods.

```csharp
namespace MyProject
{
    static runtimeclass Sample
    {
        static void M1();
    }
}
```

```cpp
namespace winrt::MyProject::implementation
{
    struct Sample
    {
        Sample() = default;

        static int M1() { return 1; }
    };
}
```

Code gen:
```cpp
WINRT_EXPORT namespace winrt::MyProject
{
    void Sample::M1()
    {
        return MyProject::implementation::Sample::M1(); // error: returning value from non-void function
    }
}
```

cppwinrt/component_writers.h
```diff
@@ -508,6 +508,7 @@ catch (...) { return winrt::to_hresult(); }
                             type_name,
                             method_name,
                             bind<write_consume_params>(signature),
+                            is_put_overload(method) ? "" : "return ",
```
oldnewthing, line 511: Should be `is_return_type_void(method) ? "" : "return"` to cover methods returning void. For some imaginary helper function `is_return_type_void`.

kennykerr:
@oldnewthing good catch - updated.

oldnewthing (approved):

PR #1067, opened 2021-11-29
remove WINRT_NO_MAKE_DETECTION from project templates

ChrisGuzak:
these disable an important diagnostic and have not been needed for a long time as the Xaml Compiler was fixed a few years ago.


sylveon:
Why not remove the `WINRT_NO_MAKE_DETECTION` functionality entirely?

kennykerr:
@sylveon I would be in favor of that, although it can happen as a separate PR.

jlaanstra:
> Why not remove the `WINRT_NO_MAKE_DETECTION` functionality entirely?

That would be a breaking change.

jlaanstra (approved):

sylveon:
Yes, but code that compiles only using this feature is already wrong and needs to be fixed.

Maybe this is something that could be removed for a potential C++/WinRT 3.0, then.

kennykerr:
There's unlikely to be a v3 - more likely that we just slowly evolve, ditching dead features, adding new features, relying on package versioning to allow consumers to handle any breaking changes at their own pace.

PR #1073, opened 2021-12-05
Fix compilation under Clang

### LINKED ISSUE #1069: Specialization of 'std::formatter<winrt::Windows::Foundation::IStringable, wchar_t>' breaks with conformant compilers
E.g. when building with Clang 13 & the standard conformance flags, I see the error:
```
include\winrt/impl/Windows.Foundation.1.h(245,68): error: function 'ToString' with deduced return type cannot be used before it is defined
        return std::formatter<winrt::hstring, wchar_t>::format(obj.ToString(), fc);
                                                                   ^
include\winrt/impl/Windows.Foundation.0.h(864,34): note: 'ToString' declared here
        WINRT_IMPL_AUTO(hstring) ToString() const;
```
Seems that it's because the definition of `consume_Windows_Foundation_IStringable<D>::ToString` is defined in the "normal" `winrt/Windows.Foundation.h` header, but all of the `winrt/impl/*.h` files are included first, and one of these is the header trying to consume the function in question.

sylveon:
Fixes #1069

Was manually tested by building and running this snippet:
```cpp
// silence a clang coroutine warning caused by cppwinrt. we don't actually use coroutines here.
#define _SILENCE_CLANG_COROUTINE_MESSAGE

// std::format is hidden behind /std:c++latest due to ABI concerns right now
// but the headers don't detect Clang's -std=c++2b, so manually have them believe we are using C++23.
#define _HAS_CXX23 1

#include <format>
#include <iostream>

// required for cppwinrt to not fail building with clang
#include <guiddef.h>

#include "winrt/Windows.Foundation.h"

#pragma comment(lib, "onecore.lib")

struct stringable : winrt::implements<stringable, winrt::Windows::Foundation::IStringable>
{
    winrt::hstring ToString()
    {
        return L"a stringable object";
    }
};

int main()
{
    winrt::Windows::Foundation::IStringable obj = winrt::make<stringable>();
    std::wcout << std::format(L"This is {}", obj);
}
```

With `clang -Icppwinrt/_build/x64/Debug -std=c++2b test.cpp`

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -3259,6 +3259,7 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
             w.write(strings::base_reference_produce);
             w.write(strings::base_deferral);
             w.write(strings::base_coroutine_foundation);
+            w.write(strings::base_stringable_format);
         }
         else if (namespace_name == "Windows.Foundation.Collections")
         {
@@ -3295,7 +3296,7 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
         if (namespace_name == "Windows.Foundation")
         {
             w.write(strings::base_reference_produce_1);
-            w.write(strings::base_stringable_format);
+            w.write(strings::base_stringable_format_1);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


kennykerr (approved):
Looks good to me. @dunhor ?

dunhor (approved):

PR #1074, opened 2021-12-05
Make winrt::event clearable

sylveon:
While migrating a bunch of C++/CX, I encountered a (rather mediocre) custom `Event<>` implementation. After migrating its usage to `winrt::event`, I noticed there was two major differences between them:
- `Event<>` can be copied, and it didn't seem complex to allow copying `winrt::event`.
- In C++/CX, calling the destructor (or `Dispose` from C#) would clear the list of event handlers right away. In C++/WinRT, while manually implementing `IClosable`, you cannot achieve the same: the memory associated with the event handlers (and any of their captures) is only released upon the reference count reaching 0.

So, I implemented and tested both features to achieve parity


strings/base_events.h
```diff
@@ -375,8 +375,37 @@ WINRT_EXPORT namespace winrt
         using delegate_type = Delegate;
 
         event() = default;
-        event(event<Delegate> const&) = delete;
-        event<Delegate>& operator =(event<Delegate> const&) = delete;
+
+        event(event<Delegate> const& value)
```
oldnewthing, line 379: Can just say `event` instead of `event<Delegate>` thanks to injected class names.

strings/base_events.h
```diff
@@ -375,8 +375,37 @@ WINRT_EXPORT namespace winrt
...
+        {
+            slim_lock_guard const swap_guard(value.m_swap);
+            m_targets = value.m_targets;
+        }
+
+        event<Delegate>& operator =(event<Delegate> const& value)
```
oldnewthing, line 385: Also move assignment? ("No" is a valid answer.)

```cpp
event& operator=(event const& value)
{
    return assign(value);
}
event& operator=(event&& value)
{
    return assign(std::move(value));
}

private:
template<typename T>
event& assign(T&& value)
{
 ... existing code, except that
    slim_lock_guard const swap_guard(value.m_swap);
    new_targets = value.m_targets;

becomes
    if constexpr (std::is_lvalue_reference_v<T>)
    {
        slim_lock_guard const swap_guard(value.m_swap);
        new_targets = value.m_targets;
    }
    else
    {
        slim_lock_guard const swap_guard(value.m_change);
        slim_lock_guard const swap_guard(value.m_swap);
        new_targets = std::move(value.m_targets);
    }
...
}
```

Move assignment is tricky because
1. You have to worry about the move assignment throwing an exception midway. Not an issue here, since the delegate_array move assignment is nonthrowing.
2. You have a brief window in which both events are empty. I think that's okay. If you hit that window, then your code was already race-y and could validly have called nothing back.

If you are worried about code explosion, you can factor that constexpr into a helper.

```cpp
event& operator=(event const& value)
{
    return assign(value, [](auto&& v) {
        slim_lock_guard const swap_guard(v.m_swap);
        return v.m_targets;
    });
}
event& operator=(event&& value)
{
    return assign(value, [](auto&& v) {
        slim_lock_guard const swap_guard(v.m_change);
        slim_lock_guard const swap_guard(v.m_swap);
        return const_cast<event&&>(v).m_targets;
    });
}

event& assign(event const& value, delegate_array (*extract)(event const&))
{
... as before except
    new_targets = extract(value);
}
```

But I don't think code explosion is a problem because all of the event lvalue assignments will fold into one copy, and all of the event rvalue assignments will fold into another copy. So at most two copies.

strings/base_events.h
```diff
@@ -375,8 +375,37 @@ WINRT_EXPORT namespace winrt
         using delegate_type = Delegate;
 
         event() = default;
-        event(event<Delegate> const&) = delete;
-        event<Delegate>& operator =(event<Delegate> const&) = delete;
+
+        event(event<Delegate> const& value)
+        {
+            slim_lock_guard const swap_guard(value.m_swap);
+            m_targets = value.m_targets;
```
oldnewthing, line 382: Move constructor while we're at it? ("No" is a valid answer.)

strings/base_events.h
oldnewthing, line 379: Making `event` copyable will result in previously-uncopyable types becoming copyable. I wonder if that will cause problems. (Maybe somebody had been relying on non-copyability of event to block some unwanted action. I confess that I may have done that.)

kennykerr: Yes, I'm mainly concerned about this aspect of this PR. While it may be simple enough to do, I wonder whether this is worthwhile. Copying an event doesn't seem all that useful.

strings/base_events.h
```diff
@@ -466,6 +495,24 @@ WINRT_EXPORT namespace winrt
             }
         }
 
+        void clear()
```
kennykerr, line 498: I think just focusing on `clear` to support `IClosable` is a worthy effort. I'm reluctant to add more bells and whistles to `event` in general because there are just too many variations of event handling to satisfy everyone, but `clear` seems reasonably straightforward and universal.

strings/base_events.h
sylveon: Yeah, was existing so I left as is because it could have been a style choice.

strings/base_events.h
sylveon: Move assignment/construction requires more locking than just copy assignment, so I believe a copy ends up being cheaper than a move.

strings/base_events.h
sylveon: I agree that in the general case it doesn't seem very useful, but in our use case the event is stored within a type that is captured by value in a lambda. The lambda then ends up stored in `std::function`, and `std::function` expects the callable type to be copyable. Wrapping the type holding the `event` or `event` itself in `std::unique_ptr` doesn't works, because it's move-only. While `std::shared_ptr` works, it adds ref-counting on top of existing ref-counting (from COM), so it adds some amount of overhead.

kennykerr: That's more of a problem with `std::function` or whatever library that requires copies. The trouble is allowing copying may allow developers to easily introduce unexpected behavioral issues. So you allow copies, what does that mean? Are delegates expecting to be callable from multiple event sources? It just seems like this could open things up to some surprising issue.

sylveon: Right. This isn't a performance critical path anyways so wrapping it with std::shared_ptr will do for now. Will remove that from the PR and make it only clearable.

sylveon: Done

kennykerr (approved):
Thanks!

PR #1095, opened 2022-01-20
Add visualizer for winrt::Windows::UI::Color

JohnTasler:
Make it look more like a Color value:
\# 7F 80 80 40

### DIFF
--- a/natvis/cppwinrt.natvis
+++ b/natvis/cppwinrt.natvis
@@ -20,12 +20,15 @@
     <DisplayString Condition="this == 0">null</DisplayString>
   </Type>
   <!--Primitive type visualizers-->
+  <Type Name="winrt::Windows::UI::Color">
+    <DisplayString># {A,nvoXb} {R,nvoXb} {G,nvoXb} {B,nvoXb}</DisplayString>
+  </Type>
   <Type Name="winrt::array_view&lt;*&gt;">
     <DisplayString>{{size = {m_size}, {m_data,[m_size]}}}</DisplayString>
     <Expand>
       <ArrayItems>
-          <Size>m_size</Size>
-          <ValuePointer>m_data</ValuePointer>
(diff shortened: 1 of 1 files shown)


sylveon:
Why not use an hex RGBA color format? Something like `#7f808040`

Also what if one if the value fits in a single digit? Do you get `0` or `00`?

Scottj1s:
> Why not use an hex RGBA color format? Something like `#8080407f`
> 
> Also what if one if the value fits in a single digit? Do you get `0` or `00`?

agree - can we just remove the spaces?

Scottj1s:
Great addition - thanks!

sylveon:
This is funny, I was suggesting putting alpha at the end, as CSS uses that (`#RRGGBBAA`). But then I noticed that XAML uses `#AARRGGBB`, so keeping alpha at the front makes more sense in this context.

natvis/cppwinrt.natvis
```diff
@@ -20,12 +20,15 @@
     <DisplayString Condition="this == 0">null</DisplayString>
   </Type>
   <!--Primitive type visualizers-->
+  <Type Name="winrt::Windows::UI::Color">
+    <DisplayString># {A,nvoXb} {R,nvoXb} {G,nvoXb} {B,nvoXb}</DisplayString>
```
kennykerr, line 24: Please remove the spaces and confirm that leading zeroes are emitted.

JohnTasler: I wish there was a "half-space" character. I added the spaces so it would be easier to see each byte. I will confirm the single-digit case.

JohnTasler: I've removed the spaces and confirmed that, yes, byte values less than 0x10 have the leading zero added.

kennykerr (approved):

PR #1097, opened 2022-01-29
Fix weak references to coroutines

oldnewthing:
`root_implements::NonDelegatingRelease` sets `m_references = 1` prior to initiating destruction so that the destructor sees a nonzero reference count. As a side effect, it also breaks the connection to the weak reference control block.

`~root_implements` performs a bonus `subtract_reference` to clean up any weak reference control block that may have re-materialized during destruction. The extra decrement is safe because we set `m_references = 1` earlier.

However, if the `root_implements::Release` was overriden by a derived class, the derived class's `Release` may not do the same `m_references = 1` trick, and then the bonus decrement in the destructor will decrement below zero.

Who would override `root_implements::Release`? `promise_base`, that's who. As a result, a weak reference to a coroutine turns into poison when the coroutine destructs, because the control block now has a strong reference of 4 billion and a pointer to an already-destroyed object.

The logic to set `m_references` to 1 is now moved into `subtract_reference` so that derived classes don't have to know this One Weird Trick. The actual decrementing happens in `subtract_final_reference`, which is the version that the `root_implements` destructor calls, since it knows that the object is finally going away and isn't preparing for destruction.

Setting `m_references` to 1 had been done with `memory_order_cst` (default), but I think `memory_order_relaxed` is sufficient. Since the external reference count is zero, there are no other threads with access to the object (outstanding weak references will not resolve), so there are no data dependencies.

Created new tests to validate the "weak reference created during destruction" code path, as well as to validate that weak references to coroutines work. (The `weak,coroutine` test failed prior to the fix.)

### DIFF
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -831,7 +831,7 @@ namespace winrt::impl
         virtual ~root_implements() noexcept
         {
             // If a weak reference is created during destruction, this ensures that it is also destroyed.
-            subtract_reference();
+            subtract_final_reference();
         }
 
         int32_t __stdcall GetIids(uint32_t* count, guid** array) noexcept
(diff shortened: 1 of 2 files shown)


kennykerr (approved):
Thanks!

PR #1098, opened 2022-01-29
Support borrowed ranges

sylveon:
This marks C++/WinRT ranges as "borrowed": they can be taken by value and iterators to them can be returned without dangling. It's safe for C++/WinRT because collections are reference types, so the lifetime issues typical of C++ collections do not apply.

This enables range algorithms that would previously return `std::ranges::dangling` (a dummy object that isn't dereferencable) to return the appropriate iterator instead. For example, `std::ranges::max_element` or `std::ranges::find` would be quite useless without this, as you can't do anything useful without the returned iterator.

Added tests confirming this works, as well.

The inner range namespace in the codegen is a workaround for [a compiler bug](https://developercommunity.visualstudio.com/t/symbol-cannot-be-defined-within-namespa/1650563?entry=myfeedback).

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -158,8 +158,15 @@ namespace cppwinrt
         }
     }
 
+    template<std::size_t indents = 0>
     static void write_close_namespace(writer& w)
     {
+        for (std::size_t i = 0; i < indents; ++i)
+        {
+            auto format = R"(    )";
+            w.write(format);
+        }
+
         auto format = R"(}
 )";
 
@@ -184,6 +191,15 @@ namespace cppwinrt
 )");
 
         return { w, write_close_namespace };
+    }
+
+    [[nodiscard]] static finish_with wrap_ranges_namespace(writer& w)
+    {
+        w.write(R"(    namespace ranges
+    {
+)");
+
+        return { w, write_close_namespace<1> };
     }
 
     [[nodiscard]] static finish_with wrap_type_namespace(writer& w, std::string_view const& ns)
@@ -3247,6 +3263,19 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
             auto generics = type.GenericParam();
 
             w.write("    template<%> struct formatter<%, wchar_t> : formatter<winrt::Windows::Foundation::IStringable, wchar_t> {};\n",
+                bind<write_generic_typenames>(generics),
+                type);
+        }
+    }
+
+    static void write_std_enable_borrowed_range(writer& w, TypeDef const& type)
+    {
+        std::string_view iiterable = "Windows.Foundation.Collections.IIterable`1";
+        if (type_name(type) == iiterable || implements_interface(type, iiterable))
+        {
+            auto generics = type.GenericParam();
+
+            w.write("        template<%> inline constexpr bool enable_borrowed_range<%> = true;\n",
                 bind<write_generic_typenames>(generics),
                 type);
         }
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -158,15 +158,8 @@ namespace cppwinrt
         }
     }
 
-    template<std::size_t indents = 0>
     static void write_close_namespace(writer& w)
     {
-        for (std::size_t i = 0; i < indents; ++i)
-        {
-            auto format = R"(    )";
-            w.write(format);
-        }
-
         auto format = R"(}
 )";
 
@@ -193,13 +186,13 @@ namespace cppwinrt
         return { w, write_close_namespace };
     }
 
-    [[nodiscard]] static finish_with wrap_ranges_namespace(writer& w)
+    [[nodiscard]] static finish_with wrap_std_ranges_namespace(writer& w)
     {
-        w.write(R"(    namespace ranges
-    {
+        w.write(R"(namespace std::ranges
+{
 )");
 
-        return { w, write_close_namespace<1> };
+        return { w, write_close_namespace };
     }
 
     [[nodiscard]] static finish_with wrap_type_namespace(writer& w, std::string_view const& ns)
@@ -3268,14 +3261,14 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
         }
     }
 
-    static void write_std_enable_borrowed_range(writer& w, TypeDef const& type)
+    static void write_enable_borrowed_range(writer& w, TypeDef const& type)
     {
         std::string_view iiterable = "Windows.Foundation.Collections.IIterable`1";
         if (type_name(type) == iiterable || implements_interface(type, iiterable))
         {
             auto generics = type.GenericParam();
 
-            w.write("        template<%> inline constexpr bool enable_borrowed_range<%> = true;\n",
+            w.write("    template<%> inline constexpr bool enable_borrowed_range<%> = true;\n",
                 bind<write_generic_typenames>(generics),
                 type);
         }
--- a/cppwinrt/helpers.h
+++ b/cppwinrt/helpers.h
@@ -502,8 +502,13 @@ namespace cppwinrt
     {
         for (auto&& impl : type.InterfaceImpl())
         {
-            const auto iface = impl.Interface();
-            if (iface.type() != TypeDefOrRef::TypeSpec && type_name(iface) == name)
+            auto iface = impl.Interface();
+            if (iface.type() == TypeDefOrRef::TypeSpec)
+            {
+                iface = iface.TypeSpec().Signature().GenericTypeInst().GenericType();
+            }
+
+            if (type_name(iface) == name)
             {
                 return true;
             }
--- a/strings/base_includes.h
+++ b/strings/base_includes.h
@@ -28,6 +28,10 @@
 #include <format>
 #endif
 
+#ifdef __cpp_lib_ranges
+#include <ranges>
+#endif
+
 #ifdef __cpp_lib_coroutine
 
 #include <coroutine>
--- a/strings/base_includes.h
+++ b/strings/base_includes.h
@@ -24,6 +24,12 @@
 #include <directxmath.h>
 #endif
 
+#ifdef __has_include
+#if __has_include(<version>)
+#include <version>
+#endif
+#endif
+
 #ifdef __cpp_lib_format
 #include <format>
 #endif
--- a/strings/base_iterator.h
+++ b/strings/base_iterator.h
@@ -119,11 +119,6 @@ namespace winrt::impl
             return it + n;
         }
 
-        friend fast_iterator operator-(difference_type n, fast_iterator it) noexcept
-        {
-            return it - n;
-        }
-
     private:
 
         T m_collection = nullptr;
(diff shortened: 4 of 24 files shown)


strings/base_includes.h
```diff
@@ -28,6 +28,10 @@
 #include <format>
 #endif
 
+#ifdef __cpp_lib_ranges
```
CaseyCarter, line 31: Pre-existing: I suggest "pre-loading" the feature-test macros with:
```c++
#ifdef __has_include
#if __has_include(<version>)
#include <version>
#endif
#endif
```
before testing them like this. (MSVC's STL defines all feature-test macros when any STL header is included, but that behavior is not required by the C++ Standard and hence not portable.)

strings/base_iterator.h
```diff
@@ -126,7 +126,7 @@ namespace winrt::impl
```
CaseyCarter, line 126: Pre-existing: `n - it` (the `operator-(difference_type n, fast_iterator it)` defined on lines 122-125 above) is completely bogus, and should be struck.

test/test_cpp20/ranges.cpp
```diff
@@ -45,4 +45,20 @@ TEST_CASE("ranges")
 
         REQUIRE((result == std::vector{ 2, 4, 6 }));
     }
+
+    // borrowable ranges. without enabling them, *result would fail to compile.
```
CaseyCarter, line 49: "Without <feature> this test would not compile" is pretty generic. (Nit: the standard terminology is actually "borrowed range" as made concrete by the `borrowed_range` concept, rather than "borrowable range".)

Maybe just "Validate that rvalue winrt collections are borrowed ranges"?

strings/base_iterator.h
sylveon: While you're right, I fear at this point it would be an API break. So I'll leave the final call to @kennykerr.

strings/base_includes.h
sylveon: Good idea, will add

test/test_cpp20/ranges.cpp
sylveon: Borrow is getting pretty overloaded these days haha. Will also change this.

strings/base_iterator.h
kennykerr: If Casey says its bogus, I think we should probably fix/remove it. Just include a test for whatever the correct behavior should be, if any.

sylveon:
I added the <version> include, removed the bogus operator on fast_iterator and updated comments in my test!

Had to mess around with the NuGet unit tests to get a full test run passing (they wouldn't build on VS 2022 before).

CaseyCarter (approved):

strings/base_iterator.h
```diff
@@ -126,7 +126,7 @@ namespace winrt::impl
 
     private:
 
-        T const* m_collection = nullptr;
```
kennykerr, line 129: This change would add a `AddRef`/`Release` bump to each iteration, which is why it's always been a reference. Is this strictly necessary?

sylveon: We need `fast_iterator` to hold a strong reference to the collection otherwise it would be dangling when returned and therefore cannot be borrowed (I initially forgot to do that and the unit tests found the issue right away because it crashed).

cppwinrt/code_writers.h
```diff
@@ -158,8 +158,15 @@ namespace cppwinrt
         }
     }
 
+    template<std::size_t indents = 0>
```
kennykerr, line 161: Rather than nesting namespaces, just use `namespace std::ranges` in a distinct top-level code gen block. That would be consistent with how the existing namespace are generated.

cppwinrt/helpers.h
```diff
@@ -502,8 +502,13 @@ namespace cppwinrt
...
         {
-            const auto iface = impl.Interface();
-            if (iface.type() != TypeDefOrRef::TypeSpec && type_name(iface) == name)
+            auto iface = impl.Interface();
+            if (iface.type() == TypeDefOrRef::TypeSpec)
+            {
+                iface = iface.TypeSpec().Signature().GenericTypeInst().GenericType();
+            }
+
+            if (type_name(iface) == name)
```
kennykerr, line 511: Can you explain why this change was needed?

test/test/fast_iterator.cpp
```diff
@@ -45,7 +45,7 @@ TEST_CASE("fast_iterator")
         REQUIRE(vbegin[2] == 4);
         REQUIRE(vbegin + 2 > vbegin);
         REQUIRE(2 + vbegin > vbegin);
-        REQUIRE(2 - (vbegin + 4) > vbegin);
+        REQUIRE(vbegin + 4 - 2 > vbegin);
```
kennykerr, line 48: Did this break?

strings/base_iterator.h
kennykerr: But that means folks using it in a range-base for loop are paying the price (the scenario for which the `fast_iterator` was originally conceived).

kennykerr: Although not expensive, just as you wouldn't expect to take a copy of a `std::vector` to make an iterator, you shouldn't make a copy of the collection interface either.

test/test/fast_iterator.cpp
CaseyCarter: Yes, this was testing the bogus "integer - iterator" operation. It probably makes more sense to delete the test case than to change it into a redundant test of other operations.

kennykerr (changes requested):
Let's figure out how to do this without costing more for existing code.

strings/base_iterator.h
sylveon: IIRC, range-based for loops use `++i`, not `i++`, so they don't pay the cost.

cppwinrt/helpers.h
sylveon: IIterable is a generic interface, and the previous code skipped over them

test/test/fast_iterator.cpp
sylveon: Yes, it was the bogus version. It's already covered down below (`vbegin + 2 - 2`), so removing this line makes sense

strings/base_iterator.h
kennykerr: I'm talking about the cost of creating the `fast_iterator` object.

sylveon: While it does technically get more costly, I feel like this one-time cost to be pretty minimal compared to the actual loop body, especially since WinRT is involved: every iteration of the ranged-for loop will do a AddRef/Release too because of GetAt. Not to mention that getting any property on the result of GetAt, or copying it elsewhere, etc. will also involve AddRef/Release.

A partial amount of the cost is solved by compiler-autogenerated move constructors/assignment operator.

The change also helps consistency between `fast_iterator` and `IIterator<T>` in a scenario like the following code.
```cpp
auto v = winrt::single_threaded_vector<int>({1, 2, 3});
auto it = begin(v);
v = somethingElse;
*it;
```
The `*it` expression will get the first value of whatever `somethingElse` contains. because `fast_iterator` holds a pointer to the variable `v` and `v` was reassigned after calling `begin(v)`.

If we compare against the same code, but boxing to `IIterable<int>` first:
```cpp
IIterable<int> v = winrt::single_threaded_vector<int>({1, 2, 3});
auto it = begin(v);
v = somethingElse;
*it;
```
Now `*it` will get the first value from the vector that was created on line 1, because `IIterator<int>` holds a reference to the source vector. With this change, `fast_iterator` behaves the same as `IIterator<T>`

kennykerr: > every iteration of the ranged-for loop will do a AddRef/Release too because of GetAt

There's no `AddRef`/`Release` bump when calling `GetAt` on a collection of `int` as you've used in this example. The only cost is a single virtual function call, `GetAt` itself. Yes, if your loop body is expensive it may not matter that much. On the other hand, if you're just iterator a collection of `int` then it may be impactful. I'm not trying to be difficult, but I have customers that really really care about getting rid of every last unnecessary `AddRef`/`Release` pair. 

I'm not familiar with C++ ranges - did `std::vector` iterators suddenly get more expensive to accommodate ranges? That would be surprising.

sylveon: No, but for example in debug mode it's not cheap to build one, as iterator debugging takes a lock iirc. (debug iterators don't have a trivial destructor either).

We could experiment with overloading begin for rvalue refs (e.g. `auto begin() &&;`) to return an owning iterator (moving ownership into it) and leaving the normal begin to return a non-owning iterator, eliding an `AddRef`/`Release` for the case of borrowed ranges, and leaving for loops untouched. We would need to add `bool owning;` to `fast_iterator` as ranges wants the same iterator type everywhere, and `iterator_t` uses an lvalue ref to determine the iterator type, so returning a different type from our overload is not possible. That would certainly complexify the implementation of `fast_iterator`.

Or we could disable borrowed ranges for `IVector` and `IBindableVector` (forcing explicit ownership, or converting to `IIterable` and losing out on `fast_iterator`), but that will lead to inconsistency in the API that I'm not a huge fan of.

Obviously the easiest way is just staying as is and dealing with the additional `AddRef` (using WinRT in a performance critical loop is already not the best thing), but that's already been disqualified.

kennykerr: Sounds even less appealing. 😉 Why don't we go back to the beginning and deal with the question of why this is being done in the first place. There was no issue opened prior to this PR so I have no context. What fundamental problem in C++/WinRT does this solve? Why do we need to "support borrowed ranges"? Why the extra complexity and cost?

sylveon: Supporting borrowed ranges allows consumers to directly pass the return value of an API call to the `std::ranges` algorithms without needing to store it in a variable first, and the return value will still be usable. It basically signals to those algorithms that the type it's iterating is either non-owning (`std::span` and `std::string_view` for example), or has iterators that holds a strong reference to the underlying collection (WinRT collection types). It allows us to simplify code.

```cpp
PackageManager manager;
IIterator<Package> someAppPackage = std::ranges::find_if(manager.FindPackagesForUser(L""),
    [](const Package &package)
    {
        return package.Id().FullName() == L"SomeApp_8wekyb3d8bbwe";
    });

if (someAppPackage != nullptr)
{
    // valid and not dangling
    (*someAppPackage).Launch(L"");
}
```

Without specializing `enabled_borrowed_range` to say `true`, `std::ranges::find_if` in this example above would return `std::ranges::dangling`, which is a fake object that disallows any kind of useful operation, because it assumes the iterator does not hold a strong reference so the collection would be destroyed once `std::ranges:find_if` returns and the iterator would be dangling.

kennykerr: You can write that more simply without ranges. 

```C++
PackageManager manager;
for (auto package : manager.FindPackagesForUser(L""))
{
    if (package.Id().FullName() == L"SomeApp_8wekyb3d8bbwe")
    {
        package.Launch(L"");
    }
}
```

sylveon: True, albeit that was an example. There are obviously more complex algorithms that would be harder to write and/or maintain as a normal for loop.

kennykerr: Understood, it sounds like it really needs a larger investment to get ranges supported without compromising performance. We've spoken about making C++20 the minver in future. This sounds like a good feature to include in such a rejuvenation effort.

PR #1109, opened 2022-02-24
Support the format returned by StringFromGUID2 in guid.parse.

jlaanstra:
Support the format returned by StringFromGUID2 in guid.parse. ```{(})```  are common characters used in guid formats and it would be convenient if cppwinrt would automatically handle them.

### DIFF
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -87,8 +87,12 @@ WINRT_EXPORT namespace winrt
         template <typename TStringView>
         static constexpr guid parse(TStringView value)
         {
-            value.remove_prefix(std::min(value.find_first_not_of(" {("), v.size()));
-            value.remove_suffix(std::min(value.find_last_not_of(" })"), v.size()));
+            // Handle {} and ()
+            if (value.size() == 38 && ((value[0] == '{' && value[37] == '}') || (value[0] == '(' && value[37] == ')')))
+            {
+                value.remove_prefix(1);
+                value.remove_suffix(1);
+            }
 
             if (value.size() != 36 || value[8] != '-' || value[13] != '-' || value[18] != '-' || value[23] != '-')
             {
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -85,8 +85,11 @@ WINRT_EXPORT namespace winrt
     private:
 
         template <typename TStringView>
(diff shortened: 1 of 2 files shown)


strings/base_types.h
```diff
@@ -85,8 +85,11 @@ WINRT_EXPORT namespace winrt
     private:
 
         template <typename TStringView>
-        static constexpr guid parse(TStringView const value)
```
sylveon, line 88: Why remove `const` here?

sylveon: Oh, I missed that `value.remove_prefix` was called right under

strings/base_types.h
```diff
@@ -85,8 +85,11 @@ WINRT_EXPORT namespace winrt
...
+        static constexpr guid parse(TStringView value)
         {
+            value.remove_prefix(std::min(value.find_first_not_of(" {("), v.size()));
```
sylveon, line 90: I don't think this compiles when TStringView is std::wstring_view

strings/base_types.h
jlaanstra: Can't modify a const value.

strings/base_types.h
jlaanstra: You're right. Went with the simpler option.

strings/base_types.h
```diff
@@ -87,8 +87,12 @@ WINRT_EXPORT namespace winrt
         template <typename TStringView>
         static constexpr guid parse(TStringView value)
         {
-            value.remove_prefix(std::min(value.find_first_not_of(" {("), v.size()));
-            value.remove_suffix(std::min(value.find_last_not_of(" })"), v.size()));
+            // Handle {} and ()
+            if (value.size() == 38 && ((value[0] == '{' && value[37] == '}') || (value[0] == '(' && value[37] == ')')))
```
kennykerr, line 91: Cool! I know `{guid}` is common but I'm not sure I've seen `(guid)` very often... What's the motivation behind that?

jlaanstra: Based on what the .NET version of Guid.Parse supports: https://docs.microsoft.com/en-us/dotnet/api/system.guid.parse?view=net-6.0

kennykerr (approved):

Issue #631, opened 2020-05-20
Add support for the builder pattern


kennykerr:
WinRT properties that are writable should return a reference to self (`D const&`) to allow chaining calls to various properties.

MikeGitb:
Shouldn't that be a mutable reference then?

kennykerr:
No, the reference itself is not mutable. It is the COM object being pointed to by the reference that is changed.

eklipse2k8:
Usually in a builder pattern at least commonly used in Java/Android, the class you use to build with is a mutable builder class, that once you call .build(), will return an immutable non-builder class.

I think that WinRT properties in general are quite "unstable", as some setters actually trigger state changes. A builder should be a stand-in for a constructor, and allowing normal WinRT setters to operate as builders will make for a messy codebase.

Right now the **.prop(value)** setter pattern in cppwinrt puts the onus on the developer to understand side-effects on setters.

kennykerr:
Feel free to reopen if there's interest in exploring this further.

PR #1114, opened 2022-03-05
Enable producing fluent-style (chainable) property setters

asklar:
This adds an option to cppwinrt to produce fluent-style (builder pattern, i.e. chainable) property setters.
These are especially useful when dealing with XAML or Composition objects.

It turns something like:
```cpp
button.Width(100);
button.Height(20);
button.Text(L"Click me");
// ... etc
```

into the less verbose:
```cpp
button
  .Width(100)
  .Height(20)
  .Text(L"Click me");
```

This is achieved by returning the object's type from each setter (similar to how the winrt call shim works, it's just a cast).
This behavior is set up as an opt-in (to reduce the risk of breaking existing projects) through a new `-fluent` option (settable in the app project via an msbuild property `CppWinRTFluent`. Even if we made this behavior the default, I would not expect existing projects to break either, as nobody would have depended on the return type anyway.

(BTW, it looks like I lost the permissions to create branches in the microsoft/cppwinrt repo, so I created the branch in my fork)


sylveon:
This file need an update: https://github.com/microsoft/cppwinrt/blob/master/nuget/CppWinrtRules.Project.xml

cppwinrt/code_writers.h
```diff
@@ -1150,11 +1150,34 @@ namespace cppwinrt
         }
         else
         {
-            format = R"(    template <typename D%> WINRT_IMPL_AUTO(%) consume_%<D%>::%(%) const
+            if (settings.fluent && method.SpecialName() && method.Name()._Starts_with("put_"))
```
kennykerr, line 1153: You can use `is_put_overload`.

kennykerr:
In general, I think this is fine. I wanted to do this myself (#631) but never got around to exploring it further. 

I don't see any obvious downside to adding the `const&` return on all setters and in my experience having this as an opt-in negatively affects adoption and compatibility. Perhaps we should consider remove the option and just making it available everywhere. 

This being a potentially invasive change, I also suggest we take a test build and use it to build a few large code bases before adopting officially. Building Windows, WinUI, Terminal, etc. 

I'd also like to hear opinions from regular contributors/users.. @jlaanstra @Scottj1s @oldnewthing @DefaultRyan @sylveon @dfields-msft @ChrisGuzak @davidmatson

jlaanstra:
I helped @asklar with this on Friday. Change is fine with me.

> I don't see any obvious downside to adding the const& return on all setters and in my experience having this as an opt-in negatively affects adoption and compatibility. Perhaps we should consider remove the option and just making it available everywhere.

Perhaps have it as an opt-out in case it does cause problems anywhere even after our testing? I'll happily run a test build for my project once we have a build.

oldnewthing:
The ambiguity comes from the property getter.

```cpp
// 1
auto thing1 = BuildInfo().Major(3).Minor(1).Build();
// 2
auto thing2 = WidgetBuilder().Color(Colors::Blue()).Length(200).Build();
```

Thing1 is a read of the `BuildInfo.Build` property, whereas thing2 is the result of the `WidgetBuilder.Build()` method.

I personally don't think this is too confusing. It's not a common pattern to use the fluent pattern to set a bunch of properties and then read a property at the end.

sylveon:
I agree that on-by-default is a good to have here. I haven't tried building it with my projects, but nothing makes me believe it would break anything (since I never consume the result of a property setter, nor use the result as a way to identify setters in metaprogramming).

I don't think ambiguity with stuff implementing the builder pattern is a big problem - this is kinda what we're trying to reproduce here. For being able to look at the return value: fluent setters will return `BuildInfo const&`, while a `WidgetBuilder` will return `Widget`.

jevansaks:
> I would like to make it opt-in for now.

@asklar What is the plan for making it on by default if it's opt in "for now"? Could we have it on by default and have code bases opt out? I would rather not have a single line in the RNW codebase (which arguably should have been an if/else, conditional operator that returns void is weird, no?) dictate what we do.

kennykerr:
Keep in mind that breaking changes are not such a big deal anymore. I no longer push cppwinrt to any projects. Any project that wants an update grabs a newer version and deals with whatever breaking changes during ingestion. The problem with opt-in is that it will always be used by very few projects and thus never really get well tested but costs considerably more to maintain because now I have to support two different modes.

jonwis:
The mechanism seems OK. Is this only for properties, then? Agree that there's a small potential for source breaking changes; the solution would be to cast to void? Like:

```
m_foo ? (void)tb.Thing(6) : (void)ob.Muffin(3);
```

davidmatson:
@kennykerr - the cost angle is the same if there's an option, regardless of whether it's opt-in or opt-out, right?
(I thought the main question was whether to have this feature be opt-in vs opt-out.)

oldnewthing:
The cost is higher for opt-in. We've seen this play out with the existing `-opt` flag.

1. The sample code on docs.microsoft.com will talk about the new feature, and 80% of users will file bugs saying "It doesn't work."
2. Kenny / Ryan / I will then reply and say "You have to manually enable the feature," and then we're into the rathole of "I'm not using Visual Studio. I'm using (some random IDE you've never heard of). How do I enable the feature there?"
3. The documentation folks have to add verbiage about "To get this feature, you need to..." The questions of the form "I'm not using Visual Studio. I'm using (some random IDE)" will persist.
4. Most people won't read the documentation and ask the question anyway.
5. The problem never goes away. Every round of new C++/WinRT onboarding re-encounters the problem.

Note: We're very lucky if the bug says "Fluent property setting doesn't work." Usually, the bug is "I can't get this to compile" and then we have to dig in to figure out that the "this" that they can't get to compile is a fluent property setting.

The cost of opt-out is that a small number of projects will get compile-time errors when they upgrade, and they can fix them (by applying `(void)` casts), and then they're done and never have to think about it again.

Summary:
* Opt-out: The problem eventually goes away.
* Opt-in: The problem never goes away.
*

sylveon:
Side-effecting ternaries seems like bad style anyways.

kennykerr:
> Make fluent the default (possible to opt-out)

To be clear, I don't want an option. On-by-default is better than off-by-default but an option still introduces complexity that adds cost. Two modes to test, support, explain, document, diagnose, etc. If it's a breaking change, developers can use an older version until they're ready to adopt. If it's not a good idea, we shouldn't do it. If we're not sure its a good idea, that's why I suggested we build a few large projects to get a better feel for the impact of this change (we need to do that regardless).

test/test_component_fluent/test_component_fluent.idl
```diff
@@ -0,0 +1,21 @@
+import "Windows.Foundation.idl";
```
kennykerr, line 1: Since this is always-on, you can just add this to the existing `test_component` project and test it from the existing `test` project.

jlaanstra:
> > Make fluent the default (possible to opt-out)
> 
> To be clear, I don't want an option. On-by-default is better than off-by-default but an option still introduces complexity that adds cost. Two modes to test, support, explain, document, diagnose, etc. If it's a breaking change, developers can use an older version until they're ready to adopt. If it's not a good idea, we shouldn't do it. If we're not sure its a good idea, that's why I suggested we build a few large projects to get a better feel for the impact of this change (we need to do that regardless).

It seems reasonable to have the opt-out for some amount of time (3-6 months?) to give ourselves some time to find issues if any. After that time has passed we can remove the opt-out. Thoughts?

kennykerr:
Older versions of cppwinrt are available indefinitely to give developers as much time as they need.

kennykerr:
Opened #1123 to start planning cppwinrt v3 where we can support this feature.

PR #1116, opened 2022-03-10
Draft support for CRTP in class default interface

kennykerr:
This isn't quite ready but illustrates what's involved in providing CRTP support for a default interface so that it may in turn take advantage of CRTP to return the most derived type from property setters. 

The trouble is that it breaks a few assumptions that C++/WinRT makes around a runtime class actually being its default interface type. Those broken assumptions are highlighted in the tests I had to hack up to build and pass. 

I also removed `WINRT_IMPL_AUTO` as that was just a workaround for an IntelliSense bug that has long since been fixed.

But this change lets you successfully write code like this:

```C++
int main()
{
    Compositor compositor;
    AmbientLight light = compositor.CreateAmbientLight();

    light.Color(Colors::Red()).Intensity(0.5f);
}
```

### DIFF
--- a/test/old_tests/UnitTests/array.cpp
+++ b/test/old_tests/UnitTests/array.cpp
@@ -24,11 +24,11 @@ static IAsyncOperation<IDataReader> CreateDataReader(std::initializer_list<byte>
     writer.WriteByte(1);
     writer.WriteByte(2);
     writer.WriteByte(3);
-    co_await writer.StoreAsync();
+    co_await static_cast<IAsyncOperation<uint32_t>>(writer.StoreAsync());
 
     stream.Seek(0);
     DataReader reader(stream);
-    co_await reader.LoadAsync(3);
+    co_await static_cast<IAsyncOperation<uint32_t>>(reader.LoadAsync(3));
     co_return reader;
 }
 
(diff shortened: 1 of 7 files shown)


test/old_tests/UnitTests/array.cpp
```diff
@@ -24,11 +24,11 @@ static IAsyncOperation<IDataReader> CreateDataReader(std::initializer_list<byte>
     writer.WriteByte(1);
     writer.WriteByte(2);
     writer.WriteByte(3);
-    co_await writer.StoreAsync();
+    co_await static_cast<IAsyncOperation<uint32_t>>(writer.StoreAsync());
```
kennykerr, line 27: This we could work around easily enough.

test/test_component_derived/Nested.HierarchyD.cpp
```diff
@@ -21,10 +21,10 @@ namespace winrt::test_component_derived::Nested::implementation
         test_component_base::HierarchyA a = *this;
         assert(a);
 
-        Nested::IHierarchyD id = *this;
+        Nested::IHierarchyD id = d;
```
kennykerr, line 24: This is harder to solve, perhaps not that important but I'm not sure how much code depends on this working.

PR #1120, opened 2022-03-11
Basic support for chained properties

### LINKED ISSUE #631: Add support for the builder pattern

kennykerr:
This is a simple and pragmatic solution to support the chained/fluent properties that @asklar has been experimenting with (#1114). 

The basic idea is that property setters can rely on CRTP to return the most derived type reference in all cases except where the property belongs to a class's default interface. This case is interesting because the default interface is directly inherited by the class and as such does not play along with C++/WinRT's usual CRTP pattern. To remedy this, I've "hardcoded" exclusive interfaces to always return the exclusive class type. This solves the main issue in implementing chained properties where the class type can be lost along the way. 

This is much cleaner and simpler than the other approaches I've tried, but it falls short in a class hierarchy as you end up with the last class, not necessarily the most derived class. But if you order the properties correctly it works just fine. This is clearly not ideal but perhaps a wart we can fix in future. Here's an example:

```C++
using namespace winrt;
using namespace Windows::UI::Composition;

int main()
{
    Compositor compositor;
    CompositionBrush brush = compositor.CreateColorBrush();

    SpriteVisual visual = compositor.CreateSpriteVisual();
    visual.Opacity(0.5f).IsVisible(true).Brush(brush);
}
```

Here you get the following error:

`error C2039: 'Brush': is not a member of 'winrt::Windows::UI::Composition::Visual'`

The trouble is that `Opacity` and `IsVisible` are properties of the `Visual` base class and `Brush` is a property of the derived `SpriteVisual`. The fix is simple enough, just set the `Brush` property before the `Visual` properties:

```C++
visual.Brush(brush).Opacity(0.5f).IsVisible(true);
```

While not perfect, it is far less invasive than the other solutions I experimented with. 

Fixes #631

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1124,6 +1124,20 @@ namespace cppwinrt
         auto method_name = get_name(method);
         method_signature signature{ method };
         auto async_types_guard = w.push_async_types(signature.is_async());
+        std::string put_return;
+
+        if (is_put_overload(method))
+        {
+            if (auto attribute = get_attribute(type, "Windows.Foundation.Metadata", "ExclusiveToAttribute"))
+            {
+                auto class_name = get_attribute_value<ElemSig::SystemType>(attribute, 0).name;
+                put_return = w.write_temp("\n        return reinterpret_cast<winrt::@ const&>(*this);", class_name);
+            }
+            else
+            {
+                put_return = "\n        return static_cast<D const&>(*this);";
+            }
+        }
 
         std::string_view format;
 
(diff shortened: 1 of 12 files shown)


jonwis (approved):

jlaanstra (approved):

kennykerr:
Unfortunately, this cast is causing a conversion operator to be called which results in a copy and thus an `AddRef`/`Release` bump making this quite costly. Back to the drawing board.

Issue #1121, opened 2022-03-15
Debug visualizer does not expand nested runtime class properties - VS2019/VS2022 regression


Scottj1s:
When the debug visualizer was first released with VS2017, it would expand nested runtime class properties.  With VS2019/VS2022, this functionality no longer works.

**Repro:**
1. Install the [C++/WinRT VS extension](https://marketplace.visualstudio.com/items?itemName=CppWinRTTeam.cppwinrt101804264) in VS 2019 or VS 2022 (despite the name, it supports both).
2. Create a simple C++/WinRT console app from the extension template, and add a query parameter to the uri:
```
#include "pch.h"

using namespace winrt;
using namespace Windows::Foundation;

int main()
{
    init_apartment();
    Uri uri(L"http://aka.ms/cppwinrt?foo=bar");
    printf("Hello, %ls!\n", uri.AbsoluteUri().c_str());
}

```
3. Set a BP on the printf
4. Build and F5 to debug
5. In a watch window, type "uri"
6. Expand the visualization for uri
7. Expand the nested visualization for the QueryParsed property

**Expected results:**
The QueryParsed.Size property should be displayed, with value '1'

**Actual results:**
No expansion of QueryParsed occurs.  

**Notes:**

In debugging the app with VS2017, the expansion results in BPs on both [cppwinrt_visualizer::GetChildren](https://github.com/microsoft/cppwinrt/blob/e0012b04cd3833c944984e681633df5fde6311ae/natvis/cppwinrt_visualizer.cpp#L259) and [cppwinrt_visualizer::GetItems](https://github.com/microsoft/cppwinrt/blob/e0012b04cd3833c944984e681633df5fde6311ae/natvis/cppwinrt_visualizer.cpp#L296) being hit.  But with VS2019/VS2022, only cppwinrt_visualizer::GetChildren gets called - cppwinrt_visualizer::GetItems is never called and no child items (properties) are displayed.

PR #1126, opened 2022-03-23
fix regression in nested visualizations with VS2019+

Scottj1s:
per advice from the VSSDK team, made sure GetChildren returns requested items, not just relying on GetItems.  (this lazy technique worked for VS2017, but does not with VS2019+)

### DIFF
--- a/natvis/object_visualizer.cpp
+++ b/natvis/object_visualizer.cpp
@@ -1,5 +1,4 @@
 #include "pch.h"
-#include <algorithm>
 #include <variant>
 #include "object_visualizer.h"
 #include "property_visualizer.h"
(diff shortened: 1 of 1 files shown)


Scottj1s:
fixes #1121

kennykerr:
The build failures seem to have started with the previous commit:

https://github.com/microsoft/cppwinrt/runs/5652229311

Scottj1s:
> The build failures seem to have started with the previous commit:
> 
> https://github.com/microsoft/cppwinrt/runs/5652229311

but intermittently - only 1 of 3 runs has failed.  something may have changed with catch and its assumptions of CRT compat.  will investigate.

kennykerr:
> out of disk space (sigh... PCH files).

Need modules. 😋

Scottj1s:
Looks like catch occasionally fails with the v143 toolset - backing off for now: #1127

kennykerr (approved):

PR #1127, opened 2022-03-23
catch is evidently incompatible with VS 2022 (platform toolset v143) - backing off to unblock

Scottj1s:
Looked at latest version of catch, 2.13.8, and saw no significant differences from 2.13.7 (checked in here) to suggest that would help.  So backing test collateral off to platform toolset v142 (VS1029) to unblock build for now.

### DIFF
--- a/Directory.Build.Props
+++ b/Directory.Build.Props
@@ -3,9 +3,9 @@
   <!-- Set common MSBuild properties and item metadata for CppWinRT tool and tests -->
 
   <PropertyGroup>
-    <PlatformToolset>v143</PlatformToolset>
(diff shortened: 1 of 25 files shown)


sylveon:
I don't see anything in catch's issue tracker that suggests it doesn't work with VS 2022 - what issue are we seeing here?

Scottj1s:
@sylveon Me neither, but backing off to the 142 toolset appears to be a reliable workaround.  Issues we're seeing are occasional crashes of the test binaries ("SIGSEGV - Segmentation violation signal") across random test cases.

sylveon:
Ah, I've been seeing those too but just scratched it off to my computer being weird since AzDO was running tests fine.

PR #1128, opened 2022-03-24
Provide implementation-side visualization (per request), and always show raw data

Scottj1s:
Added visualization for winrt::impl::producer<...> subclasses (implementations), per several requests
Always show raw data, for classes that inherit IInspectable but metadata is not available

### DIFF
--- a/natvis/object_visualizer.h
+++ b/natvis/object_visualizer.h
@@ -20,6 +20,12 @@ enum class PropertyCategory
     Class,
 };
 
+enum class ObjectType
+{
+    Abi,
+    Projection,
+};
+
 // Metatdata for resolving a runtime class property value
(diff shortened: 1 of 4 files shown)


natvis/object_visualizer.h
```diff
@@ -20,6 +20,12 @@ enum class PropertyCategory
     Class,
 };
 
+enum class ObjectType
```
Scottj1s, line 23: this change from bool can be ignored - thought I'd need a tri-state for implementations, but just decided to keep it.

the only real meat is winrt::impl::producer addition and removing the condition on expanding (which permits raw data to always be visible)

kennykerr (approved):
Neat!

PR #1130, opened 2022-03-30
two-phase initialization support to prevent double-destruction on handing out this pointer in ctor

Scottj1s:
Initialization routines for components are now called after construction, rather than during.  This prevents 'this' pointers from being handed out and retained during construction, leading to multiple destructor calls.  If a component has an initialize_instance function (hopefully ugly enough named to prevent need for opting out), member or free (so that the Xaml compiler doesn't require a coordinated fix), then this function is called after the implementation is completely constructed.  If the initializer throws, the instance is destructed and the exception propagates as before.  Existing code can now remove InitializeComponent calls from class constructors, or override in order to access Xaml properties after initialization.  The support is general - non-Xaml code can also define initialize_instance.

### DIFF
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -862,14 +862,6 @@ catch (...) { return winrt::to_hresult(); }
 #if defined(WINRT_FORCE_INCLUDE_%_XAML_G_H) || __has_include("%.xaml.g.h")
 
 #include "%.xaml.g.h"
-namespace winrt::@::implementation
-{
-    template <typename D, typename... Args>
-    auto initialize_instance(%T<D, Args...>& instance)
-    {
-        static_cast<D&>(instance).InitializeComponent();
-    }
-}
 
 #else
 
@@ -891,8 +883,6 @@ namespace winrt::@::implementation
                 upper,
                 include_path,
                 include_path,
-                type_namespace,
-                type_name,
                 type_namespace,
                 type_name,
                 type_name);
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -860,7 +860,17 @@ catch (...) { return winrt::to_hresult(); }
         {
             auto format = R"(
 #if defined(WINRT_FORCE_INCLUDE_%_XAML_G_H) || __has_include("%.xaml.g.h")
+
 #include "%.xaml.g.h"
+namespace winrt::@::implementation
+{
+    template <typename D, typename... Args>
+    auto initialize_instance(%T<D, Args...>& instance)
+    {
+        static_cast<D&>(instance).InitializeComponent();
+    }
+}
+
 #else
 
 namespace winrt::@::implementation
@@ -881,6 +891,8 @@ namespace winrt::@::implementation
                 upper,
                 include_path,
                 include_path,
+                type_namespace,
+                type_name,
                 type_namespace,
                 type_name,
                 type_name);
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -1240,7 +1240,7 @@ namespace winrt::impl
             }
             catch (...)
             {
-                delete instance;
+                instance->Release();
                 throw;
             }
         }
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -1228,10 +1228,10 @@ namespace winrt::impl
         static constexpr bool value = get_value<T>(0);
     };
 
-
-    template <typename T>
-    auto initialize(T* instance)
+    template<typename T, typename... Args>
+    auto create_and_initialize(Args&&... args)
     {
+        auto instance = new heap_implements<T>(std::forward<Args>(args)...);
         if constexpr (has_initializer<T>::value)
         {
             try
@@ -1262,7 +1262,7 @@ namespace winrt::impl
 
         if constexpr (!has_static_lifetime_v<D>)
         {
-            return { to_abi<result_type>(initialize(new heap_implements<D>)), take_ownership_from_abi };
+            return { to_abi<result_type>(create_and_initialize<D>()), take_ownership_from_abi };
         }
         else
         {
@@ -1276,7 +1276,7 @@ namespace winrt::impl
                 return { result, take_ownership_from_abi };
             }
 
-            result_type object{ to_abi<result_type>(initialize(new heap_implements<D>)), take_ownership_from_abi };
+            result_type object{ to_abi<result_type>(create_and_initialize<D>()), take_ownership_from_abi };
 
             static slim_mutex lock;
             slim_lock_guard const guard{ lock };
@@ -1322,17 +1322,17 @@ WINRT_EXPORT namespace winrt
         }
         else if constexpr (impl::has_composable<D>::value)
         {
-            impl::com_ref<I> result{ to_abi<I>(impl::initialize(new impl::heap_implements<D>(std::forward<Args>(args)...))), take_ownership_from_abi };
+            impl::com_ref<I> result{ to_abi<I>(impl::create_and_initialize<D>(std::forward<Args>(args)...)), take_ownership_from_abi };
             return result.template as<typename D::composable>();
         }
         else if constexpr (impl::has_class_type<D>::value)
         {
             static_assert(std::is_same_v<I, default_interface<typename D::class_type>>);
-            return typename D::class_type{ to_abi<I>(impl::initialize(new impl::heap_implements<D>(std::forward<Args>(args)...))), take_ownership_from_abi };
+            return typename D::class_type{ to_abi<I>(impl::create_and_initialize<D>(std::forward<Args>(args)...)), take_ownership_from_abi };
         }
         else
         {
-            return impl::com_ref<I>{ to_abi<I>(impl::initialize(new impl::heap_implements<D>(std::forward<Args>(args)...))), take_ownership_from_abi };
+            return impl::com_ref<I>{ to_abi<I>(impl::create_and_initialize<D>(std::forward<Args>(args)...)), take_ownership_from_abi };
         }
     }
 
@@ -1354,7 +1354,7 @@ WINRT_EXPORT namespace winrt
         }
         else
         {
-            return { impl::initialize(new impl::heap_implements<D>(std::forward<Args>(args)...)), take_ownership_from_abi };
+            return { impl::create_and_initialize<D>(std::forward<Args>(args)...), take_ownership_from_abi };
         }
     }
 
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -1218,6 +1218,56 @@ namespace winrt::impl
     };
 #endif
 
+    template<typename T>
+    class has_initialize_instance_member
+    {
+        template <typename U, typename = decltype(std::declval<U>().initialize_instance())> static constexpr bool get_value(int) { return true; }
+        template <typename> static constexpr bool get_value(...) { return false; }
+
+    public:
+        static constexpr bool value = get_value<T>(0);
+    };
+
+    template<typename T>
+    class has_initialize_instance_free
+    {
+        template <typename U, typename = decltype(initialize_instance(*(U*)nullptr))> static constexpr bool get_value(int) { return true; }
+        template <typename> static constexpr bool get_value(...) { return false; }
+
+    public:
+        static constexpr bool value = get_value<T>(0);
+    };
+
+    template <typename T>
+    auto initialize_instance(T* instance)
+    {
+        if constexpr (has_initialize_instance_member<T>::value)
+        {
+            try
+            {
+                instance->initialize_instance();
+            }
+            catch (...)
+            {
+                delete instance;
+                throw;
+            }
+        }
+        else if constexpr (has_initialize_instance_free<T>::value)
+        {
+            try
+            {
+                initialize_instance(*instance);
+            }
+            catch (...)
+            {
+                delete instance;
+                throw;
+            }
+        }
+        return instance;
+    }
+
     inline com_ptr<IStaticLifetimeCollection> get_static_lifetime_map()
     {
         auto const lifetime_factory = get_activation_factory<impl::IStaticLifetime>(L"Windows.ApplicationModel.Core.CoreApplication");
@@ -1233,7 +1283,7 @@ namespace winrt::impl
 
         if constexpr (!has_static_lifetime_v<D>)
         {
-            return { to_abi<result_type>(new heap_implements<D>), take_ownership_from_abi };
+            return { to_abi<result_type>(initialize_instance(new heap_implements<D>)), take_ownership_from_abi };
         }
         else
         {
@@ -1247,7 +1297,7 @@ namespace winrt::impl
                 return { result, take_ownership_from_abi };
             }
 
-            result_type object{ to_abi<result_type>(new heap_implements<D>), take_ownership_from_abi };
+            result_type object{ to_abi<result_type>(initialize_instance(new heap_implements<D>)), take_ownership_from_abi };
 
             static slim_mutex lock;
             slim_lock_guard const guard{ lock };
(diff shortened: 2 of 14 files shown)


sylveon:
Is there such an example of this pointers being handed out and causing double destruction?

Also, isn't this change potentially source breaking? e.g. after updating cppwinrt now InitializeComponent will be called twice unless the developer goes and update their code for all controls (which could be a lot).

In any case, I'm not sure how I feel about moving InitializeComponent out of the constructor, since it adds a place where C# and C++ code for the same thing diverges (making porting or moving between languages harder)

kennykerr:
I never liked this pattern. Not sure I like the idea of adding it now...

cppwinrt/component_writers.h
```diff
@@ -860,7 +860,17 @@ catch (...) { return winrt::to_hresult(); }
...
             auto format = R"(
 #if defined(WINRT_FORCE_INCLUDE_%_XAML_G_H) || __has_include("%.xaml.g.h")
+
 #include "%.xaml.g.h"
+namespace winrt::@::implementation
+{
+    template <typename D, typename... Args>
+    auto initialize_instance(%T<D, Args...>& instance)
+    {
+        static_cast<D&>(instance).InitializeComponent();
```
kennykerr, line 870: Clearly this is in support of Xaml. Do we really want to advertise this for other uses?

strings/base_implements.h
```diff
@@ -1218,6 +1218,56 @@ namespace winrt::impl
     };
 #endif
 
+    template<typename T>
+    class has_initialize_instance_member
```
kennykerr, line 1222: Do we really need this level of flexibility? member vs free? Why not keep it as simple as possible?

vsix/ProjectTemplates/VC/Windows Universal/BlankApp/App.h
```diff
@@ -6,7 +6,8 @@ namespace winrt::$safeprojectname$::implementation
     struct App : AppT<App>
     {
         App();
-
+        void InitializeComponent();
+        void initialize_instance(){ InitializeComponent(); }
```
kennykerr, line 10: This looks really hacky. Can't we make this a little more polished? Just specialize this for Xaml-derived classes that actually have a InitializeComponent method?

kennykerr:
> Initialization routines for components are now called after construction, rather than during. This prevents 'this' pointers from being handed out and retained during construction, leading to multiple destructor calls.

Can you provide a minimal example of this problem? Maybe there's a less intrusive solution.

oldnewthing:
@sylveon @kennykerr The scenario for this is as follows:

```cpp
namespace winrt::Contoso::implementation
{
    struct MyPage : MyPageT<MyPage>
    {
        MyPage()
        {
            this->InitializeComponent();
            something_that_throws();
        }
    };
}

template<typename D, typename... I>
void MyPageT<D, I...>::InitializeComponent()
{
    if (!_contentLoaded)
    {
        _contentLoaded = true;
        Uri resourceLocator{ L"ms-appx:///MyPage.xaml" };
        Application::LoadComponent(*this, resourceLocator, ComponentResourceLocation::Application);
    }
}
```

The `InitializeComponent` method hands out a reference to a partially constructed `this` to `Application::LoadComponent`, which dutifully calls `AddRef` to retain the reference.

And then `something_that_throws();` throws an exception, which causes construction to fail, and the object is forcible destroyed, leaving `LoadComponent` with a dangling reference.

The problem is that `InitializeComponent()` is writing a check it may not be able to cover: It is calling `LoadComponent` with an `IInspectable`, which comes with the promise that `AddRef` can be used to extend its lifetime. But since the object is still constructing, the constructor may fail, and then the object forcibly goes away, ignoring any reference count.

This is the canonical example, but there are other cases, such as event registration with a strong reference, or setting up bindings, which have the same root cause.

jlaanstra:
> 

Does Application.LoadComponent hold on to ```this``` after the LoadComponent call returns? That seems surprising to me. How would I tell xaml to give up that reference to allow my class to get destructed?

Scottj1s:
> Also, isn't this change potentially source breaking? e.g. after updating cppwinrt now InitializeComponent will be called twice unless the developer goes and update their code for all controls (which could be a lot).

Generally, that would be a concern.  But in this case (InitializeComponent, the motivation), redundant calls are idempotent.

Scottj1s:
> In any case, I'm not sure how I feel about moving InitializeComponent out of the constructor, since it adds a place where C# and C++ code for the same thing diverges (making porting or moving between languages harder)

Yes, the behavior is different for C# and C++.  They're different languages and porting between them has never been trivial.

Scottj1s:
> I never liked this pattern. Not sure I like the idea of adding it now...

The trick is how to handle initialization outside construction, without causing major disruption to existing code.

cppwinrt/component_writers.h
Scottj1s: Not necessarily - I could be persuaded to remove the general use.

strings/base_implements.h
Scottj1s: Yes, could remove member (free is required to avoid a coordinate change with Xaml compiler).

vsix/ProjectTemplates/VC/Windows Universal/BlankApp/App.h
Scottj1s: We could remove this altogether for App, since it has an empty InitializeComponent.  My concern was with introducing any kind of breaking change.  The Xaml compiler may choose to provide an implementation at some point.  But yes, could remove the generality and just specialize for InitializeComponent.

Scottj1s:
> Can you provide a minimal example of this problem? Maybe there's a less intrusive solution.

Unfortunately, I don't have a repro (they're hard to come by for this issue).  But the root cause is handing out 'this' pointers from a constructor which is a Bad Idea.  Totally open to alternatives.  But moving InitializeComponent out of the constructor is going to be somewhat disruptive.  Xaml property accessors, like MyButton(), must not be called before InitializeComponent.  So they would have to move along with the call, and that at minimum requires support for user overriding.

Scottj1s:
> Does Application.LoadComponent hold on to `this` after the LoadComponent call returns? That seems surprising to me. How would I tell xaml to give up that reference to allow my class to get destructed? Seems like as long as the LoadComponent call doesn't hold on to its own copy of `this` there is no issue?

Yes, it certainly does, via a PegNoRef call in the Xaml runtime.  And ordinarily, that would be paired with an UnpegNoRef to take Xaml's hands off the same raw pointer.  But in the case of premature object tear-down due to exceptions in construction (which can originate from winrt::check_hresult calls in the Xaml runtime as well), we then have a dangling pointer and a time-bomb when UnpegNoRef attempts to release it.

sylveon:
Don't we hit the same double-free issue with this PR?

```cpp
void MyControl::InitializeComponent()
{
    MyControlT::InitializeComponent();
    something_that_throws();
}
```

Because the XAML framework holds a reference to it but then something_that_throws causes C++/WinRT to immediately call delete on it, ignoring any ref counting going on:
https://github.com/microsoft/cppwinrt/blob/ef7c5341912a8d77031dd63974bc044e93abf605/strings/base_implements.h#L1252

ChrisGuzak:
This is a classic problem that is not specific to Xaml. There is a variation of this for free-threaded events, where the event is raised before the object is fully constructed. The advice we give is never hand out a reference to this in the constructor, use 2 phase construction based on a static factory function for the class. 

Another approach would make this the developer's problem, ask them to opt into this behavior.

test/test/initialize_instance.cpp
```diff
@@ -0,0 +1,165 @@
...
+    struct MemberInitialize : implements<MemberInitialize, IStringable>
+    {
+        bool& m_initialize_called;
+
+        MemberInitialize(bool& initialize_called) : m_initialize_called(initialize_called)
+        {
+        }
+
+        ~MemberInitialize()
+        {
```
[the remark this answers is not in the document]
kennykerr: Lol indeed - I just meant for the test. 😉

Scottj1s:
> Don't we hit the same double-free issue with this PR?

No, we don't.  The reason this is safe is because we've eliminated the overlap between exception-based destruction and storing an object pointer - all from within the constructor chain.  Here, the object that's held indirectly by InitializeComponent is fully constructed, with proper refcount bookkeeping.  The unit test exercises this scenario.

ChrisGuzak:
@Scottj1s I'd like this to be available for use outside of Xaml. I've not looked at the change yet to see if that is viable.

jlaanstra:
> @Scottj1s I'd like this to be available for use outside of Xaml. I've not looked at the change yet to see if that is viable.

Agree with this. It sounds like this is useful outside of Xaml. Could we make this similar to final_release, not required, but if you want to use it it's there?

Scottj1s:
@ChrisGuzak, @jlaanstra - there is some utility to this functionality, but I'm persuaded we shouldn't go out of our way to advertise it.  So I'll standardize on Xaml's function name, InitializeComponent, versus the more winrt-friendly name, initialize_instance.  If any class defines (or derived from a class defining) an InitializeComponent, that function will be called outside construction.

Scottj1s:
Second iteration - all component template ctors are now empty and defined inline.  This reduces the temptation to do anything interesting in that constructor.  E.g., if a developer drops in a call to MyButton(), it will fail without a preceding InitializeComponent(), which is what we're trying to discourage.

jlaanstra:
> > Does Application.LoadComponent hold on to `this` after the LoadComponent call returns? That seems surprising to me. How would I tell xaml to give up that reference to allow my class to get destructed? Seems like as long as the LoadComponent call doesn't hold on to its own copy of `this` there is no issue?
> 
> Yes, it certainly does, via a PegNoRef call in the Xaml runtime. And ordinarily, that would be paired with an UnpegNoRef to take Xaml's hands off the same raw pointer. But in the case of premature object tear-down due to exceptions in construction (which can originate from winrt::check_hresult calls in the Xaml runtime as well), we then have a dangling pointer and a time-bomb when UnpegNoRef attempts to release it.

Could UnpefNoRef simply be inside a wil::scope_exit block so it release the pointer when the stack unwinds avoiding the dangling pointer? Or are you saying that xaml holds on to ```this``` even after InitializeComponent returns? If it holds on even after InitializeComponent returns how does Xaml decide to release the pointer eventually and not cause a memory leak?

sylveon:
> Here, the object that's held indirectly by InitializeComponent is fully constructed, with proper refcount bookkeeping

But in `winrt::impl::initialize_instance` the object is `delete`'d without checking the ref count at all if the initializer throws. So unless I'm missing something, the object is deleted while Xaml still holds a reference to it, then Xaml calls `Release()` on an already deleted object. This seems especially true if there is user code that may throw after the Xaml InitializeComponent function.

> but I'm persuaded we shouldn't go out of our way to advertise it. 

I think the feature should be at least documented because it's important to let devs know that Xaml elements can't be manipulated in the constructor anymore, and there is a "magic name" they need to use to be able to do that.

strings/base_implements.h
```diff
@@ -1218,6 +1218,35 @@ namespace winrt::impl
...
+        template <typename U, typename = decltype(std::declval<U>().InitializeComponent())> static constexpr bool get_value(int) { return true; }
+        template <typename> static constexpr bool get_value(...) { return false; }
+
+    public:
+        static constexpr bool value = get_value<T>(0);
+    };
+
+
+    template <typename T>
+    auto initialize(T* instance)
```
sylveon, line 1233: We could fold `new heap_implements<D>` into this function and simplify it:

```cpp
template<typename T, typename... Args>
auto create_and_initialize(Args&&... args)
{
    auto instance = std::make_unique<heap_implements<T>>(std::forward<Args>(args)...);
    if constexpr (has_initializer<T>::value)
    {
        instance->InitializeComponent();
    }

    return instance.release();
}
```

strings/base_implements.h
```diff
@@ -1240,7 +1240,7 @@ namespace winrt::impl
             }
             catch (...)
             {
-                delete instance;
+                instance->Release();
```
sylveon, line 1243: I am not sure this is better. Consider the following:

- Xaml InitializeComponent succeeds and calls AddRef.
- User code throws
- This catch block calls Release.
- Exception propagates and Xaml calls Release.

Now instead of double-delete we get mismatched AddRef/Release, which will eventually cause a use-after-free.

It also assumes that the InitializeComponent function only takes *one* strong reference. User code could very well be passing a strong reference to some event handler, and that would make it more than one strong ref. Or user code could be overriding InitializeComponent without taking a strong ref anywhere (for example, to be able to use winrt::fire_and_forget as part of initialization).

jlaanstra: Seems like InitializeComponent throwing would be like any other function throwing. We don't do Release anywhere else so is the Release even needed?

kennykerr: No, I believe this is now correct. The Release here is unrelated to the AddRef that Xaml or anyone else might take. This Release is paired with the initial reference count that the implements class template starts off with so there's no mismatch.

vsix/ProjectTemplates/VC/Windows Universal/BlankApp/MainPage.cpp
```diff
@@ -7,11 +7,6 @@ using namespace Windows::UI::Xaml;
 
 namespace winrt::$safeprojectname$::implementation
 {
-    MainPage::MainPage()
-    {
-        InitializeComponent();
```
jlaanstra, line 12: Removing this will be confusing for developers thinking that this is missing and they will add it back. Can we keep the constructor with a comment that cppwinrt automatically calls InitializeComponent() ?

kennykerr (approved):

strings/base_implements.h
sylveon: Ah! This makes more sense.

strings/base_implements.h
kennykerr: That would use `delete` which is incorrect.

vsix/ProjectTemplates/VC/Windows Universal/BlankApp/MainPage.cpp
sylveon: That's why I wasn't a fan, it changes long-established practice (in both C# and C++/CX)

kennykerr: That's why I wasn't a fan of the original example - making it different - but this now just removes boilerplate code and doesn't replace it with anything - it's a net win.

strings/base_implements.h
sylveon: (maybe add a comment to clarify that however)

vsix/ProjectTemplates/VC/Windows Universal/BlankApp/MainPage.cpp
sylveon: Yep, that's a win, but should be documented somewhere (as well as the need to "override" InitializeComponent if XAML access during construction (for lack of a better word) is required).

strings/base_implements.h
sylveon: That was sent before the change to `->Release()` - now we can change it for `com_ptr` to be correct again

strings/base_implements.h
Scottj1s: We won't have mismatched refcounts.  The Release here is to match the constructor-initialized AddRef (to 1).  If Xaml wants to hold the pointer via its own AddRef/Release pair, that's perfectly fine.  This code will ensure that no user code will have a constructed component returned, so Xaml will be the sole owner of the (useless) object.

strings/base_implements.h
sylveon:
```cpp
template<typename T, typename... Args>
T* create_and_initialize(Args&&... args)
{
    com_ptr<T> instance;
    instance.attach(new heap_implements<T>(std::forward<Args>(args)...);

    if constexpr (has_initializer<T>::value)
    {
        instance->InitializeComponent();
    }

    return instance.release();
}
```

vsix/ProjectTemplates/VC/Windows Universal/BlankApp/MainPage.cpp
Scottj1s: My concern with adding the ctor back is that it might encourage users who were previously relying on it to then call a property accessor, like myButton().  That would return null without a preceding call to InitializeComponent, which we don't want to encourage users to add to the ctor.

sylveon: If the ctor is blank with a comment mentionning the change and  how to override InitializeComponent, I don't think it would be an issue (but then the comment would need to be localized). Users can then delete it if desired.

Scottj1s: Added comments and a link to documentation, per feedback from all.  Long-term, I'll also get InitializeComponent covered on docs.microsoft.com.

@jlaanstra, Unfortunately, there's no good way to determine whether a newer VSIX item (from template) is paired with an older version of C++/WinRT nuget without false positives/negatives.  Ideally, we'd warn in both scenarios - when InitializeComponent was *not* being called from the constructor and should be, and when it is and should not be.  A developer has two remedies - update the CppWinRT nuget, or add InitializeComponent to the constructor.  We'd prefer the former, but they could do the latter. And if they didn't want to be nagged with a false-positive, we'd need to add a #define to ignore?  Tabling this for now.

Scottj1s:
I've opened a related improvement request for the Xaml compiler, to diagnose property accessor calls without InitializeComponent having first been called.

https://microsoft.visualstudio.com/OS/_workitems/edit/38777131

Scottj1s:
thanks all!

sylveon:
Trying to adopt this in my code - but now I'm hitting a little issue. I do some stuff in a constructor that both manipulates XAML, and accesses constructor params. This can't work if I want to move the XAML manipulation stuff in an overriden InitializeComponent because I don't get passed constructor parameters there. Is the recommended solution in this case to just store the constructor parameters needed in fields? Doesn't that waste memory, especially if passing big objects/strings and then forgetting to release them at the end of the custom InitializeComponent's body?

Maybe adding a way to also pass constructor parameters to InitializeComponent would be nice. Something like checking if `is_invocable<&D::InitializeComponent, D*, Args...>` is true, if yes pass constructor params if not call without params.

PR #1136, opened 2022-04-20
Use `auto` trick to catch missing header files for auto_revoke

oldnewthing:
We forgot to apply the `auto` trick to `auto_revoke` methods to catch missing header files at compile time. As a result, people who used `auto_revoke` didn't learn about missing header files until link time.

Apply the same technique to auto-revoke event handler registration.

(Story time: A customer was running into linker errors due to a missing header file when they used `auto_revoke`. One thing they tried was switching to non-`auto_revoke` events, and then they got the compile-time error. But their conclusion wasn't "Oh, that's my problem" but rather "Oh no, that made it worse, go back!")

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1001,14 +1001,13 @@ namespace cppwinrt
         if (is_add_overload(method))
         {
             auto format = R"(        using %_revoker = impl::event_revoker<%, &impl::abi_t<%>::remove_%>;
-        [[nodiscard]] WINRT_IMPL_AUTO(%_revoker) %(auto_revoke_t, %) const;
+        [[nodiscard]] auto %(auto_revoke_t, %) const;
 )";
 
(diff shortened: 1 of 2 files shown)


kennykerr:
FYI: https://github.com/microsoft/cppwinrt/pull/1117

oldnewthing:
Thanks for the heads-up. Will rework and return.

kennykerr (approved):
Much simpler. 😄

Scottj1s:
Introduced a regression with static event subscriptions.  Repro:
[RuntimeComponentStaticEvent/SubscribeToStaticEvent.cpp at master · Scottj1s/RuntimeComponentStaticEvent (github.com)](https://github.com/Scottj1s/RuntimeComponentStaticEvent/blob/master/RuntimeComponentStaticEvent/SubscribeToStaticEvent.cpp)

Issue #1137, opened 2022-04-22
Propose make more threadsafe assignment operators


afrankeser:
Problem:
One thread is reading the IUnknown while the other thread is assigning a new pointer. For the reading thread it is possible to get a nullptr instead the old/new pointer.

The writing thread called release_ref (line 166) and enters line 167, now m_ptr is nullptr.
Then the reading thread calls the as() function and retrieves the value of m_ptr (now nullptr).
The writing thread assigns m_ptr to the new value (line 167).
_base_windows.h_
```
162  IUnknown& operator=(IUnknown const& other) noexcept
163  {
164     if (this != &other)
165     {
166         release_ref();
167         m_ptr = other.m_ptr;
168         add_ref();
169    }
170 
171    return*this;
172 }
```
Same problem in the _com_ptr::copy_ref (base_com_ptr.h)_ implementation.

Here I propose a more threadsafe implementation, which the STL _shared_ptr_ also uses.
```
IUnknown& operator=(IUnknown const& other) noexcept
{
    IUnknown tmp(other);
    swap(tmp, *this);
    
    return*this;
}
```

Jaiganeshkumaran:
If you just swap, how will you call add_ref?

tim-weis:
> If you just swap, how will you call add_ref?

@Jaiganeshkumaran Doesn't `IUnknown tmp(other);` do that for us? Which, if it did, would then still be [thread-compatible](https://youtu.be/s5PCh_FaMfM?t=1216) (I'm guessing...). Unless I'm missing something obvious. Or subtle.

---

Though, I did learn a different trick, a while back, from Arthur O'Dwyer's CppCon 2019 talk "RAII and the Rule of Zero": [By-value assignment operator](https://youtu.be/7Qgd9B1KuMQ?t=2145).

The proposition he makes is to conflate copy- and move-assignment into a single by-value assignment operator:

```cpp
IUnknown& operator=(IUnknown rhs) noexcept
{
    rhs.swap(*this);
    return *this;
}
```

The audio goes on to explain:

> Arthur O'Dwyer: "In C++, when I take by value, rather than by reference, that makes a copy. Or maybe it makes a move. Right? That's up to the caller."

I'm guessing this is in reference to guaranteed [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision), mandated as of C++17. I will not pretend to have understood even a fraction of the referenced document, and most certainly not even a fraction of a fractional of any of its ramifications.

[...]

Issue #1123, opened 2022-03-22
C++20 and beyond (cppwinrt v3)


kennykerr:
cppwinrt v2 is based on C++17, which lacks a number of new features that may be of some benefit to the cppwinrt library. I'm starting to experiment with cppwinrt v3 and what it would look like to adopt some of the C++ language and library features introduced in C++20 and C++23. cppwinrt v2 already supports C++20 coroutines but some other features are harder to introduce without more substantial and breaking changes. Here's a starting list of what I'd like to see in v3. Feel free to comment on other features that you think might be beneficial.

C++20
* Three-way comparison - to simplify comparison generation
* Modules - to simplify and reduce build time
* Concepts - to simplify implementations
* Ranges - to simplify collections
* `span` - to replace `array_view`

C++23
* Deducing this - to simplify and reduce CRTP complexity & support the builder pattern

It should be clear that the theme is *make things simpler* and not just adding more bells and whistles. If we find that adding support for X makes things more complicated, then we'd be far less likely to adopt it. By themselves, the C++20 features aren't really compelling enough to spin up a v3 of cppwinrt, but C++23's "deducing this" feature could potentially move the needle so I'm thinking of target C++23 as the min version.

sylveon:
The work in the modules PR I opened a while ago could be used, but I'd need to check if the compiler bugs I hit have been fixed.

However, there is still stuff that could be done to improve modules here. For example a single change in any API rebuilds the entire module which takes a solid 2-3 minutes on a fast machine. We should probably use module fragments (for example, one per top-level namespace) so that a change to APIs in `Contoso.*` doesn't cause the compiler to rebuild all of `Windows.*`.

Also, it obviously needs support for producing runtime components, as right now the modules code in cppwinrt only supports producing.

Another question is if modules should be something that is always on. While I don't oppose that, some people might because it puts quite a migration burden on existing code.

I believe deducing this might help quite a lot with build times as well, so it may be worth making C++23 the min version.

asklar:
I believe the last straggler compiler bug impacting modules was fixed recently, @Scottj1s will have more data on that.
_Deducing this_ will make builder pattern a lot easier to implement/consume.

Scottj1s:
@sylveon I've been building on your topic branch, working around a few module compilation issues.  I've been working with the modules devs and the last blocker has now been addressed:
[...]

PR #1139, opened 2022-04-26
Copy and swap assignment

afrankeser:
Use the copy and swap idiom for com_ptr and IUnknown assignment operators provides several benefits
* no temporary nullptr during assignment
* prevents the erroneous destruction of recursive data structures (better explained here https://youtu.be/7Qgd9B1KuMQ?t=850)
* better exception safety guaranty (when exception was thrown then no changes were made), probably not so important here

### DIFF
--- a/strings/base_com_ptr.h
+++ b/strings/base_com_ptr.h
@@ -59,33 +59,31 @@ WINRT_EXPORT namespace winrt
 
         com_ptr& operator=(com_ptr const& other) noexcept
         {
-            copy_ref(other.m_ptr);
+            com_ptr cpy(other);
+            swap(*this, cpy);
             return*this;
         }
 
         com_ptr& operator=(com_ptr&& other) noexcept
         {
-            if (this != &other)
-            {
-                release_ref();
-                m_ptr = std::exchange(other.m_ptr, {});
-            }
-
+            com_ptr cpy(std::move(other));
+            swap(*this, cpy);
             return*this;
         }
 
         template <typename U>
         com_ptr& operator=(com_ptr<U> const& other) noexcept
         {
-            copy_ref(other.m_ptr);
+            com_ptr cpy(other);
+            swap(*this, cpy);
             return*this;
         }
 
         template <typename U>
         com_ptr& operator=(com_ptr<U>&& other) noexcept
         {
-            release_ref();
-            m_ptr = std::exchange(other.m_ptr, {});
+            com_ptr cpy(std::move(other));
+            swap(*this, cpy);
             return*this;
         }
 
@@ -177,7 +175,10 @@ WINRT_EXPORT namespace winrt
 
         void copy_from(type* other) noexcept
         {
-            copy_ref(other);
+            if (other)
+                const_cast<std::remove_const_t<type>*>(other)->AddRef();
+            com_ptr cpy(other, take_ownership_from_abi);
+            swap(*this, cpy);
         }
 
         void copy_to(type** other) const noexcept
@@ -199,16 +200,6 @@ WINRT_EXPORT namespace winrt
         }
 
     private:
-
-        void copy_ref(type* other) noexcept
-        {
-            if (m_ptr != other)
-            {
-                release_ref();
-                m_ptr = other;
-                add_ref();
-            }
-        }
 
         void add_ref() const noexcept
         {
--- a/strings/base_com_ptr.h
+++ b/strings/base_com_ptr.h
@@ -176,7 +176,9 @@ WINRT_EXPORT namespace winrt
         void copy_from(type* other) noexcept
         {
             if (other)
+			{
                 const_cast<std::remove_const_t<type>*>(other)->AddRef();
+			}
             com_ptr cpy(other, take_ownership_from_abi);
             swap(*this, cpy);
         }
--- a/strings/base_com_ptr.h
+++ b/strings/base_com_ptr.h
@@ -176,9 +176,9 @@ WINRT_EXPORT namespace winrt
         void copy_from(type* other) noexcept
         {
             if (other)
-			{
+            {
                 const_cast<std::remove_const_t<type>*>(other)->AddRef();
-			}
+            }
             com_ptr cpy(other, take_ownership_from_abi);
             swap(*this, cpy);
         }
--- a/strings/base_com_ptr.h
+++ b/strings/base_com_ptr.h
@@ -66,8 +66,11 @@ WINRT_EXPORT namespace winrt
 
         com_ptr& operator=(com_ptr&& other) noexcept
         {
-            com_ptr cpy(std::move(other));
-            swap(*this, cpy);
+            if (this != &other)
+            {
+                com_ptr cpy(std::move(other));
+                swap(*this, cpy);
+            }
             return*this;
         }
 
--- a/strings/base_windows.h
+++ b/strings/base_windows.h
@@ -168,8 +168,11 @@ WINRT_EXPORT namespace winrt::Windows::Foundation
 
         IUnknown& operator=(IUnknown&& other) noexcept
         {
-            IUnknown cpy(std::move(other));
-            swap(*this, cpy);
+            if (this != &other)
+            {
+                IUnknown cpy(std::move(other));
+                swap(*this, cpy);
+            }
             return*this;
         }
 
--- a/strings/base_windows.h
+++ b/strings/base_windows.h
@@ -161,9 +161,9 @@ WINRT_EXPORT namespace winrt::Windows::Foundation
 
         IUnknown& operator=(IUnknown const& other) noexcept
         {
-			IUnknown cpy(other);
-			swap(*this, cpy);
-			return*this;
+            IUnknown cpy(other);
+            swap(*this, cpy);
+            return*this;
         }
 
         IUnknown& operator=(IUnknown&& other) noexcept
--- a/strings/base_windows.h
+++ b/strings/base_windows.h
@@ -161,24 +161,15 @@ WINRT_EXPORT namespace winrt::Windows::Foundation
 
         IUnknown& operator=(IUnknown const& other) noexcept
         {
-            if (this != &other)
-            {
-                release_ref();
-                m_ptr = other.m_ptr;
-                add_ref();
-            }
-
-            return*this;
+			IUnknown cpy(other);
+			swap(*this, cpy);
+			return*this;
         }
 
         IUnknown& operator=(IUnknown&& other) noexcept
         {
-            if (this != &other)
-            {
-                release_ref();
-                m_ptr = std::exchange(other.m_ptr, {});
-            }
-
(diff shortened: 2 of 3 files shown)


strings/base_com_ptr.h
```diff
@@ -177,7 +175,10 @@ WINRT_EXPORT namespace winrt
 
         void copy_from(type* other) noexcept
         {
-            copy_ref(other);
+            if (other)
+                const_cast<std::remove_const_t<type>*>(other)->AddRef();
```
kennykerr, line 179:
```suggestion
            if (other)
            {
                const_cast<std::remove_const_t<type>*>(other)->AddRef();
            }
```

strings/base_windows.h
```diff
@@ -161,24 +161,15 @@ WINRT_EXPORT namespace winrt::Windows::Foundation
...
-            {
-                release_ref();
-                m_ptr = other.m_ptr;
-                add_ref();
-            }
-
-            return*this;
+			IUnknown cpy(other);
+			swap(*this, cpy);
+			return*this;
```
kennykerr, line 166: 4 space tabs please

strings/base_com_ptr.h
```diff
@@ -59,33 +59,31 @@ WINRT_EXPORT namespace winrt
 
         com_ptr& operator=(com_ptr const& other) noexcept
         {
-            copy_ref(other.m_ptr);
+            com_ptr cpy(other);
+            swap(*this, cpy);
             return*this;
         }
 
         com_ptr& operator=(com_ptr&& other) noexcept
```
sylveon, line 67: Take by value to fold both move and copy into a single operator?

strings/base_com_ptr.h
```diff
@@ -59,33 +59,31 @@ WINRT_EXPORT namespace winrt
...
         {
-            if (this != &other)
-            {
-                release_ref();
-                m_ptr = std::exchange(other.m_ptr, {});
-            }
-
+            com_ptr cpy(std::move(other));
```
oldnewthing, line 69: This means that `x = std::move(x)` temporarily leaves x null, which could be surprising. While it's unlikely that somebody will write exactly that, they may end up doing so through a series of wrappers, like a `std::tuple::get` method that returns an rvalue reference.

strings/base_com_ptr.h
afrankeser: I thought about it but decided against it because it is easier for me to understand. If I have only one assignment operator, the reader will surely ask where the move operator is or whether this was perhaps forgotten.
Furthermore, it has been done in the implementation of the STL shared_ptr (Microsoft, libstd++, ...) as well and I didn't want to put myself in danger of using another implementation in such a critical place.

strings/base_com_ptr.h
afrankeser: You are right. Do you have a suggestion for a good solution?
The only thing that comes to my mind is to check again for self-assignment.

strings/base_com_ptr.h
sylveon: A small comment can avoid the concerns about thinking that move was forgotten

afrankeser: I will change it if another one is for it.

strings/base_com_ptr.h
roxk: Take by value and you don't need to check for self-assignment. This is one of the motivation for the copy and swap idiom :D

MikeGitb: @roxk: Semantically that shouldn't change anything (now the user code would have to check for self assignment if it wants to avoid the transitional null state).
@oldnewthing :  What exactly would be the problem with x bein temporarily null? How would external code observe this?

MikeGitb: > Yes, this is technically UB, but it nevertheless happens, and the "temporarily null" behavior is unexpected.

This might be a generational thing (I only started to work with c++ in ernest around the time c++11 gotfinalized), but to be honest: I don't expect anything sane to happen if I violate the c++ language rules with regards to multithreading. 

Anyway: If never null even under violation of c++ multithreading rules is a desired property, then making this a by-value-parameter would definetly be out of the question, as you then would do a unconditional move. You have to check for self assignment instead as suggested by @afrankeser.


Btw.: Your example is using copy assignement, not move assignment.

afrankeser: > Take by value and you don't need to check for self-assignment. This is one of the motivation for the copy and swap idiom :D
Yes I know but one of the motivation of the PR is to avoid temporary nullptr during assignment. I will give it a try and readd the check.

MikeGitb: I don't see any problems with the PR as is (but that isn't meant to be a proper review)

strings/base_com_ptr.h
afrankeser: Sorry, the situation has changed so move and copy are different now. @roxk exception safety should not be a problem here, because the functions are not allowed to throw (noexcept) but the current implementation fulfills that as well.

tim-weis:
This being a pretty fundamental change to a core library implementation I would certainly hope to see some more rigorous risk/benefit analysis. As far as the benefits go, I don't see anything that would make a meaningful change for the better:

> * no temporary nullptr during assignment

That seems to be an attempt at making things less surprising for folks that write and cannot change defective code. If you argue that the transient `nullptr` value were observable (which it only ever is to buggy code), then this PR is introducing a change in behavior.

Now the real solution here is: Don't write buggy code. It's not like dealing with concurrency were bleeding edge technology. It's been well established for more than a decade now that value-semantics are a powerful tool in this context (incidentally, that's what `com_ptr` implements, and what *cppwinrt.exe* generates for coroutine signatures already).

> * prevents the erroneous destruction of recursive data structures (better explained here https://youtu.be/7Qgd9B1KuMQ?t=850)

That seems to be great, in general. When it comes to COM, though, this is trying to solve a problem that cannot happen. For assignment operators the reference count for COM objects cannot even temporarily drop to zero (with the exception of self-assignment, which is implemented as a no-op anyway). Am I missing something here?

> * better exception safety guaranty (when exception was thrown then no changes were made), probably not so important here

The operators were marked `noexcept` all along. Any potential exceptions raised would have caused the process to terminate. If clients run into issues here then that would make for a far better motivation to push a change like this through.

kennykerr:
Agree with @tim-weis - I'm not inclined to accept this PR as it has raised stability concerns that outweigh the nominal benefits of modernizing the implementation. The existing implementation has stood the test of time.

afrankeser:
I agree, this is a fundamental change but a also a fundamental implementation was used. The copy-swap idiom has already been studied for benefits and risks, so I can't understand what stability concerns there are about it.

> That seems to be great, in general. When it comes to COM, though, this is trying to solve a problem that cannot happen. For assignment operators the reference count for COM objects cannot even temporarily drop to zero (with the exception of self-assignment, which is implemented as a no-op anyway). Am I missing something here?

That is not true, The added tests fails (`root = root->m_child;`) when I use the old implementation. The old implementation releases first and then increments the reference counter. This leads to the entire list being released first (root and children are destroyed) and then the child is set as the new root, which is no longer valid. Recursive data structures are not freely invented constructs (e.g. a list).

We ran into the same problem @oldnewthing mentioned and that was also my main motivation to make this PR (with the recursive structures I realized later). You can argue that bugs are revealed when there are threading issues and it's ok to surprise the developer by temporarily setting nullptr. This would no longer be the case with the new implementation. But shouldn't such problems be detected by the code (analysis) and not by the framework?

But in the end it is your decision and I will accept it.

kennykerr:
If there is indeed a bug let's open a new issue for that. The original issue preceding this PR (#1137) seems to be mostly about modernizing the implementation and talks about thread-safety goals that are out of scope for C++/WinRT. 

I take it the "bug" in question seems to come down to this example from the PR:

```C++
// assign child to root should only destroy root not its children
root = root->m_child;
```

It's not obvious to me that the assignment of the RHS should happen before the destruction of the LHS, but it's a good question.  It also sounds like there's a related issue buried in the comments. I'm open to exploring that in an issue and if there's general consensus that this is indeed a bug, we can consider a PR that focuses on this rather than as a side effect of other improvements. But even if it is a bug, we have to consider whether a breaking change here is tolerable, or whether it should wait for a future cppwinrt v3 (#1123).

Thanks again for your patience and interest in contributing - very much appreciated.

tim-weis:
In lieu of a new, dedicated issue I'm leaving a few thoughts here, for future reference.

> Am I missing something here?

Turns out I was! Apologies for not following the test case through as thoroughly as I could have. Unarguably, it does showcase a bug. What's not nearly as clear is, *where* that bug is.

---

It's not entirely unreasonable to say that `com_ptr`'s copy-assignment operator implementation is at fault, and supplanting it with an implementation based on the copy-and-swap idiom solves the immediate issue. That makes for a justifiable case in support of improving (and *changing*) the current implementation.

---

It's equally reasonable to call out the test case's `Nested` implementation: It holds a class member that's managing resources (`m_child`), yet keeps it out in the `public:`, inviting clients to store references to it with unbounded lifetime, happily supporting them in exploring the vast space that is UB.

Indeed, that's what `root = root->child;` does. Once overload resolution has settled and agreed upon the best candidate, the right-hand side of the expression constructs a reference on `Nested::m_child` that doesn't live to see the end of the expression statement. The solution to *this* problem would be to keep the resource-owning `m_child` private, and only hand out references to it in tandem with the liabilities associated of (shared) ownership, e.g. a `com_ptr<T>` value:

```cpp
struct Nested : implements<Nested, Windows::Foundation::IInspectable>
{
    // ...
    auto child() const noexcept { return m_child; }
private:
    // ...
    com_ptr<Nested> m_child;
};
```

With that change, the test case no longer fails when executing `root = root->child();`.

---

Speaking of failure, irrespective of where one might assume the defect, this one is particularly nasty: In a debug configuration, exercising the [debug heap](https://docs.microsoft.com/en-us/visualstudio/debugger/crt-debug-heap-details), the pointer previously held by `m_child` gets properly filled with a `0xdd ... dddd` pattern, failing the following pointer deref and `add_ref()` call. In a release configuration things are far less delectable: It is a straight up heap corruption, or an arbitrary memory write vulnerability. The classic dangling pointer, in preparation of a use-after-free bug.

---

The current situation isn't ideal. Regardless of whether one classifies `com_ptr`'s (and `IUnknown`'s) assignment operators as a bug or just your average C++ footgun, as a client I welcome any attempt at hardening the library implementation against misuse. Application of the copy-and-swap idiom is undoubtedly a step towards that goal.

Still, a change to two of the most foundational types in C++/WinRT absolutely necessitates a thorough risk assessment. It is not sufficient to look at an alternative implementation in isolation. Instead, it's (also) required to evaluate the effects when mixing two implementations with two distinct sets of rules, under different deployment scenarios (source vs. object code). I don't feel confident in providing that, but hopefully someone else can more easily see through that complexity.

PR #1146, opened 2022-05-10
Test whether SDKManifest.xml exists before attempting to read it.

### LINKED ISSUE #1145: cppwinrt.exe -input <version>+ fails when WindowsIoT extension SDK is installed
It turns out that not all Extension SDKs contain a \<version\>/SDKManifest.xml file, but cppwinrt.exe assumes that they do. The result is that cppwinrt will fail with the following error when run with `-input <version>+` on a system with such an SDK installed (e.g. WindowsIoT), such as https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md:

```txt
cppwinrt : error Could not read the Windows SDK's Platform.xml
```

The extension SDK directories are enumerated here, and indiscriminately passed to `add_files_from_xml()` which fails in `SHCreateStreamOnFileW()` with `ERROR_PATH_NOT_FOUND`:
https://github.com/microsoft/cppwinrt/blob/464e59198c2046125d8ee5a8cc6ff1739097345f/cppwinrt/cmd_reader.h#L472

dfields-msft:
This change fixes #1145 by testing whether an `SDKManifest.xml` exists within an Extension SDK's folder before attempting to read it, and ignores Extension SDKs which lack this file.

### DIFF
--- a/cppwinrt/cmd_reader.h
+++ b/cppwinrt/cmd_reader.h
@@ -474,7 +474,11 @@ namespace cppwinrt
                         xml_path = item.path() / sdk_version;
                         xml_path /= L"SDKManifest.xml";
 
-                        add_files_from_xml(files, sdk_version, xml_path, sdk_path);
+                        // Not all Extension SDKs include an SDKManifest.xml file; ignore those which do not (e.g. WindowsIoT).
+                        if (std::filesystem::is_regular_file(xml_path))
+                        {
+                            add_files_from_xml(files, sdk_version, xml_path, sdk_path);
+                        }
                     }
 
                     continue;
--- a/cppwinrt/cmd_reader.h
+++ b/cppwinrt/cmd_reader.h
@@ -71,17 +71,31 @@ namespace cppwinrt
         }
     }
 
+    enum class xml_requirement
+    {
+        required = 0,
+        optional
+    };
+
     inline void add_files_from_xml(
         std::set<std::string>& files,
         std::string const& sdk_version,
         std::filesystem::path const& xml_path,
-        std::filesystem::path const& sdk_path)
+        std::filesystem::path const& sdk_path,
+        xml_requirement xml_path_requirement)
     {
         com_ptr<IStream> stream;
 
-        check_xml(SHCreateStreamOnFileW(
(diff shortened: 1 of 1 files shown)


cppwinrt/cmd_reader.h
```diff
@@ -474,7 +474,11 @@ namespace cppwinrt
                         xml_path = item.path() / sdk_version;
                         xml_path /= L"SDKManifest.xml";
 
-                        add_files_from_xml(files, sdk_version, xml_path, sdk_path);
+                        // Not all Extension SDKs include an SDKManifest.xml file; ignore those which do not (e.g. WindowsIoT).
+                        if (std::filesystem::is_regular_file(xml_path))
+                        {
+                            add_files_from_xml(files, sdk_version, xml_path, sdk_path);
```
ChrisGuzak, line 480: I assume an exception happens when this is missing. perhaps handling that is better.

dfields-msft: There are only a handful of `catch` statements in this codebase, and all of them either re-throw, or print an error message before terminating, so I'm hesitant to introduce control flow logic based on exception handling here.

I did consider testing the return value of `SHCreateStreamOnFileW` against `HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND)` and returning early from `add_files_from_xml()` rather than throwing, but I felt it was clearer to have this logic at the call site where the paths are built.

sylveon: Checking first seems like a pretty obvious TOCTOU bug

dfields-msft: Fair point in general, and I'm happy to switch to reacting to failure instead if that's preferred; however in this case such a bug would require someone to be uninstalling the Windows SDK *while* running cppwinrt.exe, which seems quite impractical/unlikely.

dfields-msft: I like that idea (and the `ignore_if_missing` parameter is important since this function is also used for `Platform.xml` which *should* always be present). Update incoming...

kennykerr (approved):

PR #1149, opened 2022-05-17
Add support for composed implementations of IComponentConnector

### LINKED ISSUE #1140: InitializeComponent is not called for base classes.
With #1130, `InitializeComponent` is now called internally as part of `winrt::make` instead of within the constructor. This means that the following scenario, by default, does not work anymore:

- Class A is a base class using XAML.
- Class B derives from class A in metadata, and uses XAML too.
- Both are within the same project, so `implementation::B` directly derives from `implementation::A`, instead of using WinRT composition.

C++/WinRT will call class B's `InitializeComponent` function, but the XAML compiler did not generate code in class B's `InitializeComponent` function to call class A's `InitializeComponent` function. As a result, nothing from class A is loaded: none of the XAML is visible, no event handlers are attached, `x:Bind` bindings are not set up, trying to access a `x:Name`'d control gives `nullptr`, ets.

To fix this, a user has to explicitly go and add the following in `implementation::B`:
```cpp
void B::InitializeComponent()
{
    A::InitializeComponent();
    BT::InitializeComponent();
}
```

These kinds of things should be transparent to the user.

Scottj1s:
fixes #1140

This change adds a helper, ComponentConnectorT, to resolve the breaking change introduced with #1140.  Note that this is manual opt-in for the developer, and not an automatic fix, as that would require coordinated work with the Xaml Compiler.  I don't think it's worth the effort to do that, for several reasons:

When deriving from a base class with markup, IComponentConnector::Connect has never been reliable outside construction, such as when called back from FindName.  If a base class has a property with both the same type (e.g., TextBlock) and connectionId as a derived class,  then Connect would erroneously call the derived property setter.   The same risk applies to DisconnectUnloadedObject.  This was the opposite of the non-virtual dispatch that InitializeComponent was previously relying on.  At steady state (outside construction), virtual dispatch would be functional and the derived Connect would always be called.  However, unlike with calls from within InitializeComponent, where _dispatch_base can be used to correctly dispatch callbacks, there is no way to know how to do this with Xaml runtime calls for FindName (x:Load and template apply scenarios).  The risk exists in both directions - a base property can erroneously override a derived property and vice versa.

For that reason, the Xaml team does not recommend bases with markup, although there is nothing to prevent that or inform the developer of the risks (see [related discussion](https://github.com/microsoft/microsoft-ui-xaml/issues/5597)).  To preserve the existing (risky) behavior outside construction, _dispatch_base defaults to false to dispatch to the derived class.

### DIFF
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -797,6 +797,7 @@ catch (...) { return winrt::to_hresult(); }
                 }
                 else
                 {
+                    composable_base_name = w.write_temp("using composable_base = B;");
                     base_type_parameter = ", typename B";
                     base_type_argument = ", B";
                     no_module_lock = "no_module_lock, ";
@@ -868,7 +869,7 @@ catch (...) { return winrt::to_hresult(); }
 namespace winrt::@::implementation
 {
     template <typename D, typename... I>
-    using %T = %_base<D, I...>;
+    using %T = %_base<D, impl::marker, I...>;
 }
 
 #endif
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -869,7 +869,7 @@ catch (...) { return winrt::to_hresult(); }
 namespace winrt::@::implementation
 {
     template <typename D, typename... I>
-    using %T = %_base<D, impl::marker, I...>;
+    using %T = %_base<D, I...>;
 }
 
 #endif
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -254,7 +254,7 @@ namespace winrt::impl
     struct interface_list<>
     {
         template <typename Traits>
-        static constexpr auto find(const Traits& traits) noexcept
+        static constexpr auto find(Traits const& traits) noexcept
         {
             return traits.not_found();
         }
@@ -264,7 +264,7 @@ namespace winrt::impl
     struct interface_list<First, Rest...>
     {
         template <typename Traits>
-        static constexpr auto find(const Traits& traits) noexcept
+        static constexpr auto find(Traits const& traits) noexcept
         {
             if (traits.template test<First>())
             {
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -365,7 +365,7 @@ namespace winrt::impl
     template<typename T>
     struct find_iid_traits
     {
-        const T* _obj;
+        const T* m_object;
         const guid& m_guid;
 
         template <typename I>
@@ -377,7 +377,7 @@ namespace winrt::impl
         template <typename I>
         constexpr void* found() const noexcept
         {
-            return to_abi<I>(_obj);
+            return to_abi<I>(m_object);
         }
 
         static constexpr void* not_found() noexcept
@@ -422,7 +422,7 @@ namespace winrt::impl
     template<typename T>
     struct find_inspectable_traits
     {
-        const T* _obj;
+        const T* m_object;
 
         template <typename I>
         static constexpr bool test() noexcept
@@ -433,7 +433,7 @@ namespace winrt::impl
         template <typename I>
         constexpr void* found() const noexcept
         {
-            return to_abi<I>(_obj);
+            return to_abi<I>(m_object);
         }
 
         static constexpr void* not_found() noexcept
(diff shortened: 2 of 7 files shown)


cppwinrt/component_writers.h
```diff
@@ -868,7 +869,7 @@ catch (...) { return winrt::to_hresult(); }
 namespace winrt::@::implementation
 {
     template <typename D, typename... I>
-    using %T = %_base<D, I...>;
+    using %T = %_base<D, impl::marker, I...>;
```
Scottj1s, line 872: for classes without interfaces, this produced a compile error - not enough template parameters.  supplying the marker is a simple way to resolve the argument count.

sylveon:
Seems good to me - although there should probably be some unit testing for it. I suppose this works if we have more than 2 classes in the hierarchy with XAML markup? E.g. BaseClass -> DerivedClass -> FurtherDerivedClass. I don't actually do that but it's probably something that others might do.

Scottj1s:
> Seems good to me - although there should probably be some unit testing for it. I suppose this works if we have more than 2 classes in the hierarchy with XAML markup? E.g. BaseClass -> DerivedClass -> FurtherDerivedClass. I don't actually do that but it's probably something that others might do.

Yes, will work on adding unit tests when time permits.  And yes, I tested your scenario of 3 classes in the hierarchy to ensure it was fully general.

strings/base_xaml_component_connector.h
```diff
@@ -0,0 +1,53 @@
...
+            return D::GetBindingConnector(connectionId, target);
+        }
+
+    private:
+        static constexpr bool has_connectable_base{
+            impl::has_initializer<composable_base>::value &&
+            impl::has_interface<D, IComponentConnector>() &&
+            impl::has_interface<D, IComponentConnector2>() };
+
+        bool _dispatch_base{};
```
kennykerr, line 51:
```suggestion
        bool m_dispatch_base{};
```

Consistency.

kennykerr (approved):
Looks reasonable. I don't claim to understand all the Xaml fanciness going on here, perhaps @oldnewthing can take a peek.

strings/base_implements.h
```diff
@@ -362,30 +362,84 @@ namespace winrt::impl
...
+    constexpr bool has_interface() noexcept
+    {
+        return impl::implemented_interfaces<T>::find(has_interface_traits<I>{});
     }
 
-    struct inspectable_finder
+    template<typename T>
+    struct find_inspectable_traits
     {
+        const T* _obj;
```
kennykerr, line 425:
```suggestion
        const T* m_object;
```

strings/base_implements.h
```diff
@@ -362,30 +362,84 @@ namespace winrt::impl
         using type = typename implements_default_interface<T>::type;
     };
 
-    struct iid_finder
+    template<typename T>
+    struct find_iid_traits
     {
+        const T* _obj;
```
kennykerr, line 368:
```suggestion
        T const* m_object;
```

strings/base_implements.h
```diff
@@ -253,24 +253,24 @@ namespace winrt::impl
...
         }
     };
 
     template <typename First, typename ... Rest>
     struct interface_list<First, Rest...>
     {
-        template <typename T, typename Predicate>
-        static constexpr void* find(const T* obj, const Predicate& pred) noexcept
+        template <typename Traits>
+        static constexpr auto find(const Traits& traits) noexcept
```
kennykerr, line 267:
```suggestion
        static constexpr auto find(Traits const& traits) noexcept
```

sylveon:
Yes, that appears to be it. It seems we'd need to either have a different string for Microsoft.UI.Xaml, or have the current one be able to detect if IComponentConnector2 actually exists in the current namespace. I think the former is better, since WinUI 3 is bound to diverge more over time while system XAML is essentially frozen at this point.

Ideally this is a thing that WinUI 3 would ship by itself, but the code makes use of some `impl` stuff, and I think creating a dependency on cppwinrt's implementation details within WinUI 3 to not be a great idea (especially since future endeavors like C++20 modules support will make the `impl` namespace entirely invisible to consumers).

@Scottj1s what approach is the best here?

PR #1158, opened 2022-06-08
Static events should not use the auto trick

oldnewthing:
The same way we don't use the auto trick for static properties and static methods.

Fixes issue introduced by #1136

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -2986,13 +2986,15 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
             return;
         }
 
+        auto is_opt_type = settings.component_opt && settings.component_filter.includes(type);
+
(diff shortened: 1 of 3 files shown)


cppwinrt/code_writers.h
```diff
@@ -2986,13 +2986,15 @@ struct __declspec(empty_bases) produce_dispatch_to_overridable<T, D, %>
             return;
         }
 
+        auto is_opt_type = settings.component_opt && settings.component_filter.includes(type);
```
Scottj1s, line 2989: is_component_type?

Scottj1s (approved):
thanks for the quick turnaround!

cppwinrt/code_writers.h
oldnewthing: It's "is optimized component type". non-optimized components still can use auto.

PR #1181, opened 2022-09-09
Add explicit casts to atomic_ref_count

timguen:
Static code analysis flagged these three instances of signed/unsigned mismatches. This change adds explicit static_cast to uint32_t to indicate that this is intentional.


kennykerr (approved):
Thanks!

PR #1183, opened 2022-09-15
Default initialize `winrt::guid` members

yashrajborse:
winrt::guid members are not default initialized which may lead to undefined behavior in the usage code if not explicitly initialized.
As per C++ core guidelines: Prevents “use before set” errors
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Rc-initialize

### DIFF
--- a/strings/base_types.h
+++ b/strings/base_types.h
@@ -133,10 +133,10 @@ WINRT_EXPORT namespace winrt
 
     public:
 
-        uint32_t Data1;
-        uint16_t Data2;
-        uint16_t Data3;
-        uint8_t  Data4[8];
+        uint32_t Data1{};
+        uint16_t Data2{};
+        uint16_t Data3{};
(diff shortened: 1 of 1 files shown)


jonwis (approved):

yashrajborse:
> Unfortunately, this collides with the low-level natvis debugger extension (it's used as part of a union) and also illustrates the compatibility risk of such a change. Lacking a more compelling reason for this change, we should probably leave it as is.

Can you share more details on this? Sorry I'm unaware about its usage and would like to see how it collides. Thanks.

kennykerr:
You can look at the internal build log here: https://dev.azure.com/microsoft/Dart/_build/results?buildId=56161598&view=results

kennykerr:
Or just build and test it locally as described in the readme.

PR #1185, opened 2022-09-15
C++20 consumers should have std::source_location support when logging error information

### LINKED ISSUE #1184: C++20 consumers should have std::source_location support when logging error information
`throw_hresult` has `winrt_throw_hresult_handler` to call out to logging libraries (such as WIL) when errors are encountered.  There is an incomplete note that source information (file, function name, line) are not hooked up.  C++20 adds `std::source_location` which will unblock filling in that information.

dmachaj:
Closes #1184

This change is being made to (greatly!) improve the error logging for exceptions that originate in cppwinrt.  This is especially directed at the cases where cppwinrt and WIL are interoperating.  The extension point between them has the ability to forward the file name, line number, and function name across to be logged.  They always pass null.

With C++20 and `std::source_location` specifically that can change.  It is possible to capture that information in a standards-compliant way and forward it along to be logged.  The result is __vastly__ improved WIL logging output.

Unfortunately this change is a bit ugly and requires macros.  The reason for that is because `std::source_location` is new in C++20 and because it must be constructed by the code that is calling into cppwinrt headers.  The support floor for cppwinrt is C++17 so this type needs to be behind an `#ifdef`.  That is what leads to the macros (which evaluate to nothing for C++17).

The second bit of ugliness is that the calling code must be what constructs the `std::source_location`.  Otherwise it will get the source information for cppwinrt itself which is not helpful.  This is achieved with a default parameter for every method that can originate an error.  The caller silently constructs it before the call so it has accurate source information.  For cases where errors hop a few times this must be forwarded so it is not lost.

Other miscellaneous changes include:
* The test_cpp20 project was updated to the VS2022 toolset.  The VS2019 toolset has an Internal Compiler Error (ICE) when building C++20 code.  The crash is gone in newer versions.
* The test_cpp20 project now requests C++20 (not C++/latest).
* `hresult_error::originate` will now call `winrt_throw_hesult_handler` if it is provided.  Previously `throw winrt::hresult_error(E_WHATEVER)` would not log to WIL.  Now it does.
* `#pragma detect mismatch` for whether `std::source_location` is being used, or not.

A new test file was added to the C++20 test binary.  This test ensures that the source information is accurate (and in fact caught a subtle bug when the test was being written!).

### DIFF
--- a/Directory.Build.Props
+++ b/Directory.Build.Props
@@ -32,6 +32,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <!--<CppWinRTBuildVersion Condition="'$(CppWinRTBuildVersion)'==''">210122.210122.210122.210122</CppWinRTBuildVersion>-->
     <CppWinRTBuildVersion Condition="'$(CppWinRTBuildVersion)'==''">2.3.4.5</CppWinRTBuildVersion>
     <CppWinRTPlatform>$(Platform)</CppWinRTPlatform>
     <CppWinRTPlatform Condition="'$(Platform)'=='Win32'">x86</CppWinRTPlatform>
@@ -46,6 +47,7 @@
       <SDLCheck>true</SDLCheck>
       <ConformanceMode>true</ConformanceMode>
       <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'==''">stdcpp17</LanguageStandard>
+      <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'=='stdcpp20'">stdcpp20</LanguageStandard>
       <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'=='latest'">stdcpplatest</LanguageStandard>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
--- a/Directory.Build.Props
+++ b/Directory.Build.Props
@@ -32,7 +32,6 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!--<CppWinRTBuildVersion Condition="'$(CppWinRTBuildVersion)'==''">210122.210122.210122.210122</CppWinRTBuildVersion>-->
     <CppWinRTBuildVersion Condition="'$(CppWinRTBuildVersion)'==''">2.3.4.5</CppWinRTBuildVersion>
     <CppWinRTPlatform>$(Platform)</CppWinRTPlatform>
     <CppWinRTPlatform Condition="'$(Platform)'=='Win32'">x86</CppWinRTPlatform>
--- a/strings/base_array.h
+++ b/strings/base_array.h
@@ -511,36 +511,24 @@ WINRT_EXPORT namespace winrt
         return impl::com_array_proxy<T>(__valueSize, value);
     }
 
-    inline hstring get_class_name(Windows::Foundation::IInspectable const& object)
+    inline hstring get_class_name(Windows::Foundation::IInspectable const& object WINRT_IMPL_SOURCE_LOCATION_ARGS)
     {
         void* value{};
-#ifdef __cpp_lib_source_location
-        check_hresult((*(impl::inspectable_abi**)&object)->GetRuntimeClassName(&value), std::source_location::current());
-#else
-        check_hresult((*(impl::inspectable_abi**)&object)->GetRuntimeClassName(&value));
-#endif
+        check_hresult((*(impl::inspectable_abi**)&object)->GetRuntimeClassName(&value) WINRT_IMPL_SOURCE_LOCATION_FORWARD);
         return { value, take_ownership_from_abi };
     }
 
-    inline com_array<guid> get_interfaces(Windows::Foundation::IInspectable const& object)
+    inline com_array<guid> get_interfaces(Windows::Foundation::IInspectable const& object WINRT_IMPL_SOURCE_LOCATION_ARGS)
     {
         com_array<guid> value;
-#ifdef __cpp_lib_source_location
-        check_hresult((*(impl::inspectable_abi**)&object)->GetIids(impl::put_size_abi(value), put_abi(value)), std::source_location::current());
-#else
-        check_hresult((*(impl::inspectable_abi**)&object)->GetIids(impl::put_size_abi(value), put_abi(value)));
-#endif
+        check_hresult((*(impl::inspectable_abi**)&object)->GetIids(impl::put_size_abi(value), put_abi(value)) WINRT_IMPL_SOURCE_LOCATION_FORWARD);
         return value;
     }
 
-    inline Windows::Foundation::TrustLevel get_trust_level(Windows::Foundation::IInspectable const& object)
+    inline Windows::Foundation::TrustLevel get_trust_level(Windows::Foundation::IInspectable const& object WINRT_IMPL_SOURCE_LOCATION_ARGS)
     {
         Windows::Foundation::TrustLevel value{};
-#ifdef __cpp_lib_source_location
-        check_hresult((*(impl::inspectable_abi**)&object)->GetTrustLevel(&value), std::source_location::current());
-#else
-        check_hresult((*(impl::inspectable_abi**)&object)->GetTrustLevel(&value));
-#endif
+        check_hresult((*(impl::inspectable_abi**)&object)->GetTrustLevel(&value) WINRT_IMPL_SOURCE_LOCATION_FORWARD);
         return value;
     }
 }
--- a/strings/base_array.h
+++ b/strings/base_array.h
@@ -514,21 +514,33 @@ WINRT_EXPORT namespace winrt
     inline hstring get_class_name(Windows::Foundation::IInspectable const& object)
     {
         void* value{};
+#ifdef __cpp_lib_source_location
+        check_hresult((*(impl::inspectable_abi**)&object)->GetRuntimeClassName(&value), std::source_location::current());
+#else
         check_hresult((*(impl::inspectable_abi**)&object)->GetRuntimeClassName(&value));
+#endif
         return { value, take_ownership_from_abi };
     }
 
     inline com_array<guid> get_interfaces(Windows::Foundation::IInspectable const& object)
     {
         com_array<guid> value;
+#ifdef __cpp_lib_source_location
+        check_hresult((*(impl::inspectable_abi**)&object)->GetIids(impl::put_size_abi(value), put_abi(value)), std::source_location::current());
+#else
         check_hresult((*(impl::inspectable_abi**)&object)->GetIids(impl::put_size_abi(value), put_abi(value)));
+#endif
         return value;
     }
 
     inline Windows::Foundation::TrustLevel get_trust_level(Windows::Foundation::IInspectable const& object)
     {
         Windows::Foundation::TrustLevel value{};
+#ifdef __cpp_lib_source_location
+        check_hresult((*(impl::inspectable_abi**)&object)->GetTrustLevel(&value), std::source_location::current());
+#else
         check_hresult((*(impl::inspectable_abi**)&object)->GetTrustLevel(&value));
+#endif
         return value;
     }
 }
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -315,6 +315,16 @@ WINRT_EXPORT namespace winrt
             impl::load_runtime_function(L"combase.dll", "RoOriginateLanguageException", handler, fallback_RoOriginateLanguageException);
             WINRT_VERIFY(handler(code, message, nullptr));
 
+            if (winrt_throw_hresult_handler)
+            {
+#ifdef __cpp_lib_source_location
+                std::source_location sourceInformation = std::source_location::current();
+                winrt_throw_hresult_handler(sourceInformation.line(), sourceInformation.file_name(), sourceInformation.function_name(), WINRT_IMPL_RETURNADDRESS(), code);
+#else
+                winrt_throw_hresult_handler(0, nullptr, nullptr, WINRT_IMPL_RETURNADDRESS(), code);
+#endif
+            }
+
             com_ptr<impl::IErrorInfo> info;
             WINRT_VERIFY_(0, WINRT_IMPL_GetErrorInfo(0, info.put_void()));
             WINRT_VERIFY(info.try_as(m_info));
@@ -442,7 +452,6 @@ WINRT_EXPORT namespace winrt
         if (winrt_throw_hresult_handler)
         {
 #ifdef __cpp_lib_source_location
-            // line, filename, func
             winrt_throw_hresult_handler(sourceInformation.line(), sourceInformation.file_name(), sourceInformation.function_name(), WINRT_IMPL_RETURNADDRESS(), result);
 #else
             winrt_throw_hresult_handler(0, nullptr, nullptr, WINRT_IMPL_RETURNADDRESS(), result);
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -433,11 +433,20 @@ WINRT_EXPORT namespace winrt
         hresult_canceled(take_ownership_from_abi_t) noexcept : hresult_error(impl::error_canceled, take_ownership_from_abi) {}
     };
 
+#ifdef __cpp_lib_source_location
+    [[noreturn]] inline WINRT_IMPL_NOINLINE void throw_hresult(hresult const result, std::source_location sourceInformation = std::source_location::current())
+#else
     [[noreturn]] inline WINRT_IMPL_NOINLINE void throw_hresult(hresult const result)
+#endif
     {
         if (winrt_throw_hresult_handler)
         {
+#ifdef __cpp_lib_source_location
+            // line, filename, func
+            winrt_throw_hresult_handler(sourceInformation.line(), sourceInformation.file_name(), sourceInformation.function_name(), WINRT_IMPL_RETURNADDRESS(), result);
+#else
             winrt_throw_hresult_handler(0, nullptr, nullptr, WINRT_IMPL_RETURNADDRESS(), result);
+#endif
         }
 
         if (result == impl::error_bad_alloc)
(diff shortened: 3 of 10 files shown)


dmachaj:
I have a sample program that combines cppwinrt and WIL to validate the logging output with and without these changes.

Lines 22-45 are as follows:
```cpp
void ThrowWinrtHresult()
{
    throw winrt::hresult_error(E_INVALIDARG);
}

void CheckWinrtHresult()
{
    winrt::check_hresult(E_INVALIDARG);
}

void CheckWinrtWin32Error()
{
    winrt::check_win32(ERROR_ACCESS_AUDIT_BY_POLICY);
}

void ThrowWILException()
{
    THROW_HR(E_INVALIDARG);
}

void CheckWILException()
{
    THROW_IF_FAILED(E_INVALIDARG);
}
```

Lines 86-124 are as follows:
```cpp
    try
    {
        OutputDebugString(L"----throw winrt::hresult_error()----\r\n");
        ThrowWinrtHresult();
    }
    CATCH_LOG();
    OutputDebugString(L"----\r\n\r\n");

    try
    {
        OutputDebugString(L"----winrt::check_hresult()----\r\n");
        CheckWinrtHresult();
    }
    CATCH_LOG();
    OutputDebugString(L"----\r\n\r\n");

    try
    {
        OutputDebugString(L"----winrt::check_win32()----\r\n");
        CheckWinrtWin32Error();
    }
    CATCH_LOG();
    OutputDebugString(L"----\r\n\r\n");

    try
    {
        OutputDebugString(L"----THROW_HR----\r\n");
        ThrowWILException();
    }
    CATCH_LOG();
    OutputDebugString(L"----\r\n\r\n");

    try
    {
        OutputDebugString(L"----THROW_IF_FAILED----\r\n");
        CheckWILException();
    }
    CATCH_LOG();
    OutputDebugString(L"----\r\n\r\n");
```

## C++17 output
```
----throw winrt::hresult_error()----
(null)!00007FF6A684415E: LogHr(1) tid(a1bc) 80070057 The parameter is incorrect.
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(91)\(null)!00007FF6A68E9E05: (caller: 00007FF6A68CDD4F) LogHr(2) tid(a1bc) 80070057 The parameter is incorrect.
    Msg:[winrt::hresult_error: The parameter is incorrect.] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----winrt::check_hresult()----
(null)!00007FF6A68539DE: LogHr(3) tid(a1bc) 80070057 The parameter is incorrect.
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(99)\(null)!00007FF6A68E9E75: (caller: 00007FF6A68CDD4F) LogHr(4) tid(a1bc) 80070057 The parameter is incorrect.
    Msg:[winrt::hresult_error: The parameter is incorrect.] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----winrt::check_win32()----
(null)!00007FF6A6839A25: LogHr(5) tid(a1bc) 80070311 Access to %1 is monitored by policy rule %2.
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(107)\(null)!00007FF6A68E9EE5: (caller: 00007FF6A68CDD4F) LogHr(6) tid(a1bc) 80070311 Access to %1 is monitored by policy rule %2.
    Msg:[winrt::hresult_error: Access to %1 is monitored by policy rule %2.] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----THROW_HR----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(39)\(null)!00007FF6A6851665: (caller: 00007FF6A684215D) Exception(1) tid(a1bc) 80070057 The parameter is incorrect.
    [ThrowWILException(E_INVALIDARG)]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(115)\(null)!00007FF6A68E9F55: (caller: 00007FF6A68CDD4F) LogHr(7) tid(a1bc) 80070057 The parameter is incorrect.
    Msg:[V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(39)\(null)!00007FF6A6851665: (caller: 00007FF6A684215D) Exception(1) tid(a1bc) 80070057 The parameter is incorrect.
    [ThrowWILException(E_INVALIDARG)]
] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----THROW_IF_FAILED----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(44)\(null)!00007FF6A685181C: (caller: 00007FF6A6842180) Exception(2) tid(a1bc) 80070057 The parameter is incorrect.
    [CheckWILException(E_INVALIDARG)]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(123)\(null)!00007FF6A68E9FC5: (caller: 00007FF6A68CDD4F) LogHr(8) tid(a1bc) 80070057 The parameter is incorrect.
    Msg:[V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(44)\(null)!00007FF6A685181C: (caller: 00007FF6A6842180) Exception(2) tid(a1bc) 80070057 The parameter is incorrect.
    [CheckWILException(E_INVALIDARG)]
] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----
```

## C++20 output
```
----throw winrt::hresult_error()----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(24)\(null)!00007FF6063DAAA3: LogHr(1) tid(ab54) 80070057 The parameter is incorrect.
    [ThrowWinrtHresult]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(91)\(null)!00007FF606488975: (caller: 00007FF60646BC2F) LogHr(2) tid(ab54) 80070057 The parameter is incorrect.
    Msg:[winrt::hresult_error: The parameter is incorrect.] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----winrt::check_hresult()----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(29)\(null)!00007FF6063EBA33: LogHr(3) tid(ab54) 80070057 The parameter is incorrect.
    [CheckWinrtHresult]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(99)\(null)!00007FF6064889E5: (caller: 00007FF60646BC2F) LogHr(4) tid(ab54) 80070057 The parameter is incorrect.
    Msg:[winrt::hresult_error: The parameter is incorrect.] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----winrt::check_win32()----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(34)\(null)!00007FF6063CFB91: LogHr(5) tid(ab54) 80070311 Access to %1 is monitored by policy rule %2.
    [CheckWinrtWin32Error]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(107)\(null)!00007FF606488A55: (caller: 00007FF60646BC2F) LogHr(6) tid(ab54) 80070311 Access to %1 is monitored by policy rule %2.
    Msg:[winrt::hresult_error: Access to %1 is monitored by policy rule %2.] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----THROW_HR----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(39)\(null)!00007FF6063E9315: (caller: 00007FF6063D884D) Exception(1) tid(ab54) 80070057 The parameter is incorrect.
    [ThrowWILException(E_INVALIDARG)]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(115)\(null)!00007FF606488AC5: (caller: 00007FF60646BC2F) LogHr(7) tid(ab54) 80070057 The parameter is incorrect.
    Msg:[V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(39)\(null)!00007FF6063E9315: (caller: 00007FF6063D884D) Exception(1) tid(ab54) 80070057 The parameter is incorrect.
    [ThrowWILException(E_INVALIDARG)]
] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----

----THROW_IF_FAILED----
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(44)\(null)!00007FF6063E956C: (caller: 00007FF6063D8870) Exception(2) tid(ab54) 80070057 The parameter is incorrect.
    [CheckWILException(E_INVALIDARG)]
V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(123)\(null)!00007FF606488B35: (caller: 00007FF60646BC2F) LogHr(8) tid(ab54) 80070057 The parameter is incorrect.
    Msg:[V:\T\davidmac.randomprojects\C++WinRTAndWILInteropTesting\C++WinRTAndWILInteropTesting\App.cpp(44)\(null)!00007FF6063E956C: (caller: 00007FF6063D8870) Exception(2) tid(ab54) 80070057 The parameter is incorrect.
    [CheckWILException(E_INVALIDARG)]
] [winrt::C__WinRTAndWILInteropTesting::implementation::App::App]
----
```

The code was the same in both cases.  The only difference was changing the C++ language standard and recompiling.  The notable difference is that previously there was no logging where the exception was thrown (e.g. line 29).  Now there is logging both when it is thrown and when it is caught/logged.  This will greatly improve the debugging experience.

test/test_cpp20/custom_error.cpp
```diff
@@ -0,0 +1,49 @@
...
+        // In C++20 these fields should be filled in by std::source_location
+        REQUIRE(lineNumber == 15);
+        const auto fileNameSv = std::string_view(fileName);
+        REQUIRE(!fileNameSv.empty());
+        REQUIRE(fileNameSv.find("custom_error.cpp") != std::string::npos);
+        const auto functionNameSv = std::string_view(functionName);
+        REQUIRE(!functionNameSv.empty());
+        REQUIRE(functionNameSv == "FailOnLine15");
+
+        lineNumber; fileName; functionName;
```
dmachaj, line 29: Whoops these are no longer needed to prevent unused parameters.  Will remove.

strings/base_macros.h
```diff
@@ -54,3 +54,27 @@
...
 #endif
+
+// std::source_location is a C++20 feature, which is above the C++17 feature floor for cppwinrt.  The source location needs
+// to be the calling code, not cppwinrt itself, so that it is useful to developers building on top of this library.  As a
+// result any public-facing method that can result in an error needs a default-constructed source_location argument.  Because
+// this type does not exist in C++17 we need to use a macro to optionally add parameters and forwarding wherever appropriate.
+#ifdef __cpp_lib_source_location
+#define WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT , std::source_location sourceInformation
+#define WINRT_IMPL_SOURCE_LOCATION_ARGS , std::source_location sourceInformation = std::source_location::current()
+#define WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM std::source_location sourceInformation = std::source_location::current()
```
oldnewthing, line 65:
```suggestion
#define WINRT_IMPL_SOURCE_LOCATION_ARGS , std::source_location const& sourceInformation = std::source_location::current()
#define WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM std::source_location const& sourceInformation = std::source_location::current()
```
Adding `const&` avoids copying when forwarding. The `source_location` is [rather substantial struct](https://github.com/microsoft/STL/blob/67a96a910fb53aa4bc12801fb9683d30b2647c16/stl/inc/source_location#L53).

dmachaj: Good call.  I should have done it that way originally.

dmachaj:
@kennykerr unfortunately the build is failing because the agents do not have the v143 toolset installed

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(439,5): error MSB8020: The build tools for v143 (Platform Toolset = 'v143') cannot be found. To build using the v143 build tools, please install v143 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [D:\a\1\s\test\test_cpp20\test_cpp20.vcxproj]
```

The C++20 support in v142 is incomplete and the compiler crashes when `std::source_information` is included (let alone used).

Directory.Build.Props
```diff
@@ -46,6 +46,7 @@
       <SDLCheck>true</SDLCheck>
       <ConformanceMode>true</ConformanceMode>
       <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'==''">stdcpp17</LanguageStandard>
```
sylveon, line 48: Also add a value for 17?

```suggestion
      <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'=='' Or '$(CppWinRTLanguageStandard)'=='17'">stdcpp17</LanguageStandard>
```

Directory.Build.Props
```diff
@@ -46,6 +46,7 @@
...
+      <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'=='stdcpp20'">stdcpp20</LanguageStandard>
```
sylveon, line 49: It seems inconsistent that we're using `latest` to mean the latest standard but `stdcpp20` to mean 20.

```suggestion
      <LanguageStandard Condition="'$(CppWinRTLanguageStandard)'=='20'">stdcpp20</LanguageStandard>
```

test/test_cpp20/test_cpp20.vcxproj
```diff
@@ -39,9 +39,12 @@
     <ProjectGuid>{5FF6CD6C-515A-4D55-97B6-62AD9BCB77EA}</ProjectGuid>
     <RootNamespace>unittests</RootNamespace>
     <ProjectName>test_cpp20</ProjectName>
-    <CppWinRTLanguageStandard>latest</CppWinRTLanguageStandard>
+    <CppWinRTLanguageStandard>stdcpp20</CppWinRTLanguageStandard>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Label="Configuration">
+    <PlatformToolset>v143</PlatformToolset>
```
sylveon, line 46: Instead of setting this here, we should simply delete `test/Directory.Build.props`. The issue with catch and VS 2022 appears to be fixed.

sylveon (changes requested):

kennykerr:
> unfortunately the build is failing because the agents do not have the v143 toolset installed

Is it just a matter of upgrading the solution to target 143 instead?

dmachaj:
> > unfortunately the build is failing because the agents do not have the v143 toolset installed
> 
> Is it just a matter of upgrading the solution to target 143 instead?

The error makes it sound like the build agent does not have it installed.  I suspect that no code changes to this PR are needed to fix the toolset.  Some sort of ADO "build agent configuration" change is needed instead.  Once that happens it would probably be a good thing to update _everything_ to VS2022, but that is scope creep beyond the minimum needed for this PR.

Directory.Build.Props
kennykerr: Fixed in #1187

strings/base_meta.h
```diff
@@ -1,7 +1,7 @@
 
 WINRT_EXPORT namespace winrt
 {
-    hresult check_hresult(hresult const result);
+    hresult check_hresult(hresult const result WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT);
```
kennykerr, line 4:
```suggestion
    hresult check_hresult(hresult const result WINRT_IMPL_SOURCE_LOCATION_ARGS);
```

strings/base_error.h
```diff
@@ -571,53 +587,53 @@ WINRT_EXPORT namespace winrt
...
 
-    [[noreturn]] inline void throw_last_error()
+    [[noreturn]] inline void throw_last_error(WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM)
     {
-        throw_hresult(impl::hresult_from_win32(WINRT_IMPL_GetLastError()));
+        throw_hresult(impl::hresult_from_win32(WINRT_IMPL_GetLastError()) WINRT_IMPL_SOURCE_LOCATION_FORWARD);
     }
 
-    inline hresult check_hresult(hresult const result)
+    inline hresult check_hresult(hresult const result WINRT_IMPL_SOURCE_LOCATION_ARGS)
```
kennykerr, line 595:
```suggestion
    inline hresult check_hresult(hresult const result WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT)
```

strings/base_array.h
```diff
@@ -511,24 +511,24 @@ WINRT_EXPORT namespace winrt
         return impl::com_array_proxy<T>(__valueSize, value);
     }
 
-    inline hstring get_class_name(Windows::Foundation::IInspectable const& object)
+    inline hstring get_class_name(Windows::Foundation::IInspectable const& object WINRT_IMPL_SOURCE_LOCATION_ARGS)
```
kennykerr, line 514: Given the `check_hresult` fix I mentioned, we should be able to hold off changing `base_array.h` at all.

kennykerr (changes requested):
Other than the `check_hresult` declaration needing the default (rather than the definition), this looks great. Ideally, we can eventually apply this to the projection itself and not just to explicit calls to `check_hresult` and exceptions, but I think this is a good start to get some experience with this new C++20 feature before applying it more broadly. Applying it more broadly is also really ugly as this parameter needs to be added to every single API call, which is gross.

strings/base_error.h
```diff
@@ -344,100 +356,104 @@ WINRT_EXPORT namespace winrt
 
     struct hresult_access_denied : hresult_error
     {
-        hresult_access_denied() noexcept : hresult_error(impl::error_access_denied) {}
-        hresult_access_denied(param::hstring const& message) noexcept : hresult_error(impl::error_access_denied, message) {}
-        hresult_access_denied(take_ownership_from_abi_t) noexcept : hresult_error(impl::error_access_denied, take_ownership_from_abi) {}
+        hresult_access_denied(WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM) noexcept : hresult_error(impl::error_access_denied WINRT_IMPL_SOURCE_LOCATION_FORWARD) {}
+        hresult_access_denied(param::hstring const& message WINRT_IMPL_SOURCE_LOCATION_ARGS) noexcept : hresult_error(impl::error_access_denied, message WINRT_IMPL_SOURCE_LOCATION_FORWARD) {}
+        hresult_access_denied(take_ownership_from_abi_t WINRT_IMPL_SOURCE_LOCATION_ARGS) noexcept : hresult_error(impl::error_access_denied, take_ownership_from_abi WINRT_IMPL_SOURCE_LOCATION_FORWARD) {}
     };
```
kennykerr, line 362: Not required for this PR, but ideally we can just have a macro to stamp out these derived exception types. These were all hand-rolled because originally C++/WinRT was 100% free of macros. Unfortunately, that wasn't very practical and it now relies on a fair number of macros. 😢

strings/base_array.h
dmachaj: I think you're right.  I was able to remove this and it still compiles 👍

strings/base_error.h
dmachaj: It is very much debatable, but I think this is still more readable than a macro would be (at the cost of increased duplication).  The new source_location macro definitely makes it uglier, but my personal opinion is that it is not yet worth such a conversion.  It is borderline, though.

kennykerr (approved):

PR #1187, opened 2022-09-16
Upgrade cppwinrt solution to Visual Studio 2022

kennykerr:
* Use the MSVC v143 toolset
* Set the C++20 test project to explicitly use C++20 now that it is an option
* Upgrade catch2 to the latest v2 version and remove the old build.props workaround


kennykerr:
Hopefully @Scottj1s knows how to upgrade the Azure build to use VS2022 instead of VS2019. 😬

dmachaj (approved):

dmachaj:
> Hopefully @Scottj1s Scott Jones FTE knows how to upgrade the Azure build to use VS2022 instead of VS2019. 😬

I was able to find the build pipeline definition and upgrade everything from windows-2019 to windows-2022.  That seems to have done the trick.

PR #1190, opened 2022-09-22
Add arm64 debug visualizer and update VS SDK package for VSIX

DefaultRyan:


build_vsix.cmd
```diff
@@ -25,12 +25,13 @@ call msbuild /m /p:Configuration=%target_configuration%,Platform=arm64,CppWinRTB
 rem Build cppwinrt.exe for x86 only
 call msbuild /m /p:Configuration=%target_configuration%,Platform=x86,CppWinRTBuildVersion=%target_version% cppwinrt.sln /t:cppwinrt
 
-rem Build cppwinrt visualizer dll for x86 and x64
+rem Build cppwinrt visualizer dll for x86, x64, and arm46
```
dmachaj, line 28: `arm46` should be `arm64`

DefaultRyan: Thanks.

dmachaj (dismissed):

DefaultRyan:
The build failure is because the vpack prerelease tag (like semver) doesn't like underscores in its name, and the prerelease tag is built from the branch name, which has an underscore. Everything else was successful.

Issue #1186, opened 2022-09-15
Zero GUID when unknwn.h is not included


mcooley:
Before #1180, this code successfully prints the GUID of ICompositorInterop. After #1180 it prints `{00000000-0000-0000-0000-000000000000}`:

```
#include <winrt/Windows.UI.Composition.h>
#include <windows.ui.composition.interop.h>

int main()
{
    printf("GUID for ICompositorInterop: %ls\n", winrt::to_hstring(winrt::guid_of<ABI::Windows::UI::Composition::ICompositorInterop>()).c_str());
}
```

C++/WinRT should just work in this case, or tell me that I need to include unknwn.h before including winrt/base.h.

Maybe the MSVC part of #1180 should be reverted, since it might not be needed? Maybe the [fallback definition of guid_v](https://github.com/microsoft/cppwinrt/blob/380cb8f773e1b9da8d947f6567eeaac401e0c33c/strings/base_meta.h#L130) should trigger an error when it's used?

kennykerr:
Yes, this looks like a regression which would be hard to catch. I think that #1180 didn't need to "fix" MSVC behavior even if it was questionable. @oldnewthing

kennykerr:
I had a look but couldn't find a way to get the fallback to produce an error.

sylveon:
How about:

```cpp
template <typename T>
struct com_guid
{
#if defined(_MSC_VER) && defined(WINRT_IMPL_IUNKNOWN_DEFINED)
    static constexpr guid value{ __uuidof(T) };
#endif
};

template <typename T>
inline constexpr guid guid_v = com_guid<T>::value;
```
This should error because `value` wouldn't exist.

While the compiler error may not be the easiest to understand, it definitely beats a strange runtime error.

kennykerr:
Sounds good! Want to send me a PR?

sylveon:
I can take a look sometime this weekend

PR #1193, opened 2022-09-29
Fix gcc build (and anyone else who doesn't support __uuidof)

oldnewthing:
Say `__uuidof` only if the compiler supports it as an intrinsic. Otherwise you get "undefined symbol" errors, even if the template is never invoked.

In theory we could add a `static_assert` to warn the user if they try to use `guid_v<T>` on a classic COM interface, but that would reintroduce the MSVC regression with complex templates. So we return to old behavior where `guid_v<T>` without classic COM support just produces the zero GUID.

### DIFF
--- a/strings/base_meta.h
+++ b/strings/base_meta.h
@@ -130,9 +130,12 @@ namespace winrt::impl
 
     template <typename T>
     inline constexpr guid guid_v = classic_com_guid<T>::value;
-#else
+#elif defined(_MSC_VER) && defined(WINRT_IMPL_IUNKNOWN_DEFINED)
     template <typename T>
     inline constexpr guid guid_v{ __uuidof(T) };
+#else
+    template <typename T>
+    inline constexpr guid guid_v{};
 #endif
 
     template <typename T>


oldnewthing:
This may be a dup of #1186

sylveon:
This brings back the issue where guid_v is silently 0 when winrt/base is included before any classic COM header. Since MSVC doesn't need to be guarded behind `IUnknown` being defined, removing defined(WINRT_IMPL_IUNKNOWN_DEFINED)` should fix that.

In my previous PR comment I had a suggestion that let other compilers error with a nice static_assert while also skipping the classic_com_guid struct while on MSVC. Does it not work here?

oldnewthing:
@sylveon From my experiments, it looks like using `__uuidof` without `WINRT_IMPL_IUNKNOWN_DEFINED` produces a

> error C2027: use of undefined type '_GUID'

however, the error is not raised until the `guid_v` is invoked, so it's okay to use it in MSVC mode even without `WINRT_IMPL_IUNKNOWN_DEFINED`. The caller will get an error about `_GUID`, which is acceptable.

So I think the original suggestion is fine.

* In MSVC, always use `__uuidof`. (Safe to utter its name even if `_GUID` is not defined.)
* In clang, use `__uuidof` only if enabled by `-fms-extensions`
* Everywhere else, avoid `__uuidof`.

sylveon:
Since this unspecialized guid_v is only invoked when the user uses classic COM with cppwinrt (which isn't extremely frequent) because it's explicitly specialized for WinRT objects, is the compile time impact that bad to warrant not just using classic_com_guid all the time?

oldnewthing:
Classic COM + C++/WinRT is a common scenario in some very large code bases (*cough* Windows *cough*).

sylveon:
How about this?
- On Clang with `-fms-extensions` enabled and MSVC, directly declare `guid_v` using `__uuidof` if `WINRT_IMPL_IUNKNOWN_DEFINED` is defined
- On other compilers, Clang with `-fms-extensions` is disabled, or on Clang with `-fms-extensions enabled` and MSVC when `WINRT_IMPL_IUNKNOWN_DEFINED` is not defined, fallback to using `classic_com_guid_error` to provide an error.

```cpp
    template <typename T>
    struct classic_com_guid_error
    {
#ifdef __clang__
#if !__has_declspec_attribute(uuid)
        static_assert(std::is_void_v<T> /* dependent_false */, "To use classic COM interfaces, you must compile with -fms-extensions.");
#endif 
#endif

#if defined(_MSC_VER) || defined(__clang__)
        static_assert(std::is_void_v<T> /* dependent_false */, "To use classic COM interfaces, you must include <unknwn.h> before including C++/WinRT headers.");
#else
        static_assert(std::is_void_v<T> /* dependent_false */, "Classic COM interfaces are not supported with this compiler.");
#endif
    };

    template <typename T>
#ifdef WINRT_IMPL_IUNKNOWN_DEFINED
#ifdef __clang__
#if __has_declspec_attribute(uuid)
    inline constexpr guid guid_v{ __uuidof(T) };
#else
    inline constexpr guid guid_v = classic_com_guid_error<T>::value;
#endif // __has_declspec_attribute(uuid)
#elif defined(_MSC_VER)
    inline constexpr guid guid_v{ __uuidof(T) };
#else
    inline constexpr guid guid_v = classic_com_guid_error<T>::value;
#endif // __clang__
#else
    inline constexpr guid guid_v = classic_com_guid_error<T>::value;
#endif // WINRT_IMPL_IUNKNOWN_DEFINED
```
This should result in:
- On Clang
  - If building without `-fms-extensions` you will get `To use classic COM interfaces, you must compile with -fms-extensions.`
  - If including `winrt/base.h` before `unknwn.h`, you will get `To use classic COM interfaces, you must  include <unknwn.h> before including C++/WinRT headers.`
  - If compiling without `-fms-extensions` and including `winrt/base.h` before `unknwn.h`, you will get both errors.
- On MSVC, if including `winrt/base.h` before `unknwn.h`, you will get `To use classic COM interfaces, you must  include <unknwn.h> before including C++/WinRT headers.`
- On other compilers, you will get `Classic COM interfaces are not supported with this compiler.`

This means clang also gets the "fast" compilation path where you skip instantiating a struct template if all preconditions are met.

PR #1196, opened 2022-10-07
Fix Clang-15 warnings

AtariDreams:
Mostly dealing with initialization and codestyle.

### DIFF
--- a/cppwinrt/cmd_reader.h
+++ b/cppwinrt/cmd_reader.h
@@ -196,7 +196,7 @@ namespace cppwinrt
     {
         std::string path(100, '?');
 
-        for (;;)
+        while (true)
         {
             DWORD actual_size = GetModuleFileNameA(nullptr, path.data(), 1 + static_cast<uint32_t>(path.size()));
 
@@ -602,7 +602,7 @@ namespace cppwinrt
             first_arg = true;
             *argument_count = 0;
 
-            for (;;)
+            while (true)
             {
                 if (*p)
                 {
@@ -620,7 +620,7 @@ namespace cppwinrt
                 if (*p == '\0')
                     break;
 
-                for (;;)
+                while (true)
                 {
                     copy_character = true;
 
--- a/cppwinrt/cmd_reader.h
+++ b/cppwinrt/cmd_reader.h
@@ -233,7 +233,7 @@ namespace cppwinrt
 
         auto key = open_sdk();
         uint32_t index{};
-        std::array<char, 100> subkey{};
+        std::array<char, 100> subkey;
         std::array<unsigned long, 4> version_parts{};
         std::string result;
 
--- a/cppwinrt/cmd_reader.h
+++ b/cppwinrt/cmd_reader.h
@@ -195,11 +195,10 @@ namespace cppwinrt
     inline std::string get_module_path()
     {
         std::string path(100, '?');
-        DWORD actual_size{};
 
-        while (true)
+        for (;;)
         {
-            actual_size = GetModuleFileNameA(nullptr, path.data(), 1 + static_cast<uint32_t>(path.size()));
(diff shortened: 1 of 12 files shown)


cppwinrt/component_writers.h
```diff
@@ -289,7 +289,7 @@ catch (...) { return winrt::to_hresult(); }
             bind<write_consume_args>(signature));
     }
 
-    void write_component_static_forwarder(writer& w, MethodDef const& method)
+    inline void write_component_static_forwarder(writer& w, MethodDef const& method)
```
kennykerr, line 292:
```suggestion
    static void write_component_static_forwarder(writer& w, MethodDef const& method)
```

cppwinrt/component_writers.h
```diff
@@ -643,8 +643,8 @@ catch (...) { return winrt::to_hresult(); }
         {
             if (!info.base && info.is_default)
             {
-                auto methods = info.type.MethodList();
-                offset += methods.second - methods.first;
+                auto [fst, snd] = info.type.MethodList();
+                offset += snd - fst;
```
kennykerr, line 647:
```suggestion
                auto [first, second] = info.type.MethodList();
                offset += second - first;
```

cppwinrt/cmd_reader.h
```diff
@@ -195,11 +195,10 @@ namespace cppwinrt
     inline std::string get_module_path()
     {
         std::string path(100, '?');
-        DWORD actual_size{};
 
-        while (true)
+        for (;;)
```
sylveon, line 199: I don't see how `for (;;)` is any clearer/better than `while (true)`

cppwinrt/text_writer.h
```diff
@@ -341,7 +341,7 @@ namespace cppwinrt
             std::string_view::size_type current_pos{ 0 };
             auto on_new_line = writer_base<T>::back() == '\n';
 
-            while (true)
+            for (;;)
```
sylveon, line 344: Idem

natvis/cppwinrt_visualizer.cpp
```diff
@@ -93,7 +93,7 @@ void LoadMetadata(DkmProcess* process, WCHAR const* processPath, std::string_vie
 {
     auto winmd_path = path{ processPath };
     auto probe_file = std::string{ typeName };
-    do
+    for (;;)
```
sylveon, line 96: Idem - just use `while (true)` instead of do while tho

strings/base_implements.h
```diff
@@ -692,7 +692,7 @@ namespace winrt::impl
         {
             uint32_t target = m_strong.load(std::memory_order_relaxed);
 
-            while (true)
+            for (;;)
```
sylveon, line 695: Idem

strings/base_implements.h
```diff
@@ -924,7 +924,7 @@ namespace winrt::impl
             {
                 uintptr_t count_or_pointer = m_references.load(std::memory_order_relaxed);
 
-                while (true)
+                for (;;)
```
sylveon, line 927: Idem

strings/base_implements.h
```diff
@@ -1048,7 +1048,7 @@ namespace winrt::impl
...
```
sylveon, line 1051: Idem

cppwinrt/cmd_reader.h
kennykerr: Yes, I prefer `while (true)` but I assumed Clang warns about this. Is that correct?

sylveon: Doesn't seem so: https://godbolt.org/z/GE9cW8v3z

kennykerr: Sweet, then I'd prefer to keep the while loops.

AtariDreams:
Fixed!

kennykerr (approved):
Thanks!

PR #1200, opened 2022-10-12
Improve compatibility with mingw-w64 in generated headers

alvinhochun:
This aims to make the generated headers usable by a Clang-based mingw-w64 toolchain. See https://github.com/mstorsjo/llvm-mingw/issues/307

Note that, even with these changes, the headers does not work with llvm-mingw 15 out of the box, because there are some functions and definitions missing in mingw-w64's version of `intrin.h`. We will consider fixing this upstream.

### DIFF
--- a/strings/base_includes.h
+++ b/strings/base_includes.h
@@ -1,4 +1,5 @@
 
+#include <intrin.h>
 #include <algorithm>
 #include <array>
 #include <atomic>
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -765,7 +765,7 @@ namespace cppwinrt
         {
             auto format = R"(    template <> struct abi<%>
     {
-        struct __declspec(novtable) type : inspectable_abi
+        struct WINRT_IMPL_NOVTABLE type : inspectable_abi
         {
 )";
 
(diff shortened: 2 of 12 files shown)


alvinhochun:

kennykerr:
> Ideally the CI would eventually be migrated to a yml that's committed here so 

I'd love to add a supplemental CI build using yml and GitHub Actions so that it is completely transparent - would appreciate such a contribution. The Azure build will probably stick around as its too complex to port to yml.

alvinhochun:
Thanks for the quick responses. I would love to have a CI build to check the llvm-mingw configuration, but first we need to add the missing stuff to mingw-w64 before we can try to implement the CI checks.

strings/base_includes.h
```diff
@@ -1,4 +1,6 @@
 
+#include <intrin.h>
+
```
kennykerr, line 3:
```suggestion
#include <intrin.h>
```

kennykerr:
The build failed to compile all projects. Did you test this locally?

alvinhochun:
Sorry about that, I only tested the "test" project before so I did not realize how it breaks `fast_forward.h` ("test_fast" would reveal this).

(Full disclosure: I have to downgrade the projects to toolset v142 because I don't have VS2022 installed, and it fails to build "test_cpp20" and "test_old", so I still haven't been able to run these two tests.)

kennykerr (approved):
Thanks!

PR #1201, opened 2022-10-14
Add GitHub Action build and test workflow

alvinhochun:
This is *not* intended to be a replacement to the Azure build, but as a supplemental build. The added workflow performs builds and runs tests on x86 and x64, and also tries to build the nuget package.

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -1,6 +1,9 @@
 name: MSVC Tests
 on:
   push:
+  pull_request:
+    branches:
+      - master
 
 jobs:
   test-cppwinrt:
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -23,4 +23,17 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: test-output-${{ matrix.arch }}-${{ matrix.config }}
-          path: ${{ matrix.arch }}_results.txt
+          path: "*_results.txt"
+
+      - name: Check test failure
+        run: |
+          if (Test-Path "test_failures.txt") {
+            Get-Content "test_failures.txt" | ForEach-Object {
+              Write-Error "error: Test '$_' failed!"
+            }
+            return 1
+          }
+          if (!(Test-Path "*_results.txt")) {
+            Write-Error "error: No test output found!"
+            return 1
+          }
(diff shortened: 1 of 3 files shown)


kennykerr:
I forget the magical incantation to get GitHub to notice there's a new workflow in a PR... 🤔

sylveon:
Should we try at least building the tests on ARM64?

kennykerr:
> I forget the magical incantation to get GitHub to notice there's a new workflow in a PR... 🤔

It might not run until the workflow is in master, in which case we can just merge and then validate with a subsequent PR to test the workflow. 

> Should we try at least building the tests on ARM64?

Does GitHub support ARM64 builds now?

sylveon:
The CI image is shared with Microsoft hosted pipeline agents, and it has the ARM64 cross compiler installed. It's not an ARM64 machine though, so we can only try building the tests as basic sanity check, can't actually run them.

kennykerr:
Got it, yes that's all the Azure build does as well.

kennykerr:
@alvinhochun I enabled GitHub Actions on this repo after you created this PR, so I think you just need to push another commit to trigger the new workflow to run.

riverar:
@sylveon This isn't related to the PR here but just a heads up, Microsoft has private ARM64 runners in ADO.

sylveon:
Yeah I heard about it but I'm not in the preview

alvinhochun:
The workflow is being prevented from running: https://github.com/microsoft/cppwinrt/actions/runs/3252917426

> actions/checkout@v3 is not allowed to be used in microsoft/cppwinrt. Actions in this workflow must be: within a repository that belongs to your Enterprise account or matching the following: actions/checkout@v2, actions/upload-artifact@v3.

kennykerr:
I updated it to allow `actions/checkout@v3` but will have to be restarted somehow...

alvinhochun:
Pushed again and now they're running.

kennykerr:
Woohoo - now we need Clang and GNU workflows. 😜

kennykerr (approved):
Thanks!

kennykerr:
@alvinhochun the actions seem a little unreliable: 

https://github.com/microsoft/cppwinrt/actions/runs/3267768111/jobs/5373367929

Rerunning often seems to fix them, but it would be nice to track down what's causing the frequent failures.

kennykerr:
Never mind - this is another case of #1202 - be nice if the error was easier to see tho 😜

kennykerr:
Sounds great!

PR #1205, opened 2022-10-15
check_bool<T> now returns T

### LINKED ISSUE #1204: check_bool<T> should return T
Today, winrt::check_bool<T> returns void. There are some Win32 APIs that return 0 on failure and a valid result on success eg: [GetFileVersionInfoSizeW](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfosizew) so having winrt::check_bool<T> returning T would be convenient while using them.

Jaiganeshkumaran:
Closes #1204.


kennykerr:
Thanks, just needs a test. [Perhaps here](https://github.com/microsoft/cppwinrt/blob/464e59198c2046125d8ee5a8cc6ff1739097345f/test/old_tests/UnitTests/Errors.cpp).

Jaiganeshkumaran:
> Thanks, just needs a test. [Perhaps here](https://github.com/microsoft/cppwinrt/blob/464e59198c2046125d8ee5a8cc6ff1739097345f/test/old_tests/UnitTests/Errors.cpp).

Is it fine now?

test/old_tests/UnitTests/Errors.cpp
```diff
@@ -200,6 +200,8 @@ TEST_CASE("Errors")
     SetLastError(ERROR_CANCELLED);
     REQUIRE_THROWS_AS(check_bool(static_cast<BOOL>(false)), hresult_canceled);
 
+    REQUIRE_NOTHROW(check_bool(true) == true);
```
kennykerr, line 203: `REQUIRE_NOTHROW` just asserts that the expression does not throw an exception. What you're trying to do is check that the expression is true. For that you should use the `REQUIRE` test macro.
```suggestion
    REQUIRE(check_bool(true) == true);
```

kennykerr (approved):

PR #1209, opened 2022-10-17
Update dependencies

AtariDreams:
Dependencies include: Catch, .NET SDK, etc.


kennykerr:
I'm reluctant to change these dependencies as they are tied to the VS ingestion process that is not tested here. I'll leave that for @DefaultRyan to handle as needed.

PR #1210, opened 2022-10-18
Improve CI jobs split

alvinhochun:
This basically rewrites the CI jobs based on the original `build_test_all.cmd` but split the steps into different jobs so any failures will be easier to pinpoint. This also allows the tests to be built and run in parallel, which can save about 5 minutes of full run time.

One thing to note is that the nuget test build seems to be failing even on the initial GHA workflow. This failure was hidden before because `build_test_all.cmd` did not detect the build error. Now that it is exposed, I have added a hack to purposely ignore the failure (instead only emit a warning) because I don't know how to fix it properly.

### DIFF
--- a/test/test/disconnected.cpp
+++ b/test/test/disconnected.cpp
@@ -44,94 +44,96 @@ namespace
     }
 }
 
-TEST_CASE("disconnected,handler")
+TEST_CASE("disconnected,handler,1")
 {
-    {
-        event<EventHandler<int>> source;
+    event<EventHandler<int>> source;
 
-        source.add([](auto...)
-            {
-                throw hresult_error(RPC_E_DISCONNECTED);
-            });
+    source.add([](auto...)
+        {
+            throw hresult_error(RPC_E_DISCONNECTED);
+        });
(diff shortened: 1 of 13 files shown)


test/test/disconnected.cpp
```diff
@@ -44,94 +44,96 @@ namespace
     }
 }
 
-TEST_CASE("disconnected,handler")
+TEST_CASE("disconnected,handler,1")
```
kennykerr, line 47: What's the `,1` for?

test/test/disconnected.cpp
```diff
@@ -44,94 +44,96 @@ namespace
...
+    REQUIRE(!source);
 
-        source(nullptr, 123);
-    }
+    source(nullptr, 123);
+}
 
-    {
-        auto async = Action();
+TEST_CASE("disconnected,handler,2")
```
kennykerr, line 81: Oh, you flattened these tests...

kennykerr (approved):
Much nicer output - thanks!!

PR #1212, opened 2022-10-20
Better support for mingw-w64 and GCC-compatible compilers

alvinhochun:
There are still some intrinsics and defines missing in mingw-w64 (see https://github.com/mstorsjo/llvm-mingw/issues/307) which can all be replaced with other compatible constructions. Verified by compiling a sample program locally for i686, x86_64 and aarch64 (aarch64 build is untested because I don't have a compatible system).

(CI checks for mingw-w64 will be coming soon™)

CC: @mstorsjo @cristianadam

### DIFF
--- a/strings/base_activation.h
+++ b/strings/base_activation.h
@@ -166,7 +166,7 @@ namespace winrt::impl
         return result;
 #elif defined _M_ARM64
 #if defined(__GNUC__)
-        int32_t const result = *target;
+        int64_t const result = *target;
 #else
         int64_t const result = __iso_volatile_load64(target);
 #endif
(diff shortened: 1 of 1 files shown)


kennykerr:
I'm not familiar with the GCC intrinsics. Does someone else want to take a stab at reviewing this? @DefaultRyan @oldnewthing

BenJKuhn (approved):
I can't speak to the completeness of these changes wrt to making mingw work, but the changes themselves seem correct based on a close reading. The use of __iso_volatile_load64 for the MSVC implementation is to avoid a perf hit due to MS-specific behavior described here: https://learn.microsoft.com/en-us/cpp/build/reference/volatile-volatile-keyword-interpretation?view=msvc-170. It should be unnecessary in MinGW, though if it adds the same memory barriers MS does for volatile reads, it may incur a perf penalty. I couldn't find a quick answer to that. It'll be correct and usable though either way.

PR #1215, opened 2022-10-21
Enable classic COM on mingw-w64

alvinhochun:
Mingw-w64 does not use `__declspec(uuid(x))` but has its own way of supporting `__uuidof` by defining UUIDs with a special macro `__CRT_UUID_DECL`, which does actually work here.


kennykerr:
Here we are again with classic COM. 😅 @sylveon do you want to take a stab at reviewing this, given your recent overhaul of this code?

sylveon:
From a quick read, it appears that everything is fine here.

kennykerr (approved):

PR #1216, opened 2022-10-21
Add CMake build and a limited subset of tests for llvm-mingw

alvinhochun:
This add a CMake build for building cppwinrt and a limited subset of the existing tests using the [llvm-mingw](https://github.com/mstorsjo/llvm-mingw) toolchain. A GHA CI build using this toolchain is also included.

More tests will be added later.

Depends on https://github.com/microsoft/cppwinrt/pull/1212 and https://github.com/microsoft/cppwinrt/pull/1215

CC @mstorsjo


kennykerr:
So getting mingw to work with msbuild like we do for Clang is not practical?

test/test/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+    CONFIGURE_DEPENDS
+    *.cpp
+)
+list(FILTER TEST_SRCS EXCLUDE REGEX "/(main|pch)\\.cpp")
+
+
+# We can't build test_component[*] for mingw-w64 because it doesn't have an
+# alternative to midl that can produce winmd files. Also, even if we do manage
+# to reuse the MSVC-compiled binaries, mingw-w64 is still missing
+# windowsnumerics.impl.h which is needed to provide the types
```
kennykerr, line 12: The Clang builds depend on the Windows SDK, just like MSVC. I take it mingw doesn't?

sylveon:
For clang, msbuild supports it through clang-cl. In GCC/mingw's case, there is no such tool, and VS doesn't have native support

test/test/CMakeLists.txt
sylveon: They have their own equivalent headers due to the Windows SDK license not allowing redistribution, and the SDK headers depending on too many MSVC-isms

mati865: It's also important to distinguish key aspect here: Clang can target both MSVC ABI (using Windows SDK) and GCC ABI (using mingw-w64).

alvinhochun:
Yeah, most has been said already. Another point is that mingw-w64 is commonly used as a way to cross-compile applications for Windows from another Unix-like OS, so headers and tools from MSVC or the Windows SDK usually does not work even if we ignore the licensing issues. (I will also have to look into this aspect later -- I'm not sure how we can use the cppwinrt tool and/or its generated headers for cross-compilation.)

alvinhochun:
I would like this to be merged so we can at least have some CI coverage for llvm-mingw. I can continue to add more tests later.

mingw-support/xmllite.def
```diff
@@ -0,0 +1,13 @@
+;
```
kennykerr, line 1: These are very simple def files. I'm not sure what value it provides to have them generated by a tool and what licensing issues may occur as a result. Can we remove the comments?

kennykerr (approved):

Issue #1195, opened 2022-10-06
C++ Streams support for hstring and IStringable


Jaiganeshkumaran:
Today it doesn't seem possible to output an hstring directly to a C++ stream (eg: wcout). 

```
cpp
hstring text;
// Doesn't work.
std::wcout << text;

// Works.
std::wcout << text.c_str();
std::wcout << static_cast<std::wstring_view>(text);
```

Please overload operator<< for hstring, as well as types implementing IStringable.

kennykerr:
This isn't supported but is trivial to work around as you've illustrated. Most applications don't require this support, so it wasn't deemed a necessary dependency.

sylveon:
if you really want `std::wcout << hstr` to compile:
```cpp
std::wostream& operator<<(std::wostream& os, winrt::hstring const& str)
{
    os << static_cast<std::wstring_view>(str);
    return os;
}
```

`IStringable` left as an exercise to the reader ;)

Jaiganeshkumaran:
I would still prefer it to be built-in. Even if Windows GUI apps aren't going to use wcout, they can still use other stream types like wstringstream and wifstream.

PR #1221, opened 2022-10-30
C++ Streams support for hstring and IStringable 

Jaiganeshkumaran:
Implements #1195.


strings/base_stringable_format.h
```diff
@@ -1,3 +1,8 @@
+inline std::wostream& operator<<(std::wostream& stream, winrt::Windows::Foundation::IStringable const& stringable)
```
sylveon, line 1: Use a different file for this. base_stringable_format is for std::format. base_stringable_streams?

Jaiganeshkumaran: Done.

strings/base_includes.h
```diff
@@ -10,6 +10,7 @@
 #include <map>
 #include <memory>
 #include <optional>
+#include <ostream>
```
kennykerr, line 13: Is there no way to use a feature macro rather than including this in every single project? That seems problematic.

Jaiganeshkumaran: Maybe this could be included only if WINRT_LEAN_AND_MEAN isn't defined. Or should I define a separate macro?

kennykerr: Using `WINRT_LEAN_AND_MEAN` seems like a good idea. 👍

Jaiganeshkumaran: Done.

kennykerr:
Looks good - just needs a basic test.

test/old_tests/UnitTests/streams.cpp
```diff
@@ -0,0 +1,36 @@
...
+        std::wstringstream ss;
+        winrt::hstring str = L"Hello\0World";
+        ss << str;
+        REQUIRE(ss.str() == str);
+    }
+
+    {
+        std::wstringstream ss;
+        winrt::Windows::Foundation::IStringable obj = winrt::make<stringable>();
+        ss << obj.ToString();
```
kennykerr, line 33: This tests `hstring` rather than `IStringable`.

kennykerr (approved):
Thanks!

PR #1225, opened 2022-11-07
Fix invoke call on Clang (possibly a compiler bug)

alvinhochun:
For unknown reasons, Clang resolves the `invoke` call (in `winrt::impl::promise_base::Completed` and `winrt::impl::promise_base::set_completed`) to `std::invoke` instead of `winrt::impl::invoke`, which breaks exception handling and causes some disconnected.cpp tests to crash. After changing the call to specify the namespace, 4 more tests in `disconnected.cpp` now passes on Clang.

---

To demonstrate the issue, I modified the generated `winrt/Windows::Foundation.h` and duplicated line 3765 (`winrt::impl::promise_base::set_completed`), and changed it so that one call specifies the namespace and the other doesn't. The disassembly shows how the name resolution of the two calls differ (this is with clang+mingw but the same happens with clang+msvc):

```
** 3763             if (handler)
   3764             {

    0x1ea95c <+156>: leal   -0x8(%ebp), %ecx
    0x1ea95f <+159>: calll  0x155e70                  ; winrt::Windows::Foundation::IUnknown::operator bool at base.h:2139
    0x1ea964 <+164>: testb  $0x1, %al
    0x1ea966 <+166>: jne    0x1ea971                  ; <+177> at Windows.Foundation.h
    0x1ea96c <+172>: jmp    0x1ea9a7                  ; <+231> at Windows.Foundation.h:3768:9

   1    // WARNING: Please don't edit this file. It was generated by C++/WinRT v2.3.4.5

    0x1ea971 <+177>: movl   -0x14(%ebp), %ecx

-> 3765                 invoke(handler, *this, status);

->  0x1ea974 <+180>: movl   %esp, %eax
    0x1ea976 <+182>: leal   -0xc(%ebp), %edx
    0x1ea979 <+185>: movl   %edx, 0x8(%eax)
    0x1ea97c <+188>: movl   %ecx, 0x4(%eax)
    0x1ea97f <+191>: leal   -0x8(%ebp), %ecx
    0x1ea982 <+194>: movl   %ecx, (%eax)
[...]

PR #1226, opened 2022-11-08
Try to avoid random crash with disconnected.cpp test

alvinhochun:
There are some test cases which tests throwing an hresult_error from the completed handler. This relies on `winrt::impl::invoke` catching the exception and handling it using `winrt::to_hresult`. In some test cases, the completed handler may run on a worker thread, so the test tries to wait for the handler to be called using an event, but it does not wait until `winrt::impl::invoke` has finished handling the exception. An extremely rare race condition may happen:

1. The completed handler on a worker thread sets the event.
2. The test case, having finished waiting for the event on the main thread, leaves the test case.
3. Another test case, `custom_error`, starts on the main thread and sets `winrt_to_hresult_handler` to a custom handler
4. The completed handler, which is still running on the worker thread, throws the `hresult_error` exception.
5. `winrt::impl::invoke` catches the exception, then tries to handle it with `winrt::to_hresult`.
6. `winrt::to_hresult` notices the custom `winrt_to_hresult_handler` from the `custom_error` test and calls it, instead of going through the usual code flow that is expected by the original test case.
7. The `winrt_to_hresult_handler` doesn't know how to handle `hresult_error`, so it ends up unhandled and terminates the program.

This is a real crash I have got when testing a mingw build with some tests excluded, which triggered this exact sequence of events. In theory this can also happen when testing with `--order rand` to randomize test order if you are very unlucky.

I cannot think of a way to really wait for the completed handler to finish, so I added a 500ms sleep at the end of the test to hopefully minimize the chance of this happening again.

### DIFF
--- a/test/test/disconnected.cpp
+++ b/test/test/disconnected.cpp
@@ -105,6 +105,8 @@ TEST_CASE("disconnected,handler,3")
         });
 
     WaitForSingleObject(signal.get(), INFINITE);
+    // Give some time for to_hresult() to complete.
+    Sleep(500);
 }
 
 TEST_CASE("disconnected,handler,4")
@@ -134,6 +136,8 @@ TEST_CASE("disconnected,handler,5")
         });
 
     WaitForSingleObject(signal.get(), INFINITE);
+    // Give some time for to_hresult() to complete.
+    Sleep(500);
 }
 
 // Custom action to simulate an out-of-process server that crashes before it can complete.
--- a/test/test_win7/disconnected.cpp
+++ b/test/test_win7/disconnected.cpp
@@ -94,6 +94,8 @@ TEST_CASE("disconnected,3")
         });
 
     WaitForSingleObject(signal.get(), INFINITE);
+    // Give some time for to_hresult() to complete.
+    Sleep(500);
 }
 
 TEST_CASE("disconnected,4")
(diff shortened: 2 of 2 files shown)


test/test/disconnected.cpp
```diff
@@ -105,6 +105,8 @@ TEST_CASE("disconnected,handler,3")
         });
 
     WaitForSingleObject(signal.get(), INFINITE);
+    // Give some time for to_hresult() to complete.
+    Sleep(500);
```
kennykerr, line 109: Shouldn't we build this buffer into the test that fiddles with `winrt_to_hresult_handler` rather than the test(s) that observe the fiddling? I think there's only one place where this is tested (`custom_error.cpp`).

alvinhochun: I think putting this in the disconnected test itself is better because it makes sure (or at least tries to) that any code related to this test running on the worker thread has really finished before leaving the test scope. In case it actually causes `std::terminate`, Catch2 can report that the abnormal termination happens during the correct test, instead of making it look like another test which happens to be running afterwards caused it. (This actually did happen to these tests before PR #1225.)

kennykerr:
Not sure why Azure won't kick off an internal build on PRs anymore. Anyway, I'm pretty confident the GitHub validation is sound and the Azure build still runs on master, so I'll merge this.

PR #1227, opened 2022-11-09
Add fields to filter out templates appearing in the add new item dialog for non UAP projects.

RaulPerez1:


Jaiganeshkumaran:
ViewModel template should still be available in Windows Desktop projects, because it's for creating a Windows Runtime class and it's not specifically for UWP XAML (although it seems to include UWP XAML headers).

Scottj1s (approved):
@RaulPerez1 thanks for the VSIX polish

JaiganeshKumaran:
This does not seem to work properly. I still see C++/WinRT XAML templates in a WinUI 3 project.

PR #1229, opened 2022-11-17
Enable more tests on llvm-mingw and some fixes

alvinhochun:
Changes include:

- Since a partial `windowsnumerics.impl.h` header has been added to mingw-w64 recently, we can enable the tests that depend on it. If it is not available in the current mingw-w64 toolchain, there is an option to download a copy of it from the mingw-w64 git repo to allow these tests to run.
- Enable some test cases of `test_cpp20` and `test_old` for llvm-mingw. `test_old` in particular depends on `windowsnumerics.impl.h`)
- Re-enable test code for Clang fixed by #1228.
- Enable some test cases for llvm-mingw that crashes only when built using Clang with MSVC ABI.
- Make some tests relying on error messages pass on non-English Windows systems.
- Use lowercase filename when including Windows headers, which is needed for cross-building from Linux with mingw-w64.

### DIFF
--- a/test/old_tests/UnitTests/apartment_context.cpp
+++ b/test/old_tests/UnitTests/apartment_context.cpp
@@ -46,6 +46,11 @@ namespace
         co_await context;
     }
 
+// Not yet buildable on mingw-w64. The lambda needs to have __stdcall
+// specified on it but there is a Clang crash bug blocking this:
+// https://github.com/llvm/llvm-project/issues/58366
+#if !defined(__MINGW32__)
+
     template<typename TLambda>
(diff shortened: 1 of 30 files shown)


test/old_tests/UnitTests/apartment_context.cpp
```diff
@@ -46,6 +46,11 @@ namespace
         co_await context;
     }
 
+// Not yet buildable on mingw-w64. The lambda needs to have __stdcall
```
sylveon, line 49: Since this lambda has no captures, it can be converted to a normal function with __stdcall.

kennykerr (approved):

PR #1233, opened 2022-11-20
Some minor changes to CMakeLists.txt

alvinhochun:


CMakeLists.txt
```diff
@@ -164,4 +171,6 @@ target_include_directories(cppwinrt PRIVATE "${winmd_SOURCE_DIR}")
 
 
 include(CTest)
-add_subdirectory(test)
+if(BUILD_TESTING)
```
walbourn, line 174: Can you confirm it still includes testing by default?

alvinhochun: Yes, the option is created by the CTest module and is `ON` by default: https://cmake.org/cmake/help/latest/module/CTest.html

PR #1240, opened 2022-11-27
Support for XAML types in nested namespaces

Jaiganeshkumaran:
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/7652.


kennykerr:
Such a change would need to be reviewed and signed off by the Xaml compiler team. They would also need to validate the change with a full OS build since it could affect other builds as well. I suggest you reopen the issue on the Xaml repo and work with that team to make any needed change. cc @Scottj1s

JaiganeshKumaran:
I've reopened the issue in the WinUI repo, but I think it's very unlikely for them to fix the issue anytime soon. There are already many C++ specific issues with the XAML compiler that were reported ages ago and nothing was done.

PR #1242, opened 2022-12-04
CI: Refactor LLVM setup steps and use cache

alvinhochun:
I extracted the LLVM setup steps into composite actions and enabled caching the LLVM toolchains. The clang-cl jobs are perhaps slightly faster now but not a significant difference.

### DIFF
--- a/.github/actions/setup-llvm-mingw/action.yml
+++ b/.github/actions/setup-llvm-mingw/action.yml
@@ -1,0 +1,46 @@
+name: 'Set up llvm-mingw toolchain'
+description: 'Set up llvm-mingw toolchain'
+inputs:
+  llvm-mingw-version:
+    description: 'llvm-mingw version'
+    required: true
+    default: '20220906'
+  host-arch:
+    description: 'llvm-mingw toolchain host architecture (e.g. i686, x86_64)'
+    required: true
+    default: 'x86_64'
+outputs:
+  llvm-path:
+    description: "The path in which llvm-mingw is installed to"
+    value: ${{ steps.setup-llvm.outputs.llvm-path }}
+runs:
+  using: "composite"
+  steps:
+    - name: Cache llvm-mingw
+      id: cache-llvm
+      uses: actions/cache@v3
+      with:
+        path: .llvm-mingw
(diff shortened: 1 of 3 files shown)


sylveon:
Why are we not using the preinstalled LLVM that comes with VS (and its platform toolset)? That'll avoid the download altogether and the vs version would be what most people who want to use clang-cl end up using.

alvinhochun:
I tend to use the latest release available to try things so that's what I went with initially. I suppose it may make sense to add CI jobs to test the LLVM build that comes with VS in addition to the latest upstream release, but I wouldn't replace the current ones entirely.

kennykerr (approved):

kennykerr:
These LLVM tests seems to fail pretty regularly. Any ideas? 

https://github.com/microsoft/cppwinrt/actions/runs/3620984131/jobs/6104646572

alvinhochun:
Not really. I don't know if this is true, but to me it feels like this specific failure only started happening quite recently on the runner. I have got it a lot in my fork too. I tried to debug it locally but I can't get it to happen at all on my system with the llvm-mingw build.

I do manage to get it on my local GCC build (I made a few hacks to get it to build), roughly 1 out of 20 tries? It might help me figure this out.

oldnewthing:
@kennykerr I don't see anything compiler-specific in that test. The test logs don't really say anything. I can't even find where it says which test failed!

kennykerr:
A re-run made the problem go away... 🤷‍♂️

sylveon:
It's often the same test. Did this happen before my change removing async hooks?

kennykerr:
That's possible - I'm not sure.

alvinhochun:
From rerunning some old CI jobs in my fork, it seems the failure started happening after enabling ASan in the llvm-mingw Debug builds (#1232). But I do test this config locally since that PR and I haven't got this failure at all. So there seems to be other factors at play, or perhaps it is an odd race condition which happens to trigger very often on the GHA runner?

If the failure gets too annoying we can mark this job to continue-on-error...

alvinhochun:
Hmm, actually it seems to be also happening to x86_64 release builds, but since there were no test for llvm-mingw release builds before #1232 I don't have any more data points.

Issue #1243, opened 2022-12-07
Cancel propagation broken for nested IAsyncAction after #1228


alvinhochun:
Continuing the discussion in #1242 

@sylveon It seems your change in #1228 did break cancel propagation, which shows up in the async_propagate_cancel test inside `Check([] { return ActionAction(10); });` sporadically, because it only fails if `async.Cancel()` is called after the first call to `co_await ActionAction(wait, depth - 1);` has already gone through `await_transform` (`base_coroutine_foundation.h`):

https://github.com/microsoft/cppwinrt/blob/4a5acf64457c82dbf1ce6f2e81bc0bdc1adbe51e/strings/base_coroutine_foundation.h#L603-L621

Line 611-618 is responsible for propagating the cancellation to the child coroutine. The condition and cast here is:

```c++
if constexpr (std::is_convertible_v<std::remove_reference_t<decltype(expression)>&, enable_await_cancellation&>) {
    //[...]
    static_cast<enable_await_cancellation&>(expression).set_cancellable_promise(&m_cancellable);
```

But before #1228, the check inside `notify_awaiter` was:

```c++
if constexpr (std::is_convertible_v<std::remove_reference_t<decltype(awaitable)>&, enable_await_cancellation&>) {
    // [...]
    static_cast<enable_await_cancellation&>(awaitable).set_cancellable_promise(promise /* same as &m_cancellable */);
```

`awaitable` was declared as:

```c++
decltype(get_awaiter(std::declval<T&&>())) awaitable
    /* effectively: */ = get_awaiter(static_cast<T&&>(expression));
```

`get_awaiter` makes use of `operator co_await` to convert the awaitables. The following applies:

https://github.com/microsoft/cppwinrt/blob/4a5acf64457c82dbf1ce6f2e81bc0bdc1adbe51e/strings/base_coroutine_foundation.h#L251-L272

Because this conversion does not happen anymore after #1228, the cancellation is not propagated properly, therefore the child coroutines don't get cancelled, and depending on how late `async.Canecl()` is called, the completion handler may not be invoked. My change in https://github.com/alvinhochun/cppwinrt/commit/c34c31ab9eac6c371aa0c90ffbbcf03913148069 triggers this reliably.

[...]

PR #1245, opened 2022-12-11
Make compatible with GCC

alvinhochun:
Changes for GCC compatibility:

- Added GCC-only hack for `winrt::impl::bind_in` mentioned in https://github.com/microsoft/cppwinrt/pull/1234 (upstream: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61663)
- Fixed the `WINRT_IMPL_LINK` macro to properly declare the symbol as global (__GNUC__).
- Added GCC-only hack move constructors to the awaitables in `resume_after` and `resume_on_signal` because GCC adds useless moves in `co_await` (suspect to be https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99575)
- ~~Added GCC-only hack copy constructor to `thread_pool` because GCC seems to want it when building the `thread_pool.cpp` test (suspect to be https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103963)~~
    - ~~Note that this also changes the `m_pool` field to be a `std::shared_ptr`.~~
    - (reverted)
- Fixed signed overflow in `winrt::clock` if `system_clock` has nanosecond resolution
- Miscellaneous test fixes and disabled some unbuildable or unreliable tests
- Added CI for GCC using msys2
    - I also tried adding a cross build test, but Ubuntu has only GCC 10 and even the mingw-w64 on Arch Linux isn't recent enough to have sufficiently working headers (some fixes are not in the latest release).

General changes:

- Enabled colour diagnostics and test output for all CI builds
- ~~Include `directxmath.h` only if it exists (for compatibility with older mingw-w64 headers)~~ (reverted)
- Tagged async_propagate_cancel test as mayfail so that it doesn't block CI (please fix with https://github.com/microsoft/cppwinrt/issues/1243)

I am not happy with the move/copy constructor hacks, but they seem unavoidable unless GCC have their bugs fixed...

CC @Biswa96 @mstorsjo

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -556,35 +556,18 @@ WINRT_EXPORT namespace winrt
         return awaitable{ handle, timeout };
     }
 
-    // HACK: GCC does not compile co_await on thread_pool because it wants
-    // a copy constructor. Since m_pool is not copyable, in order to provide
-    // a workaround we have to wrap it in a shared_ptr. Apply this workaround
-    // only for GCC so we don't affect other compilers.
-    // This might be related to upstream bug:
-    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103963
     struct thread_pool
     {
         thread_pool() :
-#if defined(__GNUC__) && !defined(__clang__)
-            m_pool(std::make_shared<handle_type<pool_traits>>(
-                check_pointer(WINRT_IMPL_CreateThreadpool(nullptr))))
-#else
             m_pool(check_pointer(WINRT_IMPL_CreateThreadpool(nullptr)))
-#endif
         {
-            m_environment.Pool = get_pool();
+            m_environment.Pool = m_pool.get();
         }
-
-#if defined(__GNUC__) && !defined(__clang__)
-        // HACK: See above
-        thread_pool(thread_pool const&) = default;
-        thread_pool &operator=(thread_pool const&) = delete;
-#endif
 
         void thread_limits(uint32_t const high, uint32_t const low)
         {
-            WINRT_IMPL_SetThreadpoolThreadMaximum(get_pool(), high);
-            check_bool(WINRT_IMPL_SetThreadpoolThreadMinimum(get_pool(), low));
+            WINRT_IMPL_SetThreadpoolThreadMaximum(m_pool.get(), high);
+            check_bool(WINRT_IMPL_SetThreadpoolThreadMinimum(m_pool.get(), low));
         }
 
         bool await_ready() const noexcept
@@ -649,21 +632,7 @@ WINRT_EXPORT namespace winrt
             uint32_t Size{ sizeof(environment) };
         };
 
-        pool_traits::type get_pool() const
-        {
-#if defined(__GNUC__) && !defined(__clang__)
-            return m_pool->get();
-#else
-            return m_pool.get();
-#endif
-        }
-
-#if defined(__GNUC__) && !defined(__clang__)
-        // HACK: See above
-        std::shared_ptr<handle_type<pool_traits>> m_pool;
-#else
         handle_type<pool_traits> m_pool;
-#endif
         environment m_environment;
     };
 
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -556,18 +556,35 @@ WINRT_EXPORT namespace winrt
         return awaitable{ handle, timeout };
     }
 
+    // HACK: GCC does not compile co_await on thread_pool because it wants
+    // a copy constructor. Since m_pool is not copyable, in order to provide
+    // a workaround we have to wrap it in a shared_ptr. Apply this workaround
+    // only for GCC so we don't affect other compilers.
+    // This might be related to upstream bug:
+    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103963
     struct thread_pool
     {
         thread_pool() :
+#if defined(__GNUC__) && !defined(__clang__)
+            m_pool(std::make_shared<handle_type<pool_traits>>(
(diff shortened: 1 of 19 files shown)


test/old_tests/UnitTests/hstring.cpp
```diff
@@ -272,7 +272,6 @@ TEST_CASE("hstring,operator,std::wstring_view")
     REQUIRE(L"abc" == ws);
 
     hs.clear();
-    REQUIRE(L"abc" == ws);
```
alvinhochun, line 275: I'm not sure what's the point of this test because this seems to be a use-after-free, and it crashes when run with Application Verifier enabled. So I removed it.

alvinhochun:
Ah the msys2 workflow uses the `msys2/setup-msys2@v2` action. @kennykerr can you allow it too?

strings/base_includes.h
```diff
@@ -27,8 +27,10 @@
 
 #if __has_include(<windowsnumerics.impl.h>)
 #define WINRT_IMPL_NUMERICS
+#if __has_include(<directxmath.h>)
```
sylveon, line 30: Is it intentional that directxmath.h is only included if windowsnumeric.impl.h is available?

kennykerr: This was done to work around a C++ modules bug. It's required by numerics but not by cppwinrt itself.

kennykerr: Feel free to remove it if it helps.

sylveon: Ah, can probably keep for now and I'll keep it in mind when I look into modules again soon

strings/base_coroutine_threadpool.h
```diff
@@ -528,18 +556,35 @@ WINRT_EXPORT namespace winrt
         return awaitable{ handle, timeout };
     }
 
+    // HACK: GCC does not compile co_await on thread_pool because it wants
```
kennykerr, line 559: This is just a helper - how about we just exclude it for GCC? Not loving all the hackery to get GCC to work...

test/old_tests/UnitTests/hstring.cpp
kennykerr: Good catch.

kennykerr:
> Ah the msys2 workflow uses the `msys2/setup-msys2@v2` action. @kennykerr can you allow it too?

done

strings/base_coroutine_threadpool.h
alvinhochun: Okay, I can revert this one and just exclude the `thread_pool` tests from the GCC build.

test/test/custom_error.cpp
```diff
@@ -102,10 +102,21 @@ TEST_CASE("custom_error_logger")
     // Validate that handler translated exception
     REQUIRE_THROWS_AS(check_hresult(0x80000018), hresult_illegal_delegate_assignment);
     REQUIRE(s_loggerCalled);
+#ifndef __cpp_lib_source_location
     // In C++17 these fields cannot be filled in so they are expected to be empty.
     REQUIRE(s_loggerArgs.lineNumber == 0);
     REQUIRE(s_loggerArgs.fileName == nullptr);
     REQUIRE(s_loggerArgs.functionName == nullptr);
+#else
+    // GCC can only compile these tests in C++20 so these fields will be filled in.
```
dmachaj, line 111: I don't quite understand these changes.  This test file is supposed to be compiled in C++17 mode so the source information should not be available.

alvinhochun: Using cppwinrt headers requires coroutine support to be available, and the experimental coroutine support from MSVC only works with... MSVC. For mingw-w64 targets (which use libc++ or libstdc++ depending on the toolchain) I am forced to compile everything in C++20; C++17 doesn't work.

test/test/custom_error.cpp
```diff
@@ -108,14 +108,9 @@ TEST_CASE("custom_error_logger")
...
-    // GCC can only compile these tests in C++20 so these fields will be filled in.
-    REQUIRE(s_loggerArgs.lineNumber == 103);
-    const auto fileNameSv = std::string_view(s_loggerArgs.fileName);
-    REQUIRE(!fileNameSv.empty());
-    REQUIRE(fileNameSv.find("custom_error.cpp") != std::string::npos);
-    const auto functionNameSv = std::string_view(s_loggerArgs.functionName);
-    REQUIRE(!functionNameSv.empty());
-    // Don't check the function name here. It is tested in `test_cpp20/custom_error.cpp`.
+    // GCC/Clang can only compile these tests in C++20 mode. If source_location
+    // is available these fields will be filled in. Don't do any cheks here
```
dmachaj, line 112: [nit] Spelling on "checks"

dmachaj:
FYI - I don't see any problems with these changes besides what you already addressed.  The details are beyond my expertise so I'm not going to officially sign off.  (I am not a contributor so my signoff wouldn't help much anyway).

strings/base_coroutine_threadpool.h
```diff
@@ -434,6 +447,21 @@ WINRT_EXPORT namespace winrt
                 m_handle(handle)
             {}
 
+#if defined(__GNUC__) && !defined(__clang__)
+            // HACK: GCC seems to require a move when calling operator co_await
+            // on the return value of resume_on_signal.
+            // This might be related to upstream bug:
+            // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99575
+            awaitable(awaitable &&other) noexcept :
```
kennykerr, line 455: Can't we use `= default` to force the compiler to generate this?

alvinhochun: IIRC the issue is that `std::atomic<T>` is unmovable (copy constructor is deleted).

kennykerr: Ah, that's too bad.

kennykerr (approved):

PR #1255, opened 2022-12-29
Cleaning up some warnings for Clang and GCC

alvinhochun:
Fixed or suppressed some warnings in the headers when using Clang and GCC. The last commit ("gcc: Fix -Wclass-memaccess warning") involved a small refactoring.

### DIFF
--- a/strings/base_array.h
+++ b/strings/base_array.h
@@ -345,6 +345,26 @@ WINRT_EXPORT namespace winrt
                 this->m_size = size;
             }
         }
+
+        std::pair<uint32_t, impl::arg_out<T>> detach_abi() noexcept
+        {
+#ifdef _MSC_VER
+            // https://github.com/microsoft/cppwinrt/pull/1165
+            std::pair<uint32_t, impl::arg_out<T>> result;
+            memset(&result, 0, sizeof(result));
+            result.first = this->size();
+            result.second = *reinterpret_cast<impl::arg_out<T>*>(this);
+            memset(this, 0, sizeof(com_array<T>));
+#else
+            std::pair<uint32_t, impl::arg_out<T>> result(this->size(), *reinterpret_cast<impl::arg_out<T>*>(this));
+            this->m_data = nullptr;
+            this->m_size = 0;
+#endif
+            return result;
+        }
+
+        template <typename U>
+        friend std::pair<uint32_t, impl::arg_out<U>> detach_abi(com_array<U>& object) noexcept;
     };
 
     template <typename C> com_array(uint32_t, C const&) -> com_array<std::decay_t<C>>;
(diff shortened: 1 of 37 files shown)


strings/base_array.h
```diff
@@ -345,6 +345,26 @@ WINRT_EXPORT namespace winrt
...
+        {
+#ifdef _MSC_VER
+            // https://github.com/microsoft/cppwinrt/pull/1165
+            std::pair<uint32_t, impl::arg_out<T>> result;
+            memset(&result, 0, sizeof(result));
+            result.first = this->size();
+            result.second = *reinterpret_cast<impl::arg_out<T>*>(this);
+            memset(this, 0, sizeof(com_array<T>));
+#else
+            std::pair<uint32_t, impl::arg_out<T>> result(this->size(), *reinterpret_cast<impl::arg_out<T>*>(this));
```
BenJKuhn, line 359: Doesn't this reintroduce the uninitialized padding data for non-VC compilers, per the comments in the PR linked above? What compiler warning does this resolve?

strings/base_array.h
```diff
@@ -418,14 +438,9 @@ WINRT_EXPORT namespace winrt
     }
 
     template <typename T>
-    auto detach_abi(com_array<T>& object) noexcept
+    std::pair<uint32_t, impl::arg_out<T>> detach_abi(com_array<T>& object) noexcept
```
BenJKuhn, line 441: Same comment as above. This change undoes an intentional change made to address a customer-reported issue.

strings/base_array.h
alvinhochun: The GCC warning is `'void* memset(void*, int, size_t)' clearing an object of type 'struct std::pair<unsigned int, float*>' with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]`.

This change does leave the padding uninitialized, but as said before it is "almost never going to be an issue for this function" and I am uncomfortable with the attempt to fix it potentially invoking UB. (Though I am no language lawyer so I cannot say for sure whether there is UB.)

strings/base_array.h
kennykerr: Do you recall the issue? We should have a test to avoid such a regression, but all tests pass on this PR.

kennykerr: Oh, you're referring to #1165 - that was itself to suppress a compiler warning. 🙄

kennykerr: Ideally, we can avoid the `memset` in #1165 and just suppress the code analysis warning but I don't know whether that's possible since its presumably not an actual compiler warning. Perhaps @DefaultRyan can chime in.

kennykerr: It looks to me like the original change was to suppress a warning in MSVC and this change is to suppress a warning in GCC. So if nobody objects, I'll go ahead and approve this change.

PR #1256, opened 2023-01-02
cmake: Allow using external winmd headers to bypass download

alvinhochun:
When integrating cppwinrt from another project, I want to vendor the winmd headers into the source tree to avoid performing the download during build. This change adds a `EXTERNAL_WINMD_INCLUDE_DIR` variable to enable this usage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,19 +175,29 @@ endif()
 
 # === winmd: External header-only library for reading winmd files ===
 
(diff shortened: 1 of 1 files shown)


sylveon:
This sounds ripe for trouble because of the pragma detect_mismatch included in the header. Every consumer is now locked in to this exact version of cppwinrt.

kennykerr (approved):

PR #1268, opened 2023-01-31
Move build pipeline into YAML

DefaultRyan:
Among other things, this will allow the C++/WinRT build pipeline to stay synchronized with the product code.

build.yaml contains the same build steps as the original classic pipeline. Here's a test run. https://microsoft.visualstudio.com/Dart/_build/results?buildId=63658741&view=results

### DIFF
--- a/.pipelines/build.yml
+++ b/.pipelines/build.yml
@@ -3,43 +3,48 @@
 #           MY_ACCESS_TOKEN: $(System.AccessToken)
 # Variable 'MajorVersion' was defined in the Variables tab
 # Variable 'MinorVersion' was defined in the Variables tab
-# Cron Schedules have been converted using UTC Time Zone and may need to be updated for your location
-# Agent Queue 'Azure Pipelines' was used with unrecognized Agent Specification, vmImage property must be specified to determine image - https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software
-# Multi-job configuration must be converted to matrix strategy: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-job-configuration
-# Agent Queue 'Azure Pipelines' was used with unrecognized Agent Specification, vmImage property must be specified to determine image - https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software
-# Multi-job configuration must be converted to matrix strategy: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-job-configuration
-# Agent Queue 'Azure Pipelines' was used with unrecognized Agent Specification, vmImage property must be specified to determine image - https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software
-# Multi-job configuration must be converted to matrix strategy: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-job-configuration
 trigger:
   branches:
     include:
     - refs/heads/master
   batch: True
 schedules:
-- cron: 0 2 * * *
+- cron: 0 18 * * *
   branches:
     include:
     - refs/heads/master
 name: $(MajorVersion).$(MinorVersion).$(date:yyMMdd)$(rev:.r)
 jobs:
-- job: Job_1
-  displayName: Build Binaries
+- job: BuildBinaries
   pool:
     name: Azure Pipelines
+    vmImage: 'windows-2022'
     demands:
     - msbuild
+  strategy:
+    matrix:
+      x86:
+        buildPlatform: 'x86'
+      x64:
+        buildPlatform: 'x64'
+      arm:
+        buildPlatform: 'arm'
+      arm64:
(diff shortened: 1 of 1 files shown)


kennykerr (dismissed):
Sweet!

kennykerr:
image

Weird that it seems to be "failing" because of a prior run:

image

Might just be some weird sync between GitHub and Azure.

.pipelines/build.yml
```diff
@@ -0,0 +1,653 @@
...
+    clean: true
+    persistCredentials: True
+
+  - task: NuGetToolInstaller@1
+    displayName: Use NuGet 6.0.2
+    continueOnError: True
+    inputs:
+      versionSpec: 6.0.2
+  - task: NuGetCommand@2
+
```
dmachaj, line 45: The line breaks here are strange.  There should be a blank line between tasks.  There should not be a line break between a task and its name.

.pipelines/build.yml
```diff
@@ -0,0 +1,653 @@
...
+      Contents: |
+        cppwinrt.exe
+        cppwinrt.pdb
+      TargetFolder: $(Build.ArtifactStagingDirectory)\cppwinrt
+
+  - task: CopyFiles@2
+    displayName: Stage Component cppwinrtvisualizer.*
+    condition: and(or(eq(variables['BuildPlatform'], 'x86'), eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'arm64')), eq(variables['BuildConfiguration'], 'release'), in(variables['Build.Reason'], 'Manual'))
+    inputs:
+      SourceFolder: '$(Build.SourcesDirectory)\natvis\$(BuildPlatform)\$(BuildConfiguration)\Component '
```
dmachaj, line 83: The sourceFolder string has an extra space at the end.  Also the quotes on it are mismatched compared to all other CopyFiles tasks that do not use quotes for the path.

.pipelines/build.yml
```diff
@@ -0,0 +1,653 @@
...
+      Contents: |
+        cppwinrtvisualizer.dll
+        cppwinrtvisualizer.pdb
+        cppwinrtvisualizer.vsdconfig
+      TargetFolder: $(Build.ArtifactStagingDirectory)\Component
+
+  - task: CopyFiles@2
+    displayName: Stage Standalone cppwinrtvisualizer.*
+    condition: and(or(eq(variables['BuildPlatform'], 'x86'), eq(variables['BuildPlatform'], 'x64'), eq(variables['BuildPlatform'], 'arm64')), eq(variables['BuildConfiguration'], 'release'), in(variables['Build.Reason'], 'Manual'))
```
dmachaj, line 92: IMO this condition is super difficult to read.  Can you compute it once at the top and store it in a variable/parameter with a reasonable name?

.pipelines/build.yml
```diff
@@ -0,0 +1,653 @@
...
+      artifactName: $(BuildConfiguration)_$(BuildPlatform)
+      targetPath: $(Build.ArtifactStagingDirectory)
+
+  - task: PublishSymbols@2
+    displayName: Publish symbols
+    condition: and(eq(variables['BuildConfiguration'], 'release'), in(variables['Build.Reason'], 'Manual'))
+    inputs:
+      SymbolsFolder: $(Build.ArtifactStagingDirectory)
+      SearchPattern: '**/*.pdb'
+      IndexSources: false
```
dmachaj, line 128: Why not index the source?

.pipelines/build.yml
```diff
@@ -0,0 +1,653 @@
...
+    inputs:
+      branchVersionExcludeBranch: master
+      disableWorkspace: true
+      disableMsbuildVersion: true
+      disableBuildTools: true
+
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core runtime 2.1.x (required for ESRP signing)'
+    inputs:
+      version: 2.1.x
```
dmachaj, line 155: This step seems to be why publish vpacks is failing.  .NET 3 or less is out of support and is blocked now.

.pipelines/build.yml
DefaultRyan: > The sourceFolder string has an extra space at the end. Also the quotes on it are mismatched compared to all other CopyFiles tasks that do not use quotes for the path.

Good catch. There was an extra trailing space in the classic pipeline as well. I suppose the YAML exported tried to preserve it and put the whole thing in quotes.

.pipelines/build.yml
DefaultRyan: Seems we're in a bit of a pickle. The ESRP signing task has been failing for weeks because it requires .NET Core 2.1, or at least that's what it looked like to me:

https://microsoft.visualstudio.com/Dart/_build/results?buildId=62315308&view=logs&j=33e667ca-6ac0-5edd-a669-2247a07fa473&t=53936be5-a3bb-56e6-88f6-8ad4a2fccc05

dmachaj: Can use add a second `UseDotNet` task after signing and before publishing to move to a more recent version?

DefaultRyan: Updating from EsrpCodeSigning@1 to EsrpCodeSigning@2 fixed the issue. No longer depend on .NET Core 2.1.x.

dmachaj (dismissed):

dmachaj (approved):

DefaultRyan:
The microsoft.cppwinrt build is hitting some weird NuGet timeout issues. Not related to this change.

PR #1274, opened 2023-02-14
Add a clang-specific impl->projection conversion operator

### LINKED ISSUE #1273: Bug: clang - impl type cannot convert to projected type
### Version

2.0.230207.1 and earlier (tested 2.0.210825.3)

### Summary

_No response_

### Reproducible example

```c++
#include <winrt/Windows.Foundation.h>

struct Stringy : public winrt::implements<Stringy, winrt::Windows::Foundation::IStringable> {
	Stringy() {}
	winrt::hstring ToString() {
		return L"FooBarBaz";
	}
};

int main()
{
	winrt::init_apartment();
	auto x = winrt::make_self<Stringy>();
	winrt::Windows::Foundation::IStringable broken{ *x };
	printf("Hello, %ls!\n", broken.ToString().c_str());
}
```

### Expected behavior

`works`, works.

### Actual behavior

Something is preventing the following conversion from taking place:

```
Stringy
 +--> producer_convert::operator producer_ref<IStringable>
       +--> IStringable
```

resulting in the following chain of error messages

```
1>main.cpp(18,50): error : no viable conversion from 'Stringy' to 'winrt::Windows::Foundation::IStringable'
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\impl\Windows.Foundation.1.h(170,35): message : candidate constructor (the implicit copy constructor) not viable: no known conversion from 'Stringy' to 'const winrt::Windows::Foundation::IStringable &' for 1st argument
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\impl\Windows.Foundation.1.h(170,35): message : candidate constructor (the implicit move constructor) not viable: no known conversion from 'Stringy' to 'winrt::Windows::Foundation::IStringable &&' for 1st argument
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\impl\Windows.Foundation.1.h(174,9): message : candidate constructor not viable: no known conversion from 'Stringy' to 'std::nullptr_t' for 1st argument
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\base.h(8095,9): message : candidate function
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\base.h(6749,9): message : candidate function
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\base.h(6754,9): message : candidate function
1>C:\src\ClangCppWinRTTest1\x64\Debug\Generated Files\winrt\impl\Windows.Foundation.1.h(170,35): message : passing argument to parameter here
```

### Additional comments

If you write out the (verbose, impl types) conversion from Stringy to `producer_ref<IStringable>` explicitly, it works properly with no warning/error; n.b. the `static_cast` is using `operator producer_ref<I>`.

```
winrt::Windows::Foundation::IStringable works{
    static_cast<winrt::impl::producer_ref<winrt::Windows::Foundation::IStringable>>(*x)
};
```

DHowett:
In Clang, the conversion from `producer_convert` to `const producer_ref<I>`
does not allow for the move-initialization of `I` via `I(I&&)` (implicit
or explicit) due to a language misspecification¹.

In brief², we cannot offer zero-cost initialization of projected types
from implementation types _in all meaningful cases_ in projects compiled
with clang.

Until it improves (which is tracked in llvm/llvm-project#17056), this
conversion operator will allow for C++/WinRT code to compile with Clang.

Closes #1273

¹ http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#2077 documents
the language standard bug that results in this.

² In non-brief, https://stackoverflow.com/a/70654432

### DIFF
--- a/test/old_tests/UnitTests/as_implements.cpp
+++ b/test/old_tests/UnitTests/as_implements.cpp
@@ -74,16 +74,6 @@ TEST_CASE("as_implements")
     REQUIRE(foo.get() == foo2.get());
 }
 
-TEST_CASE("as_implements_uniform_initialization")
-{
-    // uniform initialization relies on IStringable(IStringable&&),
-    // which requires non-const rvalue semantics.
-    com_ptr<Foo> foo = make_self<Foo>();
-    IStringable stringable{ foo.as<IStringable>() };
-    com_ptr<Foo> foo2 = stringable.as<Foo>();
-    REQUIRE(foo.get() == foo2.get());
-}
-
 TEST_CASE("as_implements_inheritance")
 {
     com_ptr<Bar> bar = make_self<Bar>();
@@ -147,4 +137,14 @@ TEST_CASE("as_implements_inheritance")
         com_ptr<Foo> foo2 = foo->get_strong_foo();
         REQUIRE(bar.get() == foo2.get());
     }
+}
+
+TEST_CASE("convert_to_implements_via_uniform_initialization")
+{
+    // uniform initialization relies on IStringable(IStringable&&),
+    // which requires non-const rvalue semantics.
+    com_ptr<Foo> foo = make_self<Foo>();
+    IStringable stringable{ *foo };
+    com_ptr<Foo> foo2 = stringable.as<Foo>();
+    REQUIRE(foo.get() == foo2.get());
 }
--- a/test/old_tests/UnitTests/as_implements.cpp
+++ b/test/old_tests/UnitTests/as_implements.cpp
@@ -74,6 +74,16 @@ TEST_CASE("as_implements")
     REQUIRE(foo.get() == foo2.get());
 }
 
+TEST_CASE("as_implements_uniform_initialization")
+{
+    // uniform initialization relies on IStringable(IStringable&&),
+    // which requires non-const rvalue semantics.
+    com_ptr<Foo> foo = make_self<Foo>();
+    IStringable stringable{ foo.as<IStringable>() };
+    com_ptr<Foo> foo2 = stringable.as<Foo>();
+    REQUIRE(foo.get() == foo2.get());
+}
+
 TEST_CASE("as_implements_inheritance")
 {
     com_ptr<Bar> bar = make_self<Bar>();
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -110,12 +110,27 @@ namespace winrt::impl
     template <typename D, typename I, typename Enable>
     struct producer_convert : producer<D, typename default_interface<I>::type>
     {
-        operator producer_ref<I>() const noexcept
+#ifdef __clang__
+        // This is sub-optimal in that it requires an AddRef and Release of the
+        // implementation type for every conversion, but it works around an
+        // issue where Clang ignores the conversion of producer_ref<I> const
+        // to I&& (an rvalue ref that cannot bind a const rvalue).
+        // See CWG rev. 110 active issue 2077, "Overload resolution and invalid
+        // rvalue-reference initialization"
+        operator I() const noexcept
+        {
+            I result{ nullptr };
+            copy_from_abi(result, (produce<D, typename default_interface<I>::type>*)this);
+            return result;
+        }
+#else
+        operator producer_ref<I> const() const noexcept
         {
             return { (produce<D, typename default_interface<I>::type>*)this };
         }
+#endif
 
-        operator producer_vtable<I>() const noexcept
+        operator producer_vtable<I> const() const noexcept
         {
             return { (void*)this };
         }
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -110,12 +110,12 @@ namespace winrt::impl
     template <typename D, typename I, typename Enable>
     struct producer_convert : producer<D, typename default_interface<I>::type>
     {
-        operator producer_ref<I> const() const noexcept
+        operator producer_ref<I>() const noexcept
         {
             return { (produce<D, typename default_interface<I>::type>*)this };
         }
 
-        operator producer_vtable<I> const() const noexcept
+        operator producer_vtable<I>() const noexcept
         {
             return { (void*)this };
         }


DHowett:
(I'll dig in on the ASAN failures, which didn't reproduce when I ran the test suite locally.)

lhecker:
I've been closely following this issue (since I'm a huge fan of clang) and was quite surprised about these build failures. They happen locally for me as well - here's `test_cpp20` in x64/Debug:
```
C:\Users\lhecker\projects\cppwinrt\test\test_cpp20\await_completed.cpp(49): FAILED:
  REQUIRE( consumed == 0 )
with expansion:
  3296 (0xce0) == 0

C:\Users\lhecker\projects\cppwinrt\test\test_cpp20\await_completed.cpp(49): FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  Abnormal termination

===============================================================================
test cases: 2 | 1 passed | 1 failed
assertions: 3 | 1 passed | 2 failed
```

If I revert 4363e5c37a4790c128eebb9261c71e672cec5dc6 it looks much better:
```
C:\Users\lhecker\projects\cppwinrt\test\test_cpp20\custom_error.cpp(59): FAILED:
  REQUIRE( s_loggerArgs.lineNumber == 15 )
with expansion:
  0 == 15

===============================================================================
test cases:  5 |  4 passed | 1 failed
assertions: 19 | 18 passed | 1 failed
```

Anyone knows what's going on?

DHowett:
Well, it looks like that is a load-bearing `const`. Thanks for the explanation, @lhecker!

In short, here's what happens during conversion from `implements<D, I>` to `I`.

### With `const producer_ref<I>`

1. `producer_convert::operator producer_ref<I> const` calls ctor...
2. ... `producer_ref<I>::producer_ref` which calls base type ctor...
3. `I::I(void*, take_ownership_from_abi_t)`
4. That `producer_ref` is used in an expr where an `I` is accepted.
5. _The resultant `I` is **copy-constructed**, increasing the reference count, because the source `I` (nee `producer_ref`) is `const`._
6. The `producer_ref<I>` destructs, nulling out its `ptr` member.

### With `producer_ref<I>`

Same as above, except step 5 differs:

5. _The resultant `I` is **move-constructed**, which does not increase the reference count._ This produces a projected type that does not hold a strong reference to its member.

### OK?

It seems like `producer_ref` exists to allow you to pass an implementation type off as a projected type without an `AddRef`/`Release` in the common parameter-passing case where something consumes a `const I&`.

Is `AddRef` expensive enough to justify that? This code doesn't work in spec-compliant compilers... which I thought was C++/WinRT's main goal.

kennykerr:
Yes, the `producer_ref` conversion exists to support `const&` call sites from within an implementation. Injecting `AddRef`/`Release` bumps would be a deal breaker. Whether Clang is more compliant than MSVC is debatable. Anyway, you could try adding an overload that is non-`const` without actually removing the existing conversion. Not sure if that will have other negative consequences though.

DHowett:
My next couple avenues of exploration are...

* Adding an `operator producer_ref<I>&&` to enable explicit rvalue-ref conversion
* Adding a non-`const` overload to see if it gets picked up in the appropriate places
* Adding an `operator I` that *does* result in an `AddRef`/`Release` but scoping it down to only clang

I know that the last one would be a dealbreaker, but would you be amenable to it if it didn't impact the golden path MSVC scenario?

kennykerr:
Presumably we don't want to inject spurious ref count bumps into Clang builds either but if it only impacts cases where a ref count is needed anyway, such as `emplace_back`/`push_back` in your examples, then that should be fine.

DHowett:
Okay, so, here's my contention. I've dug and dug and I can't find a way to eliminate the `AddRef`/`Release` when doing a converting copy from impl to projection with Clang.

I'd like to propose two things:

1. We offer an `operator I()` for use _only with Clang_. It imposes an extra AddRef/Release.
2. That operator is gated behind a new define, (named similar to) `CPPWINRT_I_ACCEPT_THE_COST_SO_BUILD_WITH_CLANG`

See, at the end of the day... _all I need is for my code to appear to be well-formed and compile._ I'm not shipping a product based on this, but right now we would need to fix up ~151 sites across our codebase (and annotate them with a comment explaining why the normal C++/WinRT way doesn't work so they don't get reverted) and in every one of those cases introduce our own extra `AddRef`/`Release`.

If you'd be amenable to a gated "make the code work, sub-optimally, **if the user opts in**" experience... I'd be happy to write it up.

kennykerr:
Hey, thanks for looking into it. Obviously, compiling sub-optimally is better than not compiling at all. Can we just hide it behind an `#ifdef clang`? That way we can always improve things later as smart developers help out. I just have a strong aversion to adding preprocessor definitions as they're hard to manage and version over time.

DHowett:
I'm totally cool with that. Thanks @kennykerr. I'll update this PR inplace.

DHowett:
Alright, updated the PR title, body and contents.

kennykerr (approved):
Looks good - thanks!

alvinhochun:
Can you add some tests for this?

kennykerr:
A little surprised this isn't already covered, unless a test is disabled for clang?

DHowett:
Good call. I added a test, unconditional for all compilers. :smile:

I could not find one that had been disabled for clang that would have exercised this.

test/old_tests/UnitTests/as_implements.cpp
```diff
@@ -74,6 +74,16 @@ TEST_CASE("as_implements")
     REQUIRE(foo.get() == foo2.get());
 }
 
+TEST_CASE("as_implements_uniform_initialization")
+{
+    // uniform initialization relies on IStringable(IStringable&&),
+    // which requires non-const rvalue semantics.
+    com_ptr<Foo> foo = make_self<Foo>();
+    IStringable stringable{ foo.as<IStringable>() };
```
kennykerr, line 82:
```suggestion
    IStringable stringable{ *foo };
```

Isn't something like this required to actually invoke the implicit conversion operator? Otherwise you're just calling the explicit `as` method provided by `com_ptr` which in turn calls QI.

DHowett: *facepalm*

kennykerr (approved):

PR #1276, opened 2023-02-16
Reduce stack consumption if unable to switch to apartment_context

oldnewthing:
The old code would invoke the handle recursively. The new code returns `false` from `await_suspend`, allowing the calling coroutine to resume with no stack consumption. Use `[[nodiscard]]` to ensure I updated all callers, and to avoid future errors with new callers.

Also remove `#ifdef`s to work around code generation issues in MSVC versions less than 16.11. If you are on a version that old, you should stick with an older version of C++/WinRT.

### DIFF
--- a/strings/base_coroutine_threadpool.h
+++ b/strings/base_coroutine_threadpool.h
@@ -77,7 +77,7 @@ namespace winrt::impl
         return 0;
     };
 
-    inline void resume_apartment_sync(com_ptr<IContextCallback> const& context, coroutine_handle<> handle, int32_t* failure)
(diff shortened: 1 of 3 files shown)


strings/base_coroutine_threadpool.h
```diff
@@ -77,7 +77,7 @@ namespace winrt::impl
         return 0;
     };
 
-    inline void resume_apartment_sync(com_ptr<IContextCallback> const& context, coroutine_handle<> handle, int32_t* failure)
+    [[nodiscard]] inline bool resume_apartment_sync(com_ptr<IContextCallback> const& context, coroutine_handle<> handle, int32_t* failure)
```
kennykerr, line 80: I do love `nodiscard`. 👍

kennykerr (approved):

PR #1277, opened 2023-02-17
Fix unreliable clock epoch tests

oldnewthing:
The clock epoch tests are sensitive to CPU speed and the precise moment the clock ticks over from one second to another. Change the way we do epoch tests so that they are not affected by either factor.

This fixes a spurious CI test failure where the test environment was 3 milliseconds too slow.

https://github.com/oldnewthing/cppwinrt/actions/runs/4195918696/jobs/7276654213

### DIFF
--- a/test/old_tests/UnitTests/clock.cpp
+++ b/test/old_tests/UnitTests/clock.cpp
@@ -17,6 +17,18 @@ namespace winrt
     }
 }
 
+// To confirm that two clocks have the same epoch, we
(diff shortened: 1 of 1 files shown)


kennykerr (approved):
Thanks for fixing this annoyance!

kennykerr:
A few more unreliable tests... https://github.com/microsoft/cppwinrt/actions/runs/4208012719 🙄

PR #1290, opened 2023-03-23
Expose configuring /nomidl.

jlaanstra:
This change makes it easier to generate proxy stubs by allowing modern idl to generate proxy stubs code. Midl supports this and so it seems like a good idea for cppwinrt to expose it.

/cc @Scottj1s

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.props
+++ b/nuget/Microsoft.Windows.CppWinRT.props
@@ -43,7 +43,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <InterfaceIdentifierFileName Condition="'%(Midl.InterfaceIdentifierFileName)'==''">nul</InterfaceIdentifierFileName>
             <ProxyFileName Condition="'%(Midl.ProxyFileName)'==''">nul</ProxyFileName>
             <TypeLibraryName Condition="'%(Midl.TypeLibraryName)'==''"></TypeLibraryName>
-            <NoMidl>true</NoMidl>
+            <NoMidl Condition="'%(Midl.NoMidl)'==''">true</NoMidl>
         </Midl>
         <ProjectReference Condition="'$(XamlLanguage)' != 'C++' and '$(CppWinRTEnableDefaultCopyLocalFalse)' == 'true'">
             <!-- By default, for a C++/WinRT project,
--- a/nuget/Microsoft.Windows.CppWinRT.props
+++ b/nuget/Microsoft.Windows.CppWinRT.props
@@ -43,6 +43,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <InterfaceIdentifierFileName Condition="'%(Midl.InterfaceIdentifierFileName)'==''">nul</InterfaceIdentifierFileName>
             <ProxyFileName Condition="'%(Midl.ProxyFileName)'==''">nul</ProxyFileName>
             <TypeLibraryName Condition="'%(Midl.TypeLibraryName)'==''"></TypeLibraryName>
+            <NoMidl>true</NoMidl>
         </Midl>
         <ProjectReference Condition="'$(XamlLanguage)' != 'C++' and '$(CppWinRTEnableDefaultCopyLocalFalse)' == 'true'">
             <!-- By default, for a C++/WinRT project,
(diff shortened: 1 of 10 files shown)


nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -475,8 +474,9 @@ $(XamlMetaDataProviderPch)
             <_MidlReferences Include="@(CppWinRTPlatformWinMDReferences)"/>
             <_MidlReferencesDistinct Remove="@(_MidlReferencesDistinct)" />
             <_MidlReferencesDistinct Include="@(_MidlReferences->'%(WinMDPath)'->Distinct())" />
-            <Midl Condition="'%(Midl.DisableReferences)'==''">
-                <AdditionalOptions>%(Midl.AdditionalOptions) %40"$(CppWinRTMidlResponseFile)"</AdditionalOptions>
+            <Midl>
+                <AdditionalOptions Condition="'%(Midl.NoMidl)'!='false'">%(Midl.AdditionalOptions) /nomidl</AdditionalOptions>
+                <AdditionalOptions Condition="'%(Midl.DisableReferences)'!='false' or '%(Midl.NoMidl)'!='false'">%(Midl.AdditionalOptions) /nomidl %40"$(CppWinRTMidlResponseFile)"</AdditionalOptions>
```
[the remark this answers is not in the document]
jlaanstra: When I wrote this it looked like MSBuild skipped evaluating the second AdditionalOptions line if the first one evaluated to true. /cc @rainersigwald

nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -463,7 +463,6 @@ $(XamlMetaDataProviderPch)
 
     <!--Insert Midl /references to Platform WinMDs, library reference WinMDs, and direct reference WinMDs-->
     <Target Name="CppWinRTSetMidlReferences"
-            Condition="'$(CppWinRTModernIDL)' != 'false'"
```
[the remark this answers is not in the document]
jlaanstra: Let me check my reasoning for removing this.

nuget/Microsoft.Windows.CppWinRT.props
```diff
@@ -43,6 +43,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             <InterfaceIdentifierFileName Condition="'%(Midl.InterfaceIdentifierFileName)'==''">nul</InterfaceIdentifierFileName>
             <ProxyFileName Condition="'%(Midl.ProxyFileName)'==''">nul</ProxyFileName>
             <TypeLibraryName Condition="'%(Midl.TypeLibraryName)'==''"></TypeLibraryName>
+            <NoMidl>true</NoMidl>
```
dmachaj, line 46: This name feels a bit too generic.  You are still calling `midl.exe` but are adding/removing the `/nomidl` option to it.  This _might_ be misinterpreted as skipping the call to `midl.exe` entirely.  Consider using a more specific name such as `<PassNoMidlOption>` (that one isn't great but hopefully you get the idea)

sylveon: Yeah, I'm still confused at what `/nomidl` does to begin with.

jlaanstra: I don't disagree but the /nomidl option name itself already makes little sense to me. Naming it this way at least enables easy matching to command line options.

jlaanstra: Actually according to the docs /nomidl does exactly what it says. It doesn't run midl.exe, see https://learn.microsoft.com/en-us/windows/win32/midl/-nomidl

kennykerr: Keep in mind that `MIDL.EXE` and `MIDLRT.EXE` are two very different things. The `/nomidl` makes sense if you think about it as a `MIDLRT` option that determines whether it will delegate to `MIDL` for further processing. The internals are a little muddied, since `MIDL` can also delegate to `MIDLRT`, but that's the idea at least. If all you need is a .winmd then `/nomidl` is what you want since that's something that `MIDLRT` can handle without `MIDL` getting involved, but for anything beyond the .winmd you probably want to run `MIDL` to handle the rest. 🤷‍♂️

jlaanstra:
Test failure seems unrelated?

Issue #668, opened 2020-06-24
Version information shouldn't require hand-parsing


jonwis:
Per https://task.ms/26741093 - `CPPWINRT_VERSION` is a string. Could it also be a four-part structure?

```c++
struct cppwinrt_version_t {
    uint16_t major;
    uint16_t minor;
    uint16_t release;
    uint16_t build;
};
constexpr cppwinrt_version_t get_cppwinrt_version() {
    return { 2, 0, 20609, 3 };
}
```

Scottj1s:
Note that this versioning scheme isn't SemVer-compliant.  You may want to consider restructuring it before locking it in further.

kennykerr:
Version APIs are hard. 😉 

WIL uses this to light up feature support depending on the version of C++/WinRT it is being included with. The trouble with adding a new function is that this function won't exist in older versions and thus prevent WIL from using it.

Feel free to reopen if there's interest in exploring this further.

PR #1293, opened 2023-03-29
Use safe DLL loading (avoid current directory)

oldnewthing:
This avoids DLL planting issues during the fallback search for a base trust server. It uses the `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` flag, which searches the usual places, including the application directory, but excludes the current directory. The flag is supported on Win8 onward, as well as on Win7 if you have KB2533623 installed (which you really should). If you need to support versions of Win7 that haven't been patched since 2011, then stick with whatever version of C++/WinRT you are currently using.

### DIFF
--- a/strings/base_activation.h
+++ b/strings/base_activation.h
@@ -39,7 +39,7 @@ namespace winrt::impl
 
         if (hr == impl::error_not_initialized)
         {
-            auto usage = reinterpret_cast<int32_t(__stdcall*)(void** cookie) noexcept>(WINRT_IMPL_GetProcAddress(WINRT_IMPL_LoadLibraryW(L"combase.dll"), "CoIncrementMTAUsage"));
+            auto usage = reinterpret_cast<int32_t(__stdcall*)(void** cookie) noexcept>(WINRT_IMPL_GetProcAddress(load_library(L"combase.dll"), "CoIncrementMTAUsage"));
 
             if (!usage)
             {
@@ -66,7 +66,7 @@ namespace winrt::impl
         {
             path.resize(count);
             path += L".dll";
-            library_handle library(WINRT_IMPL_LoadLibraryW(path.c_str()));
+            library_handle library(load_library(path.c_str()));
             path.resize(path.size() - 4);
 
             if (!library)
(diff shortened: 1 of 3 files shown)


kennykerr (approved):

dmachaj (approved):

jlaanstra:
Do we want to allow customization of the default, maybe via a #define?

jlaanstra (approved):

sylveon:
I'm not sure it's worth the expense of adding a customization switch to support operating systems that haven't been patched since 2011

kennykerr:
Right, the switch is to use a  different version of cppwinrt.

jonthysell:
For RNW we're going to do this:

```cpp
#ifdef CPPWINRT_USE_LOADLIBRARYEXW
#define WINRT_IMPL_LoadLibraryW(name) \
  WINRT_IMPL_LoadLibraryExW(name, nullptr, 0x00001000 /* LOAD_LIBRARY_SEARCH_DEFAULT_DIRS */)
#endif
```

where `CPPWINRT_USE_LOADLIBRARYEXW` is defined via a MSBUILD version check:

```xml
  <ItemDefinitionGroup Label="CppWinRT">
    <ClCompile>
      <PreprocessorDefinitions Condition="$([MSBuild]::VersionGreaterThanOrEquals('$(CppWinRTVersion)', '2.0.230524.3'))">
        CPPWINRT_USE_LOADLIBRARYEXW;
        %(PreprocessorDefinitions)
      </PreprocessorDefinitions>
    </ClCompile>
  </ItemDefinitionGroup>
```

because of the lack of convenient version macros to do it strictly in code: #668.

oldnewthing:
@jonthysell Note that all `WINRT_IMPL_*` names are [reserved for internal use by C++/WinRT](https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/naming) and are considered off-limits to apps. 

> In all namespaces, names beginning with WINRT_IMPL_ are reserved for C++/WinRT, and you shouldn't use them in your application.

If you want to call LoadLibrary, you can define your own `RNW_LoadLibraryW` function.

PR #1294, opened 2023-03-31
Fix flaky clock and line-number tests

oldnewthing:
Clock test needs to take double-rounding of time_t into account.

Fix fragility in line-number test by using `#line` directive.

### DIFF
--- a/test/test_cpp20/custom_error.cpp
+++ b/test/test_cpp20/custom_error.cpp
@@ -7,11 +7,10 @@ namespace
 {
     static bool s_loggerCalled = false;
 
-    // Note that we are checking that the source line number matches expectations.  If lines above this are changed
-    // then this value needs to change as well.
     void FailOnLine15()
     {
         // Validate that handler translated exception
+#line 15 // Force next line to be reported as line number 15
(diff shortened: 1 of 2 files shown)


test/test_cpp20/custom_error.cpp
```diff
@@ -7,11 +7,10 @@ namespace
 {
     static bool s_loggerCalled = false;
 
-    // Note that we are checking that the source line number matches expectations.  If lines above this are changed
-    // then this value needs to change as well.
     void FailOnLine15()
     {
         // Validate that handler translated exception
+#line 15 // Force next line to be reported as line number 15
```
dmachaj, line 13: How was this flaky?  Line 15 should be line 15 unless people are changing this code.

oldnewthing: It's not flaky, but it's fragile.

kennykerr (approved):

test/test_cpp20/custom_error.cpp
dmachaj: Ah, that makes sense.

kennykerr: The question is it flaky or is it flakey. 🤔

Issue #1299, opened 2023-04-24
Bug: winrt::capture needs detector for "REFIID, IUnknown**" methods


jonwis:
### Summary

Trying to call https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ms718237(v=vs.85) (IGetRow::GetRowFromHROW) in my app using `winrt::capture`. The signature is `HRESULT GetRowFromHROW(IUnknown **pUnkOuter, HROW hRow, REFIID riid, IUnknown **ppUnk)`.

Capture requires the capture output parameter to be `void**`, so this fails.

### Reproducible example

```rust
#include <windows.h>
#include <propsys.h>
#include <winrt/Windows.Foundation.h>

HRESULT GetRowFromHROW(
    ::IUnknown** pUnkOuter,
    int hRow,
    REFIID riid,
    ::IUnknown** ppUnk);

int main()
{
    auto r = winrt::capture<::IPropertyStore>(GetRowFromHROW, 3, nullptr);
    r->Commit();

    return 0;
}
```

### Expected behavior

Expected `capture` to adapt to three common COM capturable parameter pairs - `(REFIID, void**)`, `(REFIID, ::IUnknown**)`, `(REFIID, IInspectable**)` ... or maybe more generally "void** and things-that-derive-from-IUnknown**".

### Actual behavior

1>c:\source\repos\App1\CppWinrtTest\Debug\Generated Files\winrt\base.h(2330,48): error C2664: 'HRESULT (IUnknown **,int,const IID &,IUnknown **)': cannot convert argument 4 from 'void **' to 'IUnknown **'
1>c:\source\repos\App1\CppWinrtTest\Debug\Generated Files\winrt\base.h(2330,48): message : Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or parenthesized function-style cast
1>c:\source\repos\App1\CppWinrtTest\Debug\Generated Files\winrt\base.h(2567): message : see reference to function template instantiation 'int32_t winrt::impl::capture_to<IPropertyStore,HRESULT(__cdecl *)(IUnknown **,int,const IID &,IUnknown **),_Ty,nullptr>(void **,F,_Ty &&,nullptr &&)' being compiled
1>        with
1>        [
1>            _Ty=int,
1>            F=HRESULT (__cdecl *)(IUnknown **,int,const IID &,IUnknown **)
1>        ]

### Additional comments

[...]

PR #1301, opened 2023-04-28
Add `capture` support for unconventional result types

kennykerr:
Fixes: #1299


jonwis (dismissed):

strings/base_com_ptr.h
```diff
@@ -7,16 +7,27 @@ WINRT_EXPORT namespace winrt
 
 namespace winrt::impl
 {
+    struct capture_decay
+    {
+        void** result;
+
+        template <typename T>
+        operator T** ()
```
sylveon, line 15: Maybe check if `T` derives from IUnknown?

test/old_tests/UnitTests/capture.cpp
```diff
@@ -41,6 +48,12 @@ HRESULT __stdcall CreateCapture(int value, GUID const& iid, void** object) noexc
     return capture->QueryInterface(iid, object);
 }
 
+HRESULT __stdcall CreateCapture2(int value, GUID const& iid, ::IUnknown** object) noexcept
```
sylveon, line 51: IInspectable was mentionned in the original issue. Sounds fair to test it too.

strings/base_com_ptr.h
kennykerr: In most cases `T` will be `void` - the whole point here is to throw away type safety and just take whatever is provided, as long as it's a `T**`.

oldnewthing:
I was originally worried that this might be too permissive. There are some methods that take a REFIID followed by something that *isn't* the matched output pointer. However, they are exceedingly rare, and the only one that does it at the end of the parameter list that I could find (mind you, I gave up halfway through the D's) is IDebugProperty::EnumMembers.

Unit test needs to include the forwarder pattern:

```cpp
com_ptr<ICapture> f = capture<ICapture>([&](auto riid, auto ppv) { return CreateCapture2(10, riid, ppv); });
```

This pattern is used to accommodate methods that don't put the riid/ppv as the last two parameters.

```cpp
HRESULT WeirdMethod(REFIID riid, char const* interloper, IUnknown** result, int straggler);

e.capture<IUnknown>(WeirdMethod, [](auto interloper, auto straggler, riid, auto result) { return WeirdMethod(riid, interloper, result, straggler); }, "interloper", 10);
```

PR #1304, opened 2023-05-15
Remove explicitly setting PreferredToolArchitecture, since VS 2022 handles this more comprehensively

### LINKED ISSUE #1302: Bug: PreferredToolArchitecture x64 should default, not override, conditionally for x86
### Version

_No response_

### Summary

PreferredToolArchitecture is currently set to x64 to address out of memory issues due to large PCH files, when targeting x86.  Using the x64 toolchain to cross compile for x86 allows much larger memory use and reduces these OOMs.  This is a problem for ARM/ARM64 builds, causing underbuilds.  These architectures should use their own native toolchains.  PreferredToolArchitecture should be set to x64 for x86 only as a conditional default.

### Reproducible example

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional comments

_No response_

Scottj1s:
fixes #1302

With VS 2022, PreferredToolArchitecture is now set appropriately for every architecture (via Microsoft.Cpp.ToolsetLocation.props), so should no longer be set unconditionally by C++/WinRT

### DIFF
--- a/nuget/Microsoft.Windows.CppWinRT.props
+++ b/nuget/Microsoft.Windows.CppWinRT.props
@@ -13,7 +13,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <PropertyGroup>
-        <PreferredToolArchitecture>x64</PreferredToolArchitecture>
         <CanReferenceWinRT>true</CanReferenceWinRT>
         <CppWinRTPackage Condition="'$(CppWinRTEnabled)' != 'true'">true</CppWinRTPackage>
         <CppWinRTPackage Condition="'$(CppWinRTPackage)' != 'true'">false</CppWinRTPackage>


nuget/Microsoft.Windows.CppWinRT.props
```diff
@@ -13,7 +13,6 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </PropertyGroup>
 
     <PropertyGroup>
-        <PreferredToolArchitecture>x64</PreferredToolArchitecture>
```
dmachaj, line 16: Should this be scoped to the VS2022 build tools (so that VS2019 still gets it set)?

dmachaj: Or the opposite; only set it for VS2019 and then VS2022+ do not set it because VS does the right thing already.

kennykerr (approved):

dmachaj (approved):

nuget/Microsoft.Windows.CppWinRT.props
Scottj1s: It's actually a bug to set it in VS2019 when building for ARM64 (reported elsewhere).  And since it's just a build optimization, I think better to remove it altogether, than to effectively copy the intelligence in the VS 2022 Microsoft.Cpp.ToolsetLocation.props.

Issue #1306, opened 2023-05-17
Bug: cppwinrt does not support chinese path


HO-COOH:
### Version

2.0.230225.1

### Summary

As title suggest

### Reproducible example

```rust
1. On a chinese version of windows, create a username that has Chinese characters.
2. Create a C++ UWP project, update cppwinrt nuget if necessary.
3. Compile.
```

### Expected behavior

Compiles fine

### Actual behavior

```
已启动生成...
1>------ 已启动生成: 项目: BlankApp1, 配置: Debug x64 ------
1>cppwinrt : error Path 'C:\Users\涓枃鐢ㄦ埛\source\repos\BlankApp1\x64\Debug\Merged\BlankApp1.winmd' is not a file or directory
1>C:\Users\中文用户\source\repos\BlankApp1\packages\Microsoft.Windows.CppWinRT.2.0.230225.1\build\native\Microsoft.Windows.CppWinRT.targets(848,9): error MSB3073: 命令“"C:\Users\中文用户\source\repos\BlankApp1\packages\Microsoft.Windows.CppWinRT.2.0.230225.1\build\native\..\..\bin\"cppwinrt @"x64\Debug\BlankApp1.vcxproj.cppwinrt_comp.rsp"”已退出，代码为 1。
1>已完成生成项目“BlankApp1.vcxproj”的操作 - 失败。
========== 生成: 0 成功，1 失败，0 最新，0 已跳过 ==========
========= 生成 开始于 11:40 AM，并花费了 00.643 秒 ==========

```

![image](https://github.com/microsoft/cppwinrt/assets/42881734/b5731feb-fe23-4ab3-b31f-489240279e6d)

### Additional comments

Windows 10 1809

Does NOT enable the `beta UTF-8` support in language settings, because normal user won't have this enabled.

kennykerr:
Thanks for reporting! Fix is here: #1307

driver1998:
Using UTF-8 codepage manifest won't help OP here though, since it is supported in v1903+.

walbourn:
Officially Windows 10, Version 1809 (Build 17763) has been out of the support lifecycle for 2-3 years.

https://learn.microsoft.com/en-us/lifecycle/products/windows-10-home-and-pro

https://learn.microsoft.com/en-us/lifecycle/products/windows-10-enterprise-and-education

Windows Server 2019 is Build 17763 so that's really the only currently supported scenario for it. Mainstream support ends in ~7 months.

[...]

PR #1307, opened 2023-05-17
Add UTF-8 path support

kennykerr:
This takes advantage of a Windows 10 manifest property to set the [process code page to UTF-8](https://docs.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page#set-a-process-code-page-to-utf-8).

Prior to this update:

```
cppwinrt.exe -in d:\Test❤️\Windows.winmd -out d:\Test❤️
cppwinrt : error Path 'd:\Test??\Windows.winmd' is not a file or directory
```

With this update:

```
D:\git\cppwinrt\_build\x86\Release\cppwinrt.exe -in d:\Test❤️\Windows.winmd -out d:\Test❤️
```

Input is parsed correctly and output is generated correctly.

I'm reluctant to add a test for this scenario to avoid complicating the build.

Fixes: #1306

### DIFF
--- a/cppwinrt/app.manifest
+++ b/cppwinrt/app.manifest
@@ -3,6 +3,7 @@
 <application xmlns="urn:schemas-microsoft-com:asm.v3">
     <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
         <ws2:longPathAware>true</ws2:longPathAware>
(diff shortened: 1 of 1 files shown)


alvinhochun:
> @alvinhochun any suggestions for the Linux build failure?

Looks like a random unrelated failure in that one test. I suspect a thread left over from another test just happened to interfere with it, perhaps?

oldnewthing (approved):

Issue #1278, opened 2023-02-20
co_yield support for IIterable<T>


JaiganeshKumaran:
Provide the ability to generate a WinRT iterator object by using C++ coroutines.

By the way, please add an issue template for feature requests.

PR #1323, opened 2023-06-20
Add synchronous generator support for IIterable<T>

Fulgen301:
This implements #1278.

`IIterable::First()` throws `E_CHANGED_STATE` if it is called more than one time since the coroutine has already been resumed at that point; however, I am unsure whether this violates the [API contract](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.collections.iiterable-1.first?view=winrt-22621#windows-foundation-collections-iiterable-1-first):

> If changes are made to the collection, such as adding, modifying, or deleting elements, the iterator is permitted to raise an exception for all future operations.

### DIFF
--- a/strings/base_coroutine_collections.h
+++ b/strings/base_coroutine_collections.h
@@ -1,0 +1,204 @@
+
+#ifdef WINRT_IMPL_COROUTINES
+namespace winrt::impl
+{
+	template<typename Derived, typename TResult>
+	struct iterable_promise_base : implements<Derived, winrt::Windows::Foundation::Collections::IIterable<TResult>>
+	{
+	private:
+		struct iterator;
+
+		enum class IterationStatus
+		{
+			Initial,
+			Producing,
+			Value,
+			Done
+		};
+
+	public:
+		unsigned long __stdcall Release() noexcept
+		{
+			uint32_t const remaining = this->subtract_reference();
+
+			if (remaining == 0)
+			{
+				std::atomic_thread_fence(std::memory_order_acquire);
+				coroutine_handle<Derived>::from_promise(*static_cast<Derived*>(this)).destroy();
+			}
+
+			return remaining;
+		}
+
+		winrt::Windows::Foundation::Collections::IIterable<TResult> get_return_object() noexcept
+		{
+			return *this;
+		}
+
+		suspend_always initial_suspend() const noexcept
+		{
+			return {};
+		}
+
+		struct final_suspend_awaiter
+		{
+			iterable_promise_base* promise;
+
+			bool await_ready() const noexcept
+			{
+				return false;
+			}
+
+			void await_resume() const noexcept
+			{
+			}
+
+			bool await_suspend(coroutine_handle<>) const noexcept
+			{
+				uint32_t const remaining = promise->subtract_reference();
+
+				if (remaining == 0)
+				{
+					std::atomic_thread_fence(std::memory_order_acquire);
+				}
+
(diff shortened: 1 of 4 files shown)


Fulgen301:

strings/base_coroutine_collections.h
```diff
@@ -0,0 +1,204 @@
...
+namespace std
+{
+	template<typename T, typename... Args>
+	struct coroutine_traits<winrt::Windows::Foundation::Collections::IIterable<T>, Args...>
+	{
+		struct promise_type final : winrt::impl::iterable_promise_base<promise_type, T>
+		{
+			suspend_always yield_value(T&& value) noexcept
+			{
+				m_result = value;
```
sylveon, line 182: `std::move(value)`?

test/test/coro_collections.cpp
```diff
@@ -0,0 +1,38 @@
+// Intentionally not using pch...
```
sylveon, line 1: I'd recommend adding a test for GetMany(), as well as normally iterating e.g. `for (hstring h : Generator())`).

What about value types and reference types?

strings/base_coroutine_collections.h
```diff
@@ -0,0 +1,204 @@
...
+
+		private:
+			com_ptr<Derived> m_promise;
+		};
+
+		IterationStatus m_status{ IterationStatus::Initial };
+	};
+}
+
+namespace std
```
sylveon, line 173:
```suggestion
#ifdef __cpp_lib_coroutine
namespace std
#else
namespace std::experimental
#endif
```

test/test/coro_collections.cpp
sylveon: I'm curious what's the behavior regarding AddRef/Release counts. Also, it might be a good idea to add a test checking that not completing the iteration but dropping the IIterable & IIterator will free the coroutine frame.

sylveon (commented):
Looks pretty good, you got around to it before me lol

kennykerr:
Unfortunately, a project maintainer is not currently available to review this pull request. Please see the [contributing guide](https://github.com/microsoft/cppwinrt/blob/master/docs/contributing.md) for more information. Feel free to keep the conversation going on the related issue.

strings/base_coroutine_collections.h
```diff
@@ -0,0 +1,204 @@
+
+#ifdef WINRT_IMPL_COROUTINES
+namespace winrt::impl
+{
+	template<typename Derived, typename TResult>
+	struct iterable_promise_base : implements<Derived, winrt::Windows::Foundation::Collections::IIterable<TResult>>
+	{
+	private:
+		struct iterator;
```
ChrisGuzak, line 9: indentation looks wrong here. make sure to use 4 spaces to indent

jonwis:
> The https://github.com/microsoft/wil library is where we've been collecting implementation helpers such as this. You might want to open an issue on that repo for consideration.

Yes - this looks much more appropriate for wil or https://github.com/microsoft/cpp-async as it doesn't _really_ require cppwinrt on its own. Looking forward to the PR on either of those places!

sylveon:
Feels weird to have to IAsyncOperation support be built-in but not IIterable. Perhaps both should be moved to wil?

kennykerr:
Async coroutine support is fundamental to C++/WinRT for both consumption and production. Generator support for iteration, while cool, has very narrowly defined application.

Issue #1325, opened 2023-06-22
Bug: cpp20 source location test case failure


dmachaj:
### Summary
```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_cpp[20](https://github.com/microsoft/cppwinrt/actions/runs/5338545026/jobs/9676159938#step:15:21).exe is a Catch v2.13.9 host application.
Run with -? for options

-------------------------------------------------------------------------------
custom_error_logger
-------------------------------------------------------------------------------
D:\a\cppwinrt\cppwinrt\test\test_cpp20\custom_error.cpp(48)
...............................................................................

D:\a\cppwinrt\cppwinrt\test\test_cpp20\custom_error.cpp(70): FAILED:
  REQUIRE( functionNameSv == "FailOnLine15" )
with expansion:
  "void __cdecl `anonymous-namespace'::FailOnLine15(void)"
  ==
  "FailOnLine15"

===============================================================================
test cases:  6 |  5 passed | 1 failed
assertions: 26 | [25](https://github.com/microsoft/cppwinrt/actions/runs/5338545026/jobs/9676159938#step:15:26) passed | 1 failed

Error: Process completed with exit code 1.
```

riverar:
@dmachaj Think you missed a ` ``` ` after your log block.

dmachaj:
I didn't have any ` at all, so I'm not quite sure where the formatting came from.  Adding ``` before and after seems to have done the trick, though.

PR #1326, opened 2023-06-22
Fix cpp20 source location test failure resulting from newer compiler

dmachaj:
What's this all about?
Fix a test case failure caused by changing compiler behavior in a newer compiler version.  The signature that MSVC provides for source_location has changed recently.

Instead of chasing the full decorated name, with each compiler having a slight variation, we now match just the meaningful substring that we care about "FailOnLine15".  This should now be stable across compilers and across compiler versions.

Fixes: #1325

### DIFF
--- a/test/test_cpp20/custom_error.cpp
+++ b/test/test_cpp20/custom_error.cpp
@@ -61,13 +61,10 @@ TEST_CASE("custom_error_logger")
     REQUIRE(fileNameSv.find("custom_error.cpp") != std::string::npos);
     const auto functionNameSv = std::string_view(s_loggerArgs.functionName);
     REQUIRE(!functionNameSv.empty());
-#if defined(__GNUC__) && !defined(__clang__)
-    REQUIRE(functionNameSv == "void {anonymous}::FailOnLine15()");
-#elif defined(__GNUC__) && defined(__clang__)
-    REQUIRE(functionNameSv == "void (anonymous namespace)::FailOnLine15()");
-#else
-    REQUIRE(functionNameSv == "void __cdecl `anonymous-namespace'::FailOnLine15(void)");
-#endif
+    // Every compiler has a slightly different naming approach for this function, and even the same
+    // compiler can change its mind over time.  Instead of matching the entire function name just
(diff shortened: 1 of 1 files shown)


kennykerr (dismissed):
Thanks!

sylveon:
Would it make more sense to validate that functionNameSv contains the name of the function instead of using a strict ==

dmachaj:
> Would it make more sense to validate that functionNameSv contains the name of the function instead of using a strict ==

That's a good point.  That should still provide the validation we need while also being far more stable.  Great suggestion 👍

test/test_cpp20/custom_error.cpp
```diff
@@ -61,13 +61,10 @@ TEST_CASE("custom_error_logger")
...
-    REQUIRE(functionNameSv == "void {anonymous}::FailOnLine15()");
-#elif defined(__GNUC__) && defined(__clang__)
-    REQUIRE(functionNameSv == "void (anonymous namespace)::FailOnLine15()");
-#else
-    REQUIRE(functionNameSv == "FailOnLine15");
-#endif
+    // Every compiler has a slightly different naming approach for this function, and even the same
+    // compiler can change its mind over time.  Instead of matching the entire function name just
+    // match against the part we care about.
+    REQUIRE(functionNameSv.find("FailOnLine15" != std::string_view::npos));
```
kennykerr, line 67:
```suggestion
    REQUIRE((functionNameSv.find("FailOnLine15" != std::string_view::npos)));
```

Might need something like this to keep the preprocessor from exploding.

kennykerr: Also, that looks like a typo: `find("FailOnLine15"`

dmachaj: Wow, not sure how that compiled locally 🤦.  Fixed locally and will push as soon as build_test_all.cmd completes.

kennykerr: The preprocessor works in mysterious ways. 😊

kennykerr (approved):

Issue #1289, opened 2023-03-22
Allow passing array_view as span


JaiganeshKumaran:
I want to be able to use std::span for portablity and implicitly pass an array_view from WinRT to it.

kennykerr:
The reason the issues page only lets you create a bug report is because cppwinrt is in maintenance mode and no longer receiving new feature work. cppwinrt serves an important and specific role, but further feature development risks destabilizing the project. Additional helpers are regularly contributed to complimentary projects such as https://github.com/microsoft/wil/.

JaiganeshKumaran:
This is not an external helper, it must be added to the array_view type itself.

JaiganeshKumaran:
You can provide conversion operators although std::span using size_t for size could be an issue as array_view uses uin32_t.

sylveon:
> The reason the issues page only lets you create a bug report is because cppwinrt is in maintenance mode and no longer receiving new feature work.

This is a sad state of affairs considering WinRT is supposed to be the future of Windows development (WASDK & WinUI)

JaiganeshKumaran:
We need C++/WinRT to be updated for the win32metadata project to shine for C++, now that C++/Win32 is archived. We need a cleaner way of consuming COM and Win32 APIs without much of the manual HResult handling, and smart pointer noise.

kennykerr:
I don't think that #1331 is the best approach, as I indicated in https://github.com/microsoft/cppwinrt/issues/1289#issuecomment-1481317055 - `array_view` was always meant as a bandaid while we waited for `span` to standardize - but I will reopen the issue for discussion. 

I've tagged a few project maintainers. If one of them is available to offer guidance and mentorship for this issue, they will reach out according to the [contributing guide](https://github.com/microsoft/cppwinrt/blob/master/docs/contributing.md) to discuss and agree on an approach.

https://github.com/orgs/microsoft/teams/cppwinrt-maintainers

sylveon:
I'm on the side of just removing array_view, and using std::span - but that would mean bumping the minimum to C++20.

kennykerr:
> I'm on the side of just removing array_view, and using std::span - but that would mean bumping the minimum to C++20.

I'd be in favor of that.

JaiganeshKumaran:
> I'm on the side of just removing array_view, and using std::span - but that would mean bumping the minimum to C++20.
Why not simply alias std::span as array_view on C++20, but keep it for C++17?

[...]

PR #1331, opened 2023-07-12
Implicit conversion between std::span and winrt::array_view

JaiganeshKumaran:
Implements #1289.


kennykerr:
I don't think this is the best approach, as I indicated in the associated issue https://github.com/microsoft/cppwinrt/issues/1289#issuecomment-1481317055 but I will reopen the issue for discussion.

PR #1338, opened 2023-08-07
Increase foldability of various templates

### LINKED ISSUE #1334: Bug: implements_delegate non-fold-ability
### Version

2.0.220131.2

### Summary

Using SizeBench, some reasonable fraction of our DLL is taken up in `winrt::impl::implements_delegate` methods that aren't foldable for one of two reasons:

* The offset of the reference count member shifts around from +10 to +18 depending on the complexity of the delegate
* The "which guid are you?" checks end up generating relative `lea ...` operations.

### Reproducible example

_No response_

### Expected behavior

It _may_ be a linker/optimizer bug that the folder is not working "hard enough."

### Actual behavior

Multiple nonfolded specializations of delegate types taking ~1% of binary space (~15k out of ~1500k).

### Additional comments

In other places, I've used a "base" type to move as many things down to non-templated members as possible, with a templated caller sending in a parameters. For instance, in base.h, this appears to work and collapse all those methods into a small handful of specialized "push some parameters and call a helper" method.

Things that could use a similar treatment:

* winrt::event<T>::add - make_agile_delegate's use of guid_of<T>
* root_implements<T> - find_interface use of guids
* await_adapter<T>::await_suspend - looks like vtable handling

In my 1.5mb DLL, SizeBench claims those three waste another 1%.

```c++
namespace winrt::impl
{
#if defined(_MSC_VER)
#pragma warning(push)
#pragma warning(disable:4458) // declaration hides class member (okay because we do not use named members of base class)
#endif

    struct implements_delegate_base
    {
        WINRT_IMPL_NOINLINE uint32_t __stdcall AddRef() noexcept
        {
            return ++m_references;
        }

        WINRT_IMPL_NOINLINE uint32_t __stdcall Release() noexcept
        {
            return --m_references;
        }

        WINRT_IMPL_NOINLINE uint32_t QueryInterfaceCommon(guid const& id, void** result, unknown_abi* abi_t_ptr, guid const& delegate_id) noexcept
        {
            if ((delegate_id == id) || is_guid_of<Windows::Foundation::IUnknown>(id) || is_guid_of<IAgileObject>(id))
            {
                *result = abi_t_ptr;
                AddRef();
                return 0;
            }

            if (is_guid_of<IMarshal>(id))
            {
                return make_marshaler(abi_t_ptr, result);
            }

            *result = nullptr;
            return error_no_interface;
        }

    public:
        atomic_ref_count m_references{ 1 };
    };

    template <typename T, typename H>
    struct implements_delegate : implements_delegate_base, abi_t<T>, H, update_module_lock
    {
        implements_delegate(H&& handler) : H(std::forward<H>(handler))
        {
        }

        int32_t __stdcall QueryInterface(guid const& id, void** result) noexcept final { return implements_delegate_base::QueryInterfaceCommon(id, result, static_cast<abi_t<T>*>(this), guid_of<T>()); }

        uint32_t __stdcall AddRef() noexcept final { return implements_delegate_base::AddRef(); }

        uint32_t __stdcall Release() noexcept final
        {
            auto remaining = implements_delegate_base::Release();
            if (remaining == 0)
            {
                delete static_cast<delegate<T, H>*>(this);
            }
            return remaining;
        }
    };
```

jonwis:
SizeBench revealed heavily non-foldability of certain C++/WinRT templates contributing to increased Windows binary sizes. Some small tweaks dropped the size of `test.exe` (release, x64) by ~132kb. (5180416 -> 5030400, or 2.8% ... which isn't a super lot, but it's nonzero.

Fixes #1334 and related

## Changes

**Enable nuspec iteration** - Bump the default nuspec version to 3.0.0.0 so `build_nuget.cmd` and iterating with a local nuget source is easier.

**Add solution files** - Niceness for people who live in Visual Studio.

**Fold delegate bases** - Forced the layout of the delegate types to be regular by moving the reference count into a base _before_ the non-ABI types. This folds nearly all the AddRef/QueryInterface instances into one. Releases are still unique because they call up into dtors.

**Fold "report" in invoke** - The invoke helper is a try/catch{report} but the body of "report" is identical on each invoke specialization. Move it to a unique non-inlineable method.

**Fold event::add** - Make the delegate first, then all the "expand the collection" instances are identical.

## Further Work

SizeBench reports the following additional foldability wastage, quoted from the same test.exe:

* make_agile_delegate<> - 3.2kb; the agile thunk is unique basically per delegate `::Invoke` vtable and offset adjusters.
* root_implements::<T1,T2,T3>::query_interface - 4.1kb; again unique per vtable instance, also related to inlining choices
* winrt::event<T1>::add - 3.2kb; all cost related to inlined use of `make_agile_delegate`
* winrt::impl::runtime_class_name<T1,T2>::get() - 1.4kb; should probably switch to std::wstring_view but this may be a breaking change; currently induces an `hstring` construction.
 
## Testing

Tested by `build_and_test.cmd` - all built, all passed

### DIFF
--- a/strings/base_delegate.h
+++ b/strings/base_delegate.h
@@ -6,7 +6,7 @@ namespace winrt::impl
 #pragma warning(disable:4458) // declaration hides class member (okay because we do not use named members of base class)
 #endif
 
-    template<bool try_marshaling> struct implements_delegate_base
+    struct implements_delegate_base
     {
         WINRT_IMPL_NOINLINE uint32_t increment_reference() noexcept
         {
@@ -18,21 +18,18 @@ namespace winrt::impl
             return --m_references;
         }
 
-        WINRT_IMPL_NOINLINE uint32_t query_interface(guid const& id, void** result, unknown_abi* outerAbiPtr, guid const& outerId)
+        WINRT_IMPL_NOINLINE uint32_t query_interface(guid const& id, void** result, unknown_abi* derivedAbiPtr, guid const& derivedId) noexcept
         {
-            if ((id == outerId) || is_guid_of<Windows::Foundation::IUnknown>(id) || is_guid_of<IAgileObject>(id))
+            if (id == derivedId || is_guid_of<Windows::Foundation::IUnknown>(id) || is_guid_of<IAgileObject>(id))
             {
-                *result = outerAbiPtr;
+                *result = derivedAbiPtr;
                 increment_reference();
                 return 0;
             }
 
-            if constexpr (try_marshaling)
+            if (is_guid_of<IMarshal>(id))
             {
-                if (is_guid_of<IMarshal>(id))
-                {
-                    return make_marshaler(outerAbiPtr, result);
-                }
+                return make_marshaler(derivedAbiPtr, result);
             }
 
             *result = nullptr;
@@ -44,7 +41,7 @@ namespace winrt::impl
     };
 
     template <typename T, typename H>
-    struct implements_delegate : abi_t<T>, implements_delegate_base<true>, H, update_module_lock
+    struct implements_delegate : abi_t<T>, implements_delegate_base, H, update_module_lock
     {
         implements_delegate(H&& handler) : H(std::forward<H>(handler))
         {
@@ -118,7 +115,7 @@ namespace winrt::impl
     };
 
     template <typename H, typename R, typename... Args>
-    struct variadic_delegate final : variadic_delegate_abi<R, Args...>, implements_delegate_base<false>, H, update_module_lock
+    struct variadic_delegate final : variadic_delegate_abi<R, Args...>, implements_delegate_base, H, update_module_lock
     {
         variadic_delegate(H&& handler) : H(std::forward<H>(handler))
         {
--- a/strings/base_delegate.h
+++ b/strings/base_delegate.h
@@ -6,8 +6,42 @@ namespace winrt::impl
 #pragma warning(disable:4458) // declaration hides class member (okay because we do not use named members of base class)
 #endif
 
+    struct implements_delegate_base
+    {
+        WINRT_IMPL_NOINLINE uint32_t inc_ref() noexcept
+        {
+            return ++m_references;
+        }
+
+        WINRT_IMPL_NOINLINE uint32_t dec_ref() noexcept
+        {
+            return --m_references;
+        }
+
+        WINRT_IMPL_NOINLINE uint32_t QueryInterfaceImpl(guid const& id, void** result, unknown_abi* outerAbiPtr, guid const& outerId)
+        {
+            if ((id == outerId) || is_guid_of<Windows::Foundation::IUnknown>(id) || is_guid_of<IAgileObject>(id))
+            {
+                *result = outerAbiPtr;
+                inc_ref();
+                return 0;
+            }
+
+            if (is_guid_of<IMarshal>(id))
+            {
+                return make_marshaler(outerAbiPtr, result);
+            }
+
+            *result = nullptr;
+            return error_no_interface;
+        }
+
+    private:
+        atomic_ref_count m_references{ 1 };
+    };
+
     template <typename T, typename H>
-    struct implements_delegate : abi_t<T>, H, update_module_lock
+    struct implements_delegate : abi_t<T>, implements_delegate_base, H, update_module_lock
     {
         implements_delegate(H&& handler) : H(std::forward<H>(handler))
         {
@@ -15,30 +49,17 @@ namespace winrt::impl
 
         int32_t __stdcall QueryInterface(guid const& id, void** result) noexcept final
         {
-            if (is_guid_of<T>(id) || is_guid_of<Windows::Foundation::IUnknown>(id) || is_guid_of<IAgileObject>(id))
-            {
-                *result = static_cast<abi_t<T>*>(this);
-                AddRef();
-                return 0;
-            }
-
-            if (is_guid_of<IMarshal>(id))
-            {
-                return make_marshaler(this, result);
-            }
-
-            *result = nullptr;
-            return error_no_interface;
+            return QueryInterfaceImpl(id, result, static_cast<abi_t<T>*>(this), guid_of<T>());
         }
 
         uint32_t __stdcall AddRef() noexcept final
         {
-            return ++m_references;
+            return inc_ref();
         }
 
         uint32_t __stdcall Release() noexcept final
         {
-            auto const remaining = --m_references;
+            auto const remaining = dec_ref();
 
             if (remaining == 0)
             {
@@ -47,10 +68,6 @@ namespace winrt::impl
 
             return remaining;
         }
-
-    private:
-
-        atomic_ref_count m_references{ 1 };
     };
 
     template <typename T, typename H>
--- a/strings/base_events.h
+++ b/strings/base_events.h
@@ -342,23 +342,26 @@ namespace winrt::impl
         return 1;
     }
 
-    WINRT_IMPL_NOINLINE static bool report_failed_invoke()
+    struct report_helper
     {
-        int32_t const code = to_hresult();
-
-        static int32_t(__stdcall * handler)(int32_t, int32_t, void*) noexcept;
-        impl::load_runtime_function(L"combase.dll", "RoTransformError", handler, fallback_RoTransformError);
-        handler(code, 0, nullptr);
-
-        if (code == static_cast<int32_t>(0x80010108) || // RPC_E_DISCONNECTED
-            code == static_cast<int32_t>(0x800706BA) || // HRESULT_FROM_WIN32(RPC_S_SERVER_UNAVAILABLE)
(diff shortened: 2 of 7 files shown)


strings/base_events.h
```diff
@@ -510,6 +499,32 @@ WINRT_EXPORT namespace winrt
 
     private:
 
+        WINRT_IMPL_NOINLINE event_token add_agile(delegate_type delegate)
```
dmachaj, line 502: >delegate_type [](http://example.com/codeflow?start=50&length=13)

Can this be `const&` to avoid a copy?  Or does that somehow defeat the foldability?

dmachaj: Nevermind, you move out of it so you need a copy

jonwis: ... and it's the unnamed temporary from the call to make_agile_delegate.

dmachaj (approved):
:shipit:

strings/base_delegate.h
```diff
@@ -6,39 +6,63 @@ namespace winrt::impl
 #pragma warning(disable:4458) // declaration hides class member (okay because we do not use named members of base class)
 #endif
 
-    template <typename T, typename H>
-    struct implements_delegate : abi_t<T>, H, update_module_lock
+    template<bool try_marshaling> struct implements_delegate_base
     {
-        implements_delegate(H&& handler) : H(std::forward<H>(handler))
+        WINRT_IMPL_NOINLINE uint32_t inc_ref() noexcept
```
kennykerr, line 11: The cppwinrt library avoids abbreviations so this would be `increment_reference`.

strings/base_delegate.h
```diff
@@ -6,39 +6,63 @@ namespace winrt::impl
...
+            return ++m_references;
         }
 
-        int32_t __stdcall QueryInterface(guid const& id, void** result) noexcept final
+        WINRT_IMPL_NOINLINE uint32_t dec_ref() noexcept
+        {
+            return --m_references;
+        }
+
+        WINRT_IMPL_NOINLINE uint32_t QueryInterfaceImpl(guid const& id, void** result, unknown_abi* outerAbiPtr, guid const& outerId)
```
kennykerr, line 21: Non-projected names in cppwinrt are `lower_case` to avoid collisions. The `implements` class template for example uses `query_interface`.

jonwis:
For posterity, switching `name_of<>` and `get_name<>` to use `std::wstring_view` was basically a net wash.  No test code compilation errors that would signal a breaking change, overall binary increased by about 512 bytes.

One area of interest is in `GetRuntimeClassName`. The inliner is super aggressive about splatting `winrt::hstring(anything)` inline (and `~hstring`) rather than hoisting them out.  In PGO-optimized binaries this ctor is normally de-inlined since it's replicated countless times throughout the code. Making a custom `winrt::hstring get_runtime_class_name(std::wstring_view)` would likely not really help, as PGO should automatically produce that that when used.

jonwis:
It _looks_ like I can get another 8kb by regularizing the tail of `query_interface`. Running total after that is ~140kb.

It also looks like the cluster around `make_agile_delegate` could be connected to a simplification on how delegates are generated. Delegate ctors have a lot of overloads - lambda, function, raw-pointer member, com_ptr member, weak_ref member, shared_ptr member, weak_ptr member.  Under the covers they all route through the lambda form, doing various forms of "lock" and "call method." I wonder if there's a simplification of those to a two-form version.  Unless we were willing to try some stronger type-erasure through a std::function-alike, I'm not sure this can be reduced further.  It's probably another 3.2kb on this test binary, so under 0.5%.

strings/base_delegate.h
```diff
@@ -6,39 +6,63 @@ namespace winrt::impl
...
+            return ++m_references;
         }
 
-        int32_t __stdcall QueryInterface(guid const& id, void** result) noexcept final
+        WINRT_IMPL_NOINLINE uint32_t decrement_reference() noexcept
+        {
+            return --m_references;
+        }
+
+        WINRT_IMPL_NOINLINE uint32_t query_interface(guid const& id, void** result, unknown_abi* outerAbiPtr, guid const& outerId)
```
kennykerr, line 21:
```suggestion
        WINRT_IMPL_NOINLINE uint32_t query_interface(guid const& id, void** result, unknown_abi* outerAbiPtr, guid const& outerId) noexcept
```

kennykerr: Also, "outer" is really confusing just because of how that is used elsewhere in COM. Not sure what would be better though.

strings/base_delegate.h
```diff
@@ -6,39 +6,63 @@ namespace winrt::impl
 #pragma warning(disable:4458) // declaration hides class member (okay because we do not use named members of base class)
 #endif
 
-    template <typename T, typename H>
-    struct implements_delegate : abi_t<T>, H, update_module_lock
+    template<bool try_marshaling> struct implements_delegate_base
```
kennykerr, line 9: The `variadic_delegate` doesn't need to implement `IMarshal` but if it makes things simpler (by de-templatizing this) we could just do so conistently and avoid the split.

strings/base_events.h
```diff
@@ -342,14 +342,9 @@ namespace winrt::impl
...
-    template <typename Delegate, typename... Arg>
-    bool invoke(Delegate const& delegate, Arg const&... args) noexcept
+    struct report_helper
     {
-        try
-        {
-            delegate(args...);
-        }
-        catch (...)
+        WINRT_IMPL_NOINLINE static bool report_failed_invoke()
```
kennykerr, line 347: This seems overly complicated. The cppwinrt approach is to use `inline WINRT_IMPL_NOINLINE` on a free function the same way `winrt::throw_hresult` is written.

strings/base_implements.h
```diff
@@ -1156,44 +1156,48 @@ namespace winrt::impl
                 return 0;
             }
 
-            if constexpr (is_agile::value)
+            return query_interface_common(id, object);
+        }
+
+        WINRT_IMPL_NOINLINE int32_t query_interface_common(guid const& id, void** object) noexcept
```
kennykerr, line 1162: Is this really an improvement? `query_interface_common` is still within a template (so that things like `is_agile` works) but now the guid comparison for `IAgileObject` is always performed even when it is not needed. Same for `IWeakReferenceSource` and `IInspectable`. Even if there is a slight code size improvement, this ends up being more costly at run time as those QI evaluations are costly.

kennykerr:
Overall looks like goodness. Just a few comments.

strings/base_delegate.h
```diff
@@ -73,15 +93,16 @@ namespace winrt::impl
                 return delegate;
             }
 
+            const auto id = guid_of<T>();
```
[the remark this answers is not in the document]
kennykerr: I don't mind either way, but it is interesting. 😊

kennykerr:
> One area of interest is in GetRuntimeClassName. The inliner is super aggressive about splatting winrt::hstring(anything) inline (and ~hstring) rather than hoisting them out. In PGO-optimized binaries this ctor is normally de-inlined since it's replicated countless times throughout the code. Making a custom winrt::hstring get_runtime_class_name(std::wstring_view) would likely not really help, as PGO should automatically produce that that when used.

For [windows-rs](https://github.com/microsoft/windows-rs) I managed to write a "constexpr" version of `hstring` so that it can be declared as a compile-time constant. This is mainly used for literals, but it also turns these `GetRuntimeClassName` into a cheap pointer copy. There are some limitations, so I'm not sure its super practical for cppwinrt to adopt though.

sylveon:
Making the constructor of `param::hstring` constexpr seems pretty trivial, and it would enable constexpr fast-pass strings.

kennykerr:
> Making the constructor of `param::hstring` constexpr seems pretty trivial, and it would enable constexpr fast-pass strings.

I was referring to return values, which has a lot more optimizing value.

strings/base_delegate.h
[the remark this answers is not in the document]
jonwis: It seems to have made a tiny difference in folding of the resolve-and-call lambdas?  Consider these two interfaces:

```c++
[uuid(a)] interface IFoo { void Invoke(IPropertyValue v); }
[uuid(b)] interface IBar { void Invoke(IJsonValue v); }
```

The actual call sequence to call invoke is the same - load pointer-sized value of `v` from lambda storage, push value, call vtable[7]. The difference is the IID passed to `Resolve`. Making that also a member of lambda storage means this little funclet can be identical for every combination of `Invoke` parameter sets, rather than for parameter sets x interface IIDs.

strings/base_events.h
jonwis: I was innundated with compiler errors about "removing unreferenced method." The way around that was `WINRT_IMPL_NOINLINE inline ...` which seemed weird. I'll try again.

strings/base_delegate.h
jonwis: Maybe `derivedAbiPtr` and `derivedId` ?

strings/base_implements.h
jonwis: OK, reverting to the previous model (but keeping the split main-vs-common) retains the 140kb difference.

kennykerr (approved):
LGTM!

PR #1343, opened 2023-08-19
Support for std::span for winrt::array_view and winrt::com_array

### LINKED ISSUE #1289: Allow passing array_view as span

jonwis:
This fixes #1289 per "option C" by completing what #1331 started.

* Constructors for both array_view and com_array that take a span
* Conversions on both array_view and com_array that produce a span
* Tests for the conversions and constructions
* Additional ctad support for easy use

@JaiganeshKumaran thank you for the initial start - see if this meets your needs, please.


test/test_cpp20/array_span.cpp
```diff
@@ -0,0 +1,172 @@
...
+//
+// This test illustrates an array_view<T> (non-const) bound to a std::span<T> on a std::array
+//
+TEST_CASE("array,DataReader,std::span")
+{
+    auto reader = CreateDataReader({ 1, 2, 3 }).get();
+
+    std::array<byte, 3> a{};
+    std::span<byte> sp(a);
+    reader.ReadBytes(sp); // FillArray pattern
```
sylveon, line 37: Out of curiosity, are we able to directly pass std::array here now?

strings/base_array.h
```diff
@@ -31,6 +31,20 @@ WINRT_EXPORT namespace winrt
             array_view(value.begin(), static_cast<size_type>(value.size()))
         {}
 
+#ifdef __cpp_lib_span
+        template <typename C, std::size_t extent>
+        array_view(std::span<C, extent> span) noexcept :
+            array_view(span.data(), static_cast<size_type>(span.size())) 
+        {
+            WINRT_ASSERT(span.size() <= UINT_MAX);
```
JaiganeshKumaran, line 39: There are other places (like the constructor of hstring) where currently no check is done. Need to also add this assertion there. Another thing, I would probably recommend using WINRT_VERIFY instead as WINRT_ASSERT is removed in release mode.

jonwis: Looks like this should have been an `if (too large) { throw invalid arg }` instead. That's what other places do.

jonwis: But then there's the other places in array_view that are more blase about it:

```c++
        array_view(pointer first, pointer last) noexcept :
            m_data(first),
            m_size(static_cast<size_type>(last - first))
        {}

        array_view(std::initializer_list<value_type> value) noexcept :
            array_view(value.begin(), static_cast<size_type>(value.size()))
        {}
```

@kennykerr what's the overall philosophy here? While it's _super unlikely_ that anybody passes UINT_MAX elements to any WinRT API, I suppose it's not unreasonable...

jonwis: (And now you've got me worried about truncation errors in string comparisons and other things...)

kennykerr: C++/WinRT will mostly (debug) assert for things like this. The primary outlier is when an HSTRING initializer is not null terminated in which case it panics. 

I'm not opposed to turning these into exceptions but that would ideally be done consistently throughout the library and consideration be taken about code size impact.

kennykerr:
Hopefully @DefaultRyan can spare a few minutes to look this over. He probably has the best handle on the standard library best practices.

strings/base_array.h
```diff
@@ -223,6 +237,11 @@ WINRT_EXPORT namespace winrt
     template <typename C, size_t N> array_view(std::array<C, N>& value) -> array_view<C>;
     template <typename C, size_t N> array_view(std::array<C, N> const& value) -> array_view<C const>;
 
+#ifdef __cpp_lib_span
+    template <typename C, std::size_t extent> array_view(std::span<C, extent>& value) -> array_view<C>;
```
kennykerr, line 241: cppwinrt uses the undecorated integral types like `size_t` from `stdint.h` so no need for `std...`.

strings/base_array.h
JaiganeshKumaran: The issue of returning span for a local vector could be solved by deleting the rvalue qualified conversion operator but unfortunately, the standard library does not use this mechanism anywhere.

jonwis: Tracked in #1347 so we can make progress here.

kennykerr (approved):
I still think the lifetime here is bonkers, but admittedly not any worse than the C++ standard library.

dmachaj:
Sorry I'm a little late to the party.  Do these new methods affect vtable layout at all?  In other words, does span support need an ODR pragma to prevent mismatches?

kennykerr:
> Do these new methods affect vtable layout at all?

`std::span` is not an ABI type.

jonwis:
It's pure conversion methods; no change to the layout or functionality of `array_view` or `com_array` here.

Issue #1345, opened 2023-08-22
Feature Request: Allow partial custom activation handlers


CarlosNihelton:
### Version

C++/WinRT v2.0.220110.5

### Summary

The conversation  https://github.com/microsoft/cppwinrt/pull/477#discussion_r1300499404 suggests that it's desirable to have a way for an activation handler to return back to the default procedure in case it's not interested in a certain class. That's useful when tests deal with a couple of WinRT types, but only part of them are to be mocked.

The inclusion of the global `winrt_activation_handler` in #477 requires the custom activation handler to deal with all possible activation scenarios. In certain kinds of tests that might be an overkill. 

A small change as suggested below would allow handlers by convention to return a magic value meaning they cannot handle that activation and the original procedure should take over.

```cpp
        if (winrt_activation_handler)
        {
            if (auto hr = winrt_activation_handler(*(void**)(&name), guid_of<Interface>(), result);
                hr != impl::error_not_implemented) {
                    return hr;
                }
        }
```

Augmenting `test\custom_activation.cpp` to try instantiating some other class the `custom_factory` is not interested in would result in a uncorrectable failure (even if we'd remove the `REQUIRE` clauses from the handler and the custom_factory implementation code).

### Reproducible example

```cpp
// in the end of `test\custom_activation.cpp`
// An activation handler that doesn't handle WwwFormUrlDecoder activation.
bool invoked_unimplemented = false;
int32_t __stdcall another_handler(void* classId, winrt::guid const& iid, void** factory) noexcept
{
    if (iid == guid_of<IWwwFormUrlDecoderRuntimeClassFactory>()) {
        invoked_unimplemented = true;
        return winrt::impl::error_not_implemented;
    }
    *factory = detach_abi(make<custom_factory>());
    return 0;
}

TEST_CASE("partial_custom_activation")
{
    clear_factory_cache();

    // Set up global handler
    REQUIRE(!winrt_activation_handler);
    winrt_activation_handler = another_handler;
    REQUIRE(!invoked_unimplemented);

    // Activates something the handler cannot deal with. It shouldn't crash and the decoder should behave normally.
    WwwFormUrlDecoder decoder{ L"uname=c&passwd=d" };
    REQUIRE(invoked_unimplemented);
    REQUIRE(decoder.GetFirstValueByName(L"uname") == L"c");
    REQUIRE(decoder.GetFirstValueByName(L"passwd") == L"d");

    // Remove global handler
[...]

PR #1346, opened 2023-08-22
Allows partial activation handling

CarlosNihelton:
With the suggested changes a test suite can handle only the activation it's interested in, leaving others for the default implementation, conventionally returning `winrt::impl::error_not_implemented`.

Fixes: #1345


strings/base_activation.h
```diff
@@ -30,7 +30,10 @@ namespace winrt::impl
     {
         if (winrt_activation_handler)
         {
-            return winrt_activation_handler(*(void**)(&name), guid, result);
+            hresult hr = winrt_activation_handler(*(void**)(&name), guid, result);
+            if (hr != impl::error_not_implemented) {
```
oldnewthing, line 34: I would prefer a different signal. There are cases where it's possible for the real `RoGetActivationFactory()` to return `E_NOTIMPL`. For example, if it successfully loads the server DLL, gets the activation factory, calls `DllGetActivationFactory`, and the server's `DllGetActivationFactory` returns `E_NOTIMPL`, then that error bubbles all the way out from `RoGetActivationFactory`. We don't want this to trigger the fallback code path. We want this to be the definitive answer: It wasn't implemented.

I would suggest that we use the signal `(hr == 0 && !result)`. This would lead to a crash, so we can be pretty confident that nobody is doing it by mistake.

CarlosNihelton: I believe you meant `(hr == 0 && !(*result))`.

kennykerr:
`winrt_activation_handler` was added for the Shell team so I'll defer to @ChrisGuzak or @oldnewthing on whether this change is appropriate.

strings/base_activation.h
Scottj1s: I would go further and suggest a totally new hook, like 'bool try_winrt_activation(void* classId, winrt::guid const& iid, void** factory, int32_t hr) noexcept' which takes precedence over winrt_activation_handler.  then we're not muddying the existing contract.

CarlosNihelton: I changed the implementation towards @Scottj1s 's recommendation.

test/test/custom_activation.cpp
```diff
@@ -50,3 +50,41 @@ TEST_CASE("custom_activation")
...
+        invoked_unimplemented = true;
+        // to be explicit.
+        *factory = nullptr;
+        *hr = 0;
+        return false;
+    }
+    *factory = detach_abi(make<custom_factory>());
+    *hr = 0;
+    return true;
+}
```
oldnewthing, line 69: This handler is highly non-idiomatic. Should test the class name, not the interface; the fallthrough should be the failure case (in anticipation of wanting to hook multiple classes).
```cpp
std::wstring_view const name{ *reinterpret_cast<winrt::hstring*>(&classId) };
if (name == L"Windows.Foundation.Uri"v) {
    *factory = detach_abi(make<custom_factory>());
    *hr = S_OK;
    return true;
}
else
{
    // Leave unhandled
    invoked_unimplemented = true;
    return false;
}
```
This sets a good example for others who are likely to copy this code for their own `try_winrt_activation_handler` handler.

oldnewthing: Oops, I missed a spot in my previous recommendation. This function is noexcept, so we need to catch exceptions that come out of `make` and turn them into `hresult`s.

```cpp
bool __stdcall uri_only_activation_handler(void* classId, winrt::guid const& iid, void** factory, int32_t* hr) noexcept try
{
    ...
}
catch (...)
{
    return winrt::to_hresult();
}
```

test/test/custom_activation.cpp
```diff
@@ -50,3 +50,45 @@ TEST_CASE("custom_activation")
...
+// An activation handler that can only deal with Windows.Foundation.Uri.
+bool invoked_unimplemented = false;
+bool __stdcall uri_only_activation_handler(void* classId, winrt::guid const& iid, void** factory, int32_t* hr) noexcept
+{
+    using namespace std::literals::string_view_literals;
+
+    std::wstring_view const name{ *reinterpret_cast<winrt::hstring*>(&classId) };
+    // custom_factory is only interested in Windows.Foundation.Uri
+    if (name == L"Windows.Foundation.Uri"sv) {
+        *factory = detach_abi(make<custom_factory>());
```
oldnewthing, line 63: This needs to honor the `iid`.

```suggestion
        *hr = make<custom_factory>().as(iid, factory);
```

test/test/custom_activation.cpp
```diff
@@ -50,3 +50,45 @@ TEST_CASE("custom_activation")
...
+{
+    clear_factory_cache();
+
+    // Set up global handler
+    REQUIRE(!winrt_activation_handler);
+    REQUIRE(!try_winrt_activation_handler);
+    try_winrt_activation_handler = uri_only_activation_handler;
+    REQUIRE(!invoked_unimplemented);
+
+    // Activates something the handler cannot deal with. It shouldn't crash and the decoder should behave normally.
```
oldnewthing, line 85: This test should also activate a URI to confirm that the `try` is working.

test/test/custom_activation.cpp
CarlosNihelton: When I saw your first comment I considered doing the fix in another PR, but since you mentioned here I took the freedom to do some more related refactoring on the existing test case as well.

CarlosNihelton:
I guess we are done with this PR, right?!

CarlosNihelton:
Well, thanks for the attention on this topic anyway! It was quite enlightening.

Issue #1344, opened 2023-08-21
Remove old Windows 7 support code


kennykerr:
Probably want to consider dropping Windows 7 support some time as well. That could further simplify this stuff a bit.

_Originally posted by @kennykerr in https://github.com/microsoft/cppwinrt/pull/1342#discussion_r1300565346_

jonwis:
Generally this is righteous.  Kind of wish we had better telemetry to see where this is really being used.  I'm concerned that if this flows back into the PSDK copy of cppwinrt it'll break people who were somehow relying on it. https://learn.microsoft.com/en-US/lifecycle/faq/extended-security-updates says that there are Win7 variants supported until 2024.

kennykerr:
Older versions of the C++/WinRT nuget package continue to be the supported way of handling this. This is no different. We don't encourage developers to use the version of C++/WinRT that ships in the Windows SDK since that version is locked to the version that *that* particular version of the OS was built with which is almost always extremely old (currently Jan 2022). In fact, half the time it doesn't even build with current MSVC or Clang compilers. It exists in the Windows SDK for legacy reasons only.

Issue #1341, opened 2023-08-17
Bug: linux hosted builds fail due to missing WinMD


jonwis:
### Summary

#1239 added use of a Windows.WinMD file from the windows-rs repo to help validate conformance of cppwinrt's output. https://github.com/microsoft/windows-rs/pull/2609/files?file-filters%5B%5D=.winmd&show-deleted-files=true&show-viewed-files=false# renamed the file in that repo, causing the existing `curl ...` command in our action to fail.

We should probably have our own copy of the WinMD as test collateral available in this repo.  In an ideal world, we wouldn't rely on Windows.WinMD at all and instead use a local custom metaplatform.winmd, but that's not feasible due to heavy reliance on `Windows.Foundation.*` types, behaviors, and attributes. It is, after all, C++/Win[dows]RT.

kennykerr:
FYI @alvinhochun

kennykerr:
I much prefer smaller discrete yml workflows (hence [this collection](https://github.com/microsoft/windows-rs/tree/master/.github/workflows)) because then you can just disable individual workflows via the GitHub Actions UX.

kennykerr:
I've also had to remove tests that rely on GNU to build in #1348 since their import libs don't include some Windows 8 exports.

PR #1348, opened 2023-08-25
Remove old Windows 7 support code

kennykerr:
Fixes: #1344


kennykerr:
I've also removed tests that rely on GNU to build since their import libs don't include some Windows 8 exports. We can decide what to ultimately do about that here: #1341

jonwis (approved):

alvinhochun:
I believe the mingw-w64 link errors can be fixed by adding `combase` to the link libraries.

Issue #1355, opened 2023-09-07
Feature request: Overriding "resume in same apartment" behavior for `co_await IAsyncXxx`


oldnewthing:
### Version

2.0.230706.1

### Summary

By default `co_await IAsyncXxx` resumes in the same COM apartment. Sometimes, this is not desirable behavior. In C#, you can use `ConfigureAwait(false)` to disable the resume-in-same-apartment behavior. C++/WinRT would benefit from having a corresponding capability.

A wil-based version was implemented in microsoft/wil#325 but @kennykerr allowd this to be added to core C++/WinRT since it greatly simplifies the implementation, and prevents it from falling out of sync.

### Reproducible example

```cpp
#include <winrt/Windows.Foundation.h>

using namespace winrt;
using namespace winrt::Windows::Foundation;

extern IAsyncAction Something();

IAsyncAction Example()
{
    // This co_await resumes in the same apartment.
    co_await Something();

    // This co_await is permitted to resume in any apartment.
    co_await winrt::resume_agile(Something());
}
```

PR #1356, opened 2023-09-07
resume_agile to allow coroutine to resume in any apartment

oldnewthing:
By default, `co_await`'ing a Windows Runtime asynchronous operation resumes in the same COM apartment. We provide a way to override this behavior and resume in any apartment by wrapping the asynchronous operation in `resume_agile()`.

```cpp
extern IAsyncAction Example();

// This resumes in the same apartment
co_await Example();

// This resumes in any apartment
co_await resume_agile(Example());
```

The guts of the change are in `disconnect_aware_handler` which accepts a `preserve_context` template parameter, defaulting to `true`. If `false`, then we bypass the `resume_apartment()` on completion. The `false` value is provided by the `resume_agile()` function.

To remove the `resume_apartment_context` from the `disconnect_aware_handler` in the case where we don't need it, I use EBO and make it a base class, so that it disappears when an empty `ignore_apartment_context` is used.

While I was there, I introduced `movable_primitive<T>` which provides RAII-like functionality to scalars like integers and pointers. This allows us to simplify `disconnect_aware_handler` and `resume_apartment_context`, which previously had to override all the copy and move operations in order to reset the scalar on move.

Also fix the `await_adapter.cpp` tests so they don't leak DispatcherQueues, by explicitly shutting down all the dispatcher queues we created. This required moving all the controllers to top-level so we can shut them down after the tests have completed.

Fixes: #1355

### DIFF
--- a/strings/base_coroutine_foundation.h
+++ b/strings/base_coroutine_foundation.h
@@ -99,44 +99,49 @@ namespace winrt::impl
         return async.GetResults();
     }
 
-    template<typename Awaiter>
-    struct disconnect_aware_handler
+    struct ignore_apartment_context {};
+
+    template<bool preserve_context, typename Awaiter>
+    struct disconnect_aware_handler : private std::conditional_t<preserve_context, resume_apartment_context, ignore_apartment_context>
     {
         disconnect_aware_handler(Awaiter* awaiter, coroutine_handle<> handle) noexcept
             : m_awaiter(awaiter), m_handle(handle) { }
 
-        disconnect_aware_handler(disconnect_aware_handler&& other) noexcept
-            : m_context(std::move(other.m_context))
-            , m_awaiter(std::exchange(other.m_awaiter, {}))
-            , m_handle(std::exchange(other.m_handle, {})) { }
+        disconnect_aware_handler(disconnect_aware_handler&& other) = default;
 
         ~disconnect_aware_handler()
         {
-            if (m_handle) Complete();
+            if (m_handle.value) Complete();
         }
 
         template<typename Async>
         void operator()(Async&&, Windows::Foundation::AsyncStatus status)
         {
-            m_awaiter->status = status;
+            m_awaiter.value->status = status;
             Complete();
         }
 
     private:
-        resume_apartment_context m_context;
-        Awaiter* m_awaiter;
-        coroutine_handle<> m_handle;
+        movable_primitive<Awaiter*> m_awaiter;
+        movable_primitive<coroutine_handle<>, nullptr> m_handle;
 
         void Complete()
         {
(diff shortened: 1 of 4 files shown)


oldnewthing:
Open issues:

* What should we call this thing? Is it `resume_agile()`? `resume_any_apartment()`? Something else?
* Should it be a free function or a member? (Or both?)

| Name | Free | Member |
|-|-|-|
| `resume_agile` | `co_await winrt::resume_agile(Something());` | `co_await Something().resume_agile();` |
| `resume_any_apartment` | `co_await winrt::resume_any_apartment(Something());` | `co_await Something().resume_any_apartment();` |

Note that C# uses `await Something().ConfigureAwait(true);`, going for the "member function" column, though the name "ConfigureAwait" is quite bad.

jonwis:
My vote is for free function. Our policy so far has to put these things as free functions so discoverability doesn't seem like a problem here ... resume_foreground, cancellation helpers, etc. If you want this, you know you need it, so you know where to find it.  (I'm of two minds about monkeypatching via extension methods for the same reason.)

Can you capture it by move, so someone who says `co_await resume_agile(ZorpAsync())` gets the best results?

I like `resume_agile()` as the naming.

Scottj1s (approved):

kennykerr (approved):
LGTM

oldnewthing:
> Can you capture it by move, so someone who says `co_await resume_agile(ZorpAsync())` gets the best results?

I like this idea - it gives us the best of both worlds.

* `co_await resume_async(ZorpAsync())` avoids extraneous addref/release.
* `IAsyncAction op = ZorpAsync(); co_await resume_async(op);` performs the necessary addref/release - avoids UAF.
* `IAsyncAction op = ZorpAsync(); co_await resume_async(std::move(op));` transfers ownership to `resume_await` and avoids UAF.

sylveon:
await_adapter is not owning, it holds `Async const&`

oldnewthing:
> await_adapter is not owning, it holds `Async const&`

The difference is that up until now, `await_adapter` is produced only by `operator co_await`, so you knew that the `await_adapter` was going to be awaited immediately.

This change introduces a way to obtain an `await_adapter` without awaiting it, namely by doing
```cpp
auto oops = resume_agile(ZorpAsync());
co_await oops;
```

sylveon:
That is a fair assessment, though one could already get an adapter without awaiting by calling the operator manually.

To make this work, await_adapter and all operators would need to be changed to take ownership, not just resume_agile.

oldnewthing:
Somebody who invokes the operator manually `auto oops = operator co_await(ZorpAsync())` knows that they are holding a live wire.

Issue #1360, opened 2023-09-17
Ability to remove built-in reference counting from a implementation type


JaiganeshKumaran:
Currently every WinRT implementation type inheriting from implements<T> automatically gets reference counting implemented. While you could override AddRef and Release, you still get the reference count field which cannot be removed. In some scenarios, you might want to eliminate the reference count field and do it all manually. These might include:
- Allocating many objects from a shared memory pool, which means the lifetime of the object needs to be tied to the pool and you may want to a single reference count for all of them.
- A singleton or class factory with static lifetime where you could make AddRef and Release not do anything at all.

I propose adding a new marker type, winrt::no_ref_count which will remove the ref count field, as well as the default implementations of AddRef and Release.

kennykerr:
I've tagged a few project maintainers. If one of them is available to offer guidance and mentorship for this issue, they will reach out according to the [contributing guide](https://github.com/microsoft/cppwinrt/blob/master/docs/contributing.md) to discuss and agree on an approach.

@microsoft/cppwinrt-maintainers

FWIW the `implements` class template is terribly complicated already. I would recommend you simply use your own fork for such dramatic customizations.

oldnewthing:
Overriding `AddRef` and `Release` are not sufficient because `m_references` is also used for weak reference support.

Even static lifetime objects need to implement `AddRef` and `Release` so they can tear down at COM uninitialization. Otherwise, you get bugs where the DLL is unloaded (DllCanUnloadNow) because it lost track of its static lifetime objects.

JaiganeshKumaran:
> Otherwise, you get bugs where the DLL is unloaded (DllCanUnloadNow) because it lost track of its static lifetime object
Well I can prevent unload and have it has a pinned DLL.

JaiganeshKumaran:
Currently I have complex cyclic object relationship where both objects must be alive at simultaneously. Weak references won't in my case because once get destructed while the other is alive. Having this feature would be really helpful so I can use a shared ref count for the two objects. Without it, I'm forced implement the ABI interfaces (or inherit from impl::produce_t<Interface>), without the help of winrt::implements.

oldnewthing:
[...]

PR #1365, opened 2023-09-28
Complete customisation of reference counting

JaiganeshKumaran:
Implements #1360.

Notes:
- no_ref_count implies no_weak_ref.
- Users implement `add_ref` and `release_ref`, rather than `(NonDelegating)AddRef/Release`.

I have a scenario that requires this feature - where two objects dependent on each other need to have the same lifetime (one should not get deleted while other is alive). `weak_ref` won't work for this but by overriding AddRef/Release, I can share the reference count for both objects.

### DIFF
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -17,6 +17,7 @@ namespace winrt::impl
 WINRT_EXPORT namespace winrt
 {
     struct non_agile : impl::marker {};
+    struct no_ref_count : impl::marker {};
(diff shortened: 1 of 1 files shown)


kennykerr:
Unfortunately, this can't go anywhere without a maintainer to steward this feature.

https://github.com/microsoft/cppwinrt/blob/master/docs/contributing.md

I can reopen the issue if you'd like to discuss further.

PR #1373, opened 2023-11-25
Update GitHub action LLVM version to 17.0.5

jonwis:
In #1372 it looks like the build VMs were updated to a more recent VS and VCRT and STL combination, requiring an updated LLVM toolchain. This bumps the target LLM version to match.

### DIFF
--- a/.github/actions/setup-llvm-msvc/action.yml
+++ b/.github/actions/setup-llvm-msvc/action.yml
@@ -4,7 +4,7 @@ inputs:
   llvm-version:
     description: 'LLVM version'
     required: true
-    default: '15.0.5'
+    default: '17.0.5'
 outputs:
   llvm-path:
     description: "The path in which LLVM is installed to"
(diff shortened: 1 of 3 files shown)


jonwis:
A new and interesting errro!

jonwis:
Turns out that LLVM 17.0.5 resolved this complaint:

https://github.com/microsoft/cppwinrt/blob/bf4459b25aeeb7093e8262fb3b29ca70b0ab2e60/test/test_cpp20/custom_error.cpp#L38-L42

... so the test metadata needs to be updated to say that clang is no longer "!shouldfail"

jonwis:
Notably, the test_component component does not build correctly here:

```
msbuild /m /p:Configuration=Debug,Platform=x64,Clang=1,LLVMToolsVersion=17.0.5,LLVMInstallDir=C:\llvm cppwinrt.sln
```

At least on my machine, the `test_component.vcxproj` picks up `C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin\clang-cl.exe` ... everything else correctly picks up the `$(SolutionRootDir).llvm` path.  This causes a series of errors in `yvals_core.h` due to undefined `uintptr_t`

A problem for another day.

kennykerr (approved):

Issue #1300, opened 2023-04-28
Simplify static lifetime implementation


JaiganeshKumaran:
When component optimisations are enabled, to have a static lifetime factory and define the static methods in the factory, rather than the implementation type, you need to write static forwarders in the implementation type as shown here: https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/static-lifetime#examples.

Consider making the generated code in g.cpp call into the factory directly if the static_lifetime marker is present instead of calling the static method on the implementation type.

kennykerr:
The relationship between optimizations and static factories is regrettably complex but isn't something we can easily fix without a major overhaul.

JaiganeshKumaran:
Could you provide a generated base class for the implementation type that will automatically call into the factory instead?

JaiganeshKumaran:
Here's a simple solution.
```cpp
namespace winrt::Component
{
    template <int = 0>
    static auto GetObjectHelperStatics()
    {
        if constexpr (winrt::impl::has_static_lifetime_v<factory_implementation::Foo>)
        {
            return winrt::make_self<factory_implementation::Foo>();
        }
        else
        {
            // statics_type will be a new alias on the factory based on FooT's second template parameter.
            return std::add_pointer_t<factory_implementation::Foo::statics_type>{ nullptr };
        }
    }

    void Foo::StaticMethod()
    {
        // Before
        return implementation::Foo::StaticMethod();
        // After.
        return GetObjectHelperStatics()->StaticMethod();
    }

PR #1375, opened 2023-12-09
Simply static lifetime implementation

JaiganeshKumaran:
Implements #1300.

Please merge this pull request as I really need this feature.

### DIFF
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -468,29 +468,6 @@ catch (...) { return winrt::to_hresult(); }
             }
             else if (factory.statics)
             {
-                /* {
-                    auto format = R"(    template <int = 0>
-    static auto Get%Statics()
-    {
-        if constexpr (winrt::impl::has_static_lifetime_v<@::factory_implementation::%>)
-        {
-            return winrt::make_self<@::factory_implementation::%>();
-        }
-        else
-        {
-            return std::add_pointer_t<@::factory_implementation::%::statics_type>{ nullptr };
-        }
-    })";
-                    w.write(format,
-                        type_name,
-                        type_namespace,
-                        type_name,
-                        type_namespace,
-                        type_name,
-                        type_namespace,
-                        type_name);
-                }*/
-
                 for (auto&& method : factory.type.MethodList())
                 {
                     method_signature signature{ method };
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -468,6 +468,29 @@ catch (...) { return winrt::to_hresult(); }
             }
             else if (factory.statics)
             {
+                {
+                    auto format = R"(    template <int = 0>
+    static auto Get%Statics()
+    {
+        if constexpr (winrt::impl::has_static_lifetime_v<@::factory_implementation::%>)
+        {
+            return winrt::make_self<@::factory_implementation::%>();
+        }
+        else
+        {
+            return std::add_pointer_t<@::factory_implementation::%::statics_type>{ nullptr };
+        }
+    })";
+                    w.write(format,
+                        type_name,
+                        type_namespace,
+                        type_name,
+                        type_namespace,
+                        type_name,
+                        type_namespace,
+                        type_name);
+                }
+
                 for (auto&& method : factory.type.MethodList())
                 {
                     method_signature signature{ method };
@@ -498,7 +521,7 @@ catch (...) { return winrt::to_hresult(); }
                     {
                         auto format = R"(    % %::%(%)
     {
-        %@::implementation::%::%(%);
+        %Get%Statics()->%(%);
     }
 )";
 
@@ -510,7 +533,6 @@ catch (...) { return winrt::to_hresult(); }
                             method_name,
                             bind<write_consume_params>(signature),
                             ignore_return ? "" : "return ",
-                            type_namespace,
                             type_name,
                             method_name,
                             bind<write_consume_args>(signature));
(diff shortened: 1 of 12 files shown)


JaiganeshKumaran:
Another improvement can be done here - regarding constructors. I sometimes want to write the CreateInstance* functions manually due to have more than one implementation class for the runtime class.

JaiganeshKumaran:
@kennykerr @oldnewthing

Link1J:
I can't make sense of what optimizations this does.
This breaks the current system to implement something that already is doable.
Just put the method you want to implement on the `factory_implementation` as non-static.
Which is what you do, so all this saves is the empty class with unimplemented methods?
I can't see how this is helpful.

JaiganeshKumaran:
@Link1J
Static lifetime is a feature that ensures that the factory object is only created. This is useful for bundling additional state with the static class that needs to be initialised on demand.

> Just put the method you want to implement on the factory_implementation as non-static.

Yes, that is already possible. However, when you use enable component optimisations[^1], it does not compile because the wrapper code in Class.g.cpp always accesses implementation::Class, implying that the functions must be present on it, which conflicts with static lifetime as you really want to put the static methods as non-static ones on the factory to access the state. Therefore, at the moment you need to manually write forwarder methods as noted [here](https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/static-lifetime#examples).

[^1]: Component optimisations provide 'inline' wrappers for static methods on the projected type instead of calling through the interface like usual.

kennykerr:
Sorry, I've not had a moment to look at this.

jonwis:
It feels a little hinky to rely on "call a static method on a null this _kind of_ works..."

So you goal is to eliminate the "if you are using component optimizations [which everybody should be] then you will need to write forwarder methods" part?

cppwinrt/component_writers.h
```diff
@@ -468,6 +468,29 @@ catch (...) { return winrt::to_hresult(); }
             }
             else if (factory.statics)
             {
+                {
+                    auto format = R"(    template <int = 0>
+    static auto Get%Statics()
+    {
+        if constexpr (winrt::impl::has_static_lifetime_v<@::factory_implementation::%>)
```
jonwis, line 475: Am I reading this right, and the logic here is:

1. If the factory has static_lifetime, use it in the normal way
2. Otherwise, generate a null pointer and return it and call a method on that ?

Feels like this should be the other way, or some other marker beyond static-lifetime. Won't this break existing code that has static_lifetime on the factory? Maybe the check should be on the implementation type.

Similarly, where should an implementation marked static_lifetime store its state? The factory static_lifetime does instantiate the factory, it just stores it statically for the lifetime of the program.

jonwis (changes requested):

cppwinrt/component_writers.h
JaiganeshKumaran: > Feels like this should be the other way,
It is the opposite. Currently, it always directly calls into the implementation type's static method.

> Won't this break existing code that has static_lifetime on the factory?
No. By default the CRTP base class for the factory writes non-static member functions that call into the implementation type's static method. If existing code has static_lifetime for the factory but does not have the methods directly on it, it is still a non-issue because the base class would provide those.

```cpp
struct Class : implementation::ClassT<Class>
{
    static void StaticMethod();
};

struct Class : ClassT<Class, implementation::Class, static_lifetime>
{
   // Part of the base class.
   void StaticMethod() { implementation::Class::StaticMethod(); }
};

> Similarly, where should an implementation marked static_lifetime store its state?
I don't understand your question. The state should be stored in the factory as before. Nothing has changed here.

JaiganeshKumaran:
> It feels a little hinky to rely on "call a static method on a null this _kind of_ works..."
> 
Yes, but it seems to be the simplest way to solve this problem. Also, note that there is no undefined behaviour because the method has to be static and C++ allows calling static methods of a type through a pointer or reference of type.

sylveon:
Actually, it is. `->` dereferences its left operand. Dereferencing a null pointer is UB.

JaiganeshKumaran:
> Actually, it is. `->` dereferences its left operand. Dereferencing a null pointer is UB.

Okay, but it works so not going to care. It's not like C++/WinRT has no UB already.
The alternate option would be to manually add the check for static lifetime in every single function.

tim-weis:
> Okay, but it works so not going to care.

That's hard to prove (and by "hard" I mean "impossible"). I will admit that this is almost guaranteed to remain inconsequential with MSVC, given that it has to support MFC sporting its infamous [`CWnd::GetSafeHwnd()`](https://learn.microsoft.com/en-us/cpp/mfc/reference/cwnd-class#getsafehwnd) implementation. But Clang will most certainly be far less dovish in exploiting UB.

> It's not like C++/WinRT has no UB already.

If that is the case, the reasonable response should be to *remove* the UB, not to add to it.

JaiganeshKumaran:
Putting the following does not work, as only one branch is syntactically valid at a given time. Still think the nullptr trick is the best solution here.
```cpp
 if constexpr (winrt::impl::has_static_lifetime_v<test_component::factory_implementation::StaticClass>)
 {
     winrt::make_self<test_component::factory_implementation::StaticClass>()->Method();
 }
 else
 {
     test_component::factory_implementation::StaticClass::statics_type::Method(); // could cause an error in case the first branch is the chosen one.
 }
 ```

sylveon:
The entire point of if constexpr is allowing the branch not being picked to have invalid code

JaiganeshKumaran:
> The entire point of if constexpr is allowing the branch not being picked to have invalid code

All the branches should still be syntactically valid, unless the expressions are based on a template parameter (see these two examples: https://godbolt.org/z/MPc8KjxWd and https://godbolt.org/z/MPc8KjxWd). In this case I cannot turn those functions into templates, however.

JaiganeshKumaran:
Even without static lifetime, you still may want to put static methods on the factory as static, so that you can have multiple implementations of the implementation type with different names.

Issue #1376, opened 2023-12-14
Bug: TryLookup and Cancel cause noise in debugger because they RoOriginate an error that is expected


antmor:
### Summary

When calling TryLookup, or when cancelling an Async exception, there are a lot of debug spew under a debugger, for example: 

MyDll.dll!00007FFFC056E0E6: ReturnHr(201) tid(8f20) 8000000B The operation attempted to access data outside the valid range
    Msg:[winrt::hresult_error: The operation attempted to access data outside the valid range] 
(8d78.8f20): Windows Runtime Originate Error - code 40080201 (first chance)
(8d78.8f20): Windows Runtime Originate Error - code 40080201 (first chance)
(8d78.8f20): C++ EH exception - code e06d7363 (first chance)
(8d78.8f20): C++ EH exception - code e06d7363 (first chance)
path\to\file.cpp(825)\MyDll.dll!00007FFFC058C923: (caller: 00007FFFC03A4A39) LogHr(139) tid(8f20) 800704C7 The operation was canceled by the user.
    Msg:[winrt::hresult_error: The operation was canceled by the user.] 
(8d78.8f20): Windows Runtime Originate Error - code 40080201 (first chance)
(8d78.8f20): Windows Runtime Originate Error - code 40080201 (first chance)
(8d78.8f20): C++ EH exception - code e06d7363 (first chance)
(8d78.8f20): C++ EH exception - code e06d7363 (first chance)

we should reserve the ability for developers to opt out of this noisy RoOriginate spew in these two scenarios.

PR #1377, opened 2023-12-15
Implement and consume non-originating hresults

antmor:
What:
Address #1376 
- Added new constructors to winrt::hresult_error to opt _out_ of originating the error. 
- Added an optional templated bool to map constructors and factories, to opt in to non-originating E_BOUNDS errors, since these errors are explicitly not helpful.
- To the cancellation_token, added a way to opt in to non-originating errors. 
Also
- added a .runsettings file to add the ability to run the tests from Visual Studio (using a Catch2ToVisualStudioTest adapter)

Why:
When debugging explorer.exe, there are a lot of Cancellations and E_Bounds failures that pop up. To make the debugging experience better, let's opt out of those errors. 

Verified:
ran build_test_all.cmd
Ran tests in Visual Studio:
image

### DIFF
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -77,15 +77,6 @@ namespace winrt::impl
     };
 }
 
-//struct WINRT_IMPL_NOVTABLE IIgnoreNextError : unknown_abi
-//{
-//    virtual void __stdcall IgnoreNextError() noexcept = 0;
-//};
-//
-//template <> inline constexpr guid guid_v<IIgnoreNextError>{ 0x5b0d3235, 0x4dba, 0x4d44, { 0x86,0x5e,0x8f,0x1d,0x0e,0x4f,0xd0,0x4d } };
-//
-
-
 WINRT_EXPORT namespace winrt
 {
     template <typename D, typename T, typename Version = impl::no_collection_version>
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -503,7 +503,7 @@ WINRT_EXPORT namespace winrt
         };
     };
 
-    template <typename D, typename K, typename V, bool ShouldOriginate = true, typename Version = impl::no_collection_version>
+    template <typename D, typename K, typename V, bool avoid_bounds_error_origination = false, typename Version = impl::no_collection_version>
     struct map_view_base : iterable_base<D, Windows::Foundation::Collections::IKeyValuePair<K, V>, Version>
     {
         V Lookup(K const& key) const
@@ -513,7 +513,7 @@ WINRT_EXPORT namespace winrt
 
             if (pair == static_cast<D const&>(*this).get_container().end())
             {
-                if constexpr (ShouldOriginate)
+                if constexpr (avoid_bounds_error_origination)
                 {
                     throw hresult_out_of_bounds();
                 }
@@ -546,8 +546,8 @@ WINRT_EXPORT namespace winrt
 
     };
 
-    template <typename D, typename K, typename V, bool ShouldOriginate = true>
-    struct map_base : map_view_base<D, K, V, ShouldOriginate, impl::collection_version>
+    template <typename D, typename K, typename V, bool avoid_bounds_error_origination = false>
+    struct map_base : map_view_base<D, K, V, avoid_bounds_error_origination, impl::collection_version>
     {
         Windows::Foundation::Collections::IMapView<K, V> GetView() const
         {
@@ -579,7 +579,7 @@ WINRT_EXPORT namespace winrt
             auto found = container.find(static_cast<D const&>(*this).wrap_value(key));
             if (found == container.end())
             {
-                if constexpr (ShouldOriginate)
+                if constexpr (avoid_bounds_error_origination)
                 {
                     throw hresult_out_of_bounds();
                 }
@@ -602,8 +602,8 @@ WINRT_EXPORT namespace winrt
         }
     };
 
-    template <typename D, typename K, typename V, bool ShouldOriginate = true>
-    struct observable_map_base : map_base<D, K, V, ShouldOriginate>
+    template <typename D, typename K, typename V, bool avoid_bounds_error_origination = false>
+    struct observable_map_base : map_base<D, K, V, avoid_bounds_error_origination>
     {
         event_token MapChanged(Windows::Foundation::Collections::MapChangedEventHandler<K, V> const& handler)
         {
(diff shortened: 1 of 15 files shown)


JaiganeshKumaran:
I don't think this is a good solution. You do want the error information when you call Lookup instead of TryLookup, but the latter does not really exist on the type.

Scottj1s:
Generally, the approach here would be to use a policy template type so that there's no runtime impact (storage or execution).  That would also support generality for other projected types.  

E.g.,
single_threaded_map<hstring, int, ignore_errors<hresult_out_of_bounds>>(...)

Can you refactor?

kennykerr:
This seems like another good candidate to [migrate to WIL](https://github.com/microsoft/cppwinrt/pull/1342#issuecomment-1828157945) and let cppwinrt focus on language projection.

Scottj1s:
> This seems like another good candidate to [migrate to WIL](https://github.com/microsoft/cppwinrt/pull/1342#issuecomment-1828157945) and let cppwinrt focus on language projection.

Or at least split the responsibility - factor out an error handling policy template with default here, and let WIL supply specializations and/or subclasses

test/test/async_check_cancel.cpp
```diff
@@ -111,8 +187,10 @@ TEST_CASE("async_check_cancel", "[.clang-crash]")
 TEST_CASE("async_check_cancel")
 #endif
 {
-    Check(Action);
-    Check(ActionWithProgress);
-    Check(Operation);
-    Check(OperationWithProgress);
+    //Check(Action);
```
ChrisGuzak, line 190: what should happen to these commented out checks?

strings/base_error.h
```diff
@@ -206,11 +206,27 @@ WINRT_EXPORT namespace winrt
             originate(code, nullptr WINRT_IMPL_SOURCE_LOCATION_FORWARD);
         }
 
+        explicit hresult_error(hresult const code, bool avoidOrigination WINRT_IMPL_SOURCE_LOCATION_ARGS) noexcept : m_code(verify_error(code))
```
dmachaj, line 209: >WINRT_IMPL_SOURCE_LOCATION_ARGS [](http://example.com/codeflow?start=73&length=31)

If you aren't originating then you might not need source location either

dmachaj: Wait this is a bool not an overload so the originating verison still needs it.

test/test_cpp20/custom_error.cpp
```diff
@@ -64,6 +65,32 @@ TEST_CASE("custom_error_logger")
...
+        REQUIRE(s_loggerCalled);
+
+        // now create an empty map that _doesn't_ log
+        s_loggerCalled = false;
+        IMap<hstring, int> c2 = single_threaded_map<hstring, int>(std::map<hstring, int>{}, true);
+        val = c2.TryLookup(L"hello"sv);
+        REQUIRE(!val);
+        // ensure that the logger wasn't called
+        REQUIRE(!s_loggerCalled);
+    }
```
dmachaj, line 92: This feels like a separate test case instead of a continuation of the existing one.

test/test/async_check_cancel.cpp
antmor: whoops will restore. commented out for easier debugging

strings/base_error.h
```diff
@@ -430,6 +446,18 @@ WINRT_EXPORT namespace winrt
         hresult_canceled(take_ownership_from_abi_t WINRT_IMPL_SOURCE_LOCATION_ARGS) noexcept : hresult_error(impl::error_canceled, take_ownership_from_abi WINRT_IMPL_SOURCE_LOCATION_FORWARD) {}
     };
 
+    struct non_originating_hresult_canceled : hresult_error
```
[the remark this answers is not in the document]
antmor: i think making hresult_error a templated class would break a lot of code that already does 
```
try {}
catch(winrt::hresult_error& e)

```

so that's why i didn't go with that approach.

antmor:
> I don't think this is a good solution. You do want the error information when you call Lookup instead of TryLookup, but the latter does not really exist on the type.

The error for Lookup will still be thrown, and it will still be an E_BOUNDS hresult. The only difference is that it won't be ro-originated, thus it might be a tad harder to debug, but i assume E_BOUNDS is expected for a Lookup to fail, and thus it is not helpful to rooriginate a <not found>. And we would still log _other_ errors outside of the E_BOUNDS.

strings/base_coroutine_threadpool.h
```diff
@@ -180,12 +180,23 @@ WINRT_EXPORT namespace winrt
             return m_propagate_cancellation;
         }
 
+        bool enable_avoid_rooriginate_cancel(bool value) noexcept
```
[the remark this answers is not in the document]
antmor: or disable_rooriginate_on_cancel?

For naming, i want it to be clear that we are opting-out of the roorigination..

kennykerr:
> Or at least split the responsibility - factor out an error handling policy template with default here, and let WIL supply specializations and/or subclasses

Or just use a `WINRT_MIN_ORIGINATE` preprocessor option. 😏

JaiganeshKumaran:
> > I don't think this is a good solution. You do want the error information when you call Lookup instead of TryLookup, but the latter does not really exist on the type.
> 
> The error for Lookup will still be thrown, and it will still be an E_BOUNDS hresult. The only difference is that it won't be ro-originated, thus it might be a tad harder to debug, but i assume E_BOUNDS is expected for a Lookup to fail, and thus it is not helpful to rooriginate a . And we would still log _other_ errors outside of the E_BOUNDS.

Then make it so that throwing E_BOUNDS/hresult_out_of_bounds never originates, instead of adding a separate bool?

JaiganeshKumaran:
@kennykerr Hey Kenny, why don't you check out my pull request (#1375)?

strings/base_collections_base.h
```diff
@@ -536,6 +543,16 @@ WINRT_EXPORT namespace winrt
...
+
+    //    // IIgnoreNextError
+    //    void IgnoreNextError()
+    //    {
+    //        m_ignoreNextError = true;
+    //    }
+    //private:
+    //    bool m_ignoreNextError{ false };
+    protected:
+        bool m_dontOriginateBoundsError{ false };
```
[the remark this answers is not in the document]
antmor: defaulted to always Originate, and you must pass in false to avoid originating.

antmor:
> > > This seems like another good candidate to [migrate to WIL](https://github.com/microsoft/cppwinrt/pull/1342#issuecomment-1828157945) and let cppwinrt focus on language projection.
> >
> > Or at least split the responsibility - factor out an error handling policy template with default here, and let WIL supply specializations and/or subclasses
> 
> Or just use a `WINRT_MIN_ORIGINATE` preprocessor option. 😏

I think we want something more scoped to these scenarios. Having it be all-or-nothing is a heavy hammer. I think it does make sense for cppwinrt to add a hook for both _origination_ and _logging_, such that all HRESULTS will end up asking if there is an winrt_originate_and_log function, and if so, call that, else do a "default". (Right now, DavidMac only added the _logging_ function). But at this function layer, we don't know the source of the hresult, whether it was a map_View or a vector_view or the async task framework, or a e_notimpl from a `.as` call, or an explicit `throw winrt::hresult_error(...);` by consuming code. 
- For TryLookup E_BOUNDS, since TryLookup is a cppwinrt construct, and cppwinrt provides the map_view_base instance, we would have a map_view_rooriginate plugin function explicitly separate from the above. Or we could solve this at the ABI layer and create an interface IAttemptingLookupSuppressErrors that would tell the map we are asking to suppress errors. But at that point, might as well create IMapView2 with TryLookup explicitly. Either of those options are a bigger lift than fixing it at the _implementation_ of map_view_base.
- For Async task cancellation, we could add a hook explicitly for ro-originating cancels.

the other options are for wil to re-implement the whole async framework / map_view_base creation. But those are also heavier lifts.

Issue #1380, opened 2023-12-21
Bug: winrt::hresult_error::originate asserts when using a clr.dll based error info object


ChrisGuzak:
### Version

C++/WinRT v2.0.220418.1

### Summary

In `winrt::hresult_error::originate`, running in the context of a `Microsoft::VisualStudio::CppUnitTestFramework` project, where the error info object is provided by clr.dll and it does not implement `IRestrictedErrorInfo`. the `WINRT_VERIFY(info.try_as(m_info))` line fails the assert.

base.h, ~4892
```cpp
            com_ptr<impl::IErrorInfo> info;
            WINRT_VERIFY_(0, WINRT_IMPL_GetErrorInfo(0, info.put_void()));
            WINRT_VERIFY(info.try_as(m_info));
```

Is this a bug in the CLR or is the design intended to support this?

### Reproducible example

```cpp
TEST_METHOD(CatchCppWinRtException)
{
    try
    {
        winrt::check_hresult(E_NOT_SET);
    }
    catch (...)
    {
    }
}
```

### Expected behavior

no assert, but I'm not sure if this is bug in the CLR or cppwinrt

### Actual behavior

assert, in debug build

jonwis:
Looks like the `hresult_error` case handles this properly:

https://github.com/microsoft/cppwinrt/blob/2511bf7fcb54a64c1e596a950788e2ee30cd1e4c/strings/base_error.h#L217-L221

While the `originate` path does not:

https://github.com/microsoft/cppwinrt/blob/2511bf7fcb54a64c1e596a950788e2ee30cd1e4c/strings/base_error.h#L315-L318

And worse, the `originate` path is triggered in the "the current info does not implement IRestrictedErrorInfo" case of the `hresult_error` ctor.

I'd expect that this is a bad assumption on C++/WinRT's part - there's certainly runtimes that will `SetErrorInfo(pCustomErrorInfo)` that does not implement `IRestrictedErrorInfo` so we should harden the code with that in mind.

kennykerr:
The latter is preceded by a call to `RoOriginateLanguageException` so that assertion should always hold.

jonwis:
[...]

PR #1386, opened 2024-01-16
Fail gracefully when error reporting is suppressed

kennykerr:
Fixes: #1380

### DIFF
--- a/strings/base_error.h
+++ b/strings/base_error.h
@@ -313,8 +313,8 @@ WINRT_EXPORT namespace winrt
             }
 
             com_ptr<impl::IErrorInfo> info;
-            WINRT_VERIFY_(0, WINRT_IMPL_GetErrorInfo(0, info.put_void()));
-            WINRT_VERIFY(info.try_as(m_info));
+            WINRT_IMPL_GetErrorInfo(0, info.put_void());
+            info.try_as(m_info);
         }
 
         static hresult verify_error(hresult const code) noexcept
--- a/test/test/suppress_error_info.cpp
+++ b/test/test/suppress_error_info.cpp
@@ -1,0 +1,17 @@
+#include "pch.h"
+#include <roerrorapi.h>
+
+TEST_CASE("suppress_error_info")
+{
+    winrt::check_hresult(RoSetErrorReportingFlags(RO_ERROR_REPORTING_SUPPRESSSETERRORINFO));
+
+    // Since the error information is suppressed, the best we can hope for is that C++/WinRT
+    // will provide a generic message for the HRESULT.
+    REQUIRE(winrt::hresult_error(E_FAIL, L"message").message() == L"Unspecified error");
+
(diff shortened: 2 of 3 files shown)


jonwis (approved):

strings/base_error.h
```diff
@@ -313,8 +313,8 @@ WINRT_EXPORT namespace winrt
             }
 
             com_ptr<impl::IErrorInfo> info;
-            WINRT_VERIFY_(0, WINRT_IMPL_GetErrorInfo(0, info.put_void()));
-            WINRT_VERIFY(info.try_as(m_info));
+            WINRT_IMPL_GetErrorInfo(0, info.put_void());
```
dmachaj, line 316: `(void)` so it is clear to static analyzers and readers that the return code is intentionally ignored?

kennykerr: Be nice if we could run said static analyzers as part of CI validation. I've never had much success getting Visual Studio's built in static analysis to flag anything in C++/WinRT.

dmachaj: Some internal repos are set up to run Code Analysis during release builds (but not debug builds, because they add a lot of wallclock time).  The warnings/errors show up the same as normal compiler warnings/errors so it is hard to tell whether they are from static analysis or just the normal compiler.  Perhaps the same thing is happening here?

kennykerr: I don't know. I just manually built this code with code analysis and I get nothing:

image

dmachaj: I manually ran an analysis build (using the test_cpp20 project as the thing that pulls in base.h) and it only flagged a single warning using the default "Microsoft Native Recommended Rules".

V:\T\cppwinrt\_build\x64\Debug\winrt\base.h(3373): warning C26800: Use of a moved from object: ''(*<args_1>)'' (lifetime.1).

```cpp
#if __cpp_lib_format >= 202207L
    template <typename... Args>
    inline hstring base_format(Args&&... args)
    {
        auto const size = std::formatted_size(std::forward<Args>(args)...);
        WINRT_ASSERT(size < UINT_MAX);
        auto const size32 = static_cast<uint32_t>(size);

        hstring_builder builder(size32);
        WINRT_VERIFY_(size32, std::format_to_n(builder.data(), size32, std::forward<Args>(args)...).size);
        return builder.to_hstring();
    }
#endif
```

(WINRT_VERIFY_ is line 3373)

dmachaj: I enabled "Run Code Analysis on Build" for the test_cpp20 project and rebuilt and it produced the same output.  It seems to work ok.

(There are also a handful of warnings in catch2 that I didn't copy off and we should probably exclude from consideration)

sylveon: To be fair I was a bit iffy about double forwarding stuff when I originally wrote this, but formatting *should* leave stuff untouched. I just wanted to avoid needing to copy stuff to format the string. Might change it to being `Args const&...` instead. Asking the STL devs for their opinion.

PR #1409, opened 2024-04-05
Test build following dependabot updates

kennykerr:
Dependabot updates seem to have hit this issue:

https://github.com/actions/upload-artifact/issues/478

Let's see if it can recover...

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,4 @@ a dev command prompt at the root of the repo _after_ following the above build i
 * Run `build_prior_projection.cmd` in the dev command prompt as well
 * Run `prepare_versionless_diffs.cmd` which removes version stamps on both current and prior projection
 * Use a directory-level differencing tool to compare `_build\$(arch)\$(flavor)\winrt` and `_reference\$(arch)\$(flavor)\winrt`
+


jsoref:
Correct, ideally your CI and dependabot should have recognized that the workflow requires both download-artifact and upload-artifact to be updated at the same time. Because they weren't, there was no way this could possibly have worked.

Beyond that, in order to upgrade a workflow that uses upload-artifact as part of a matrix, the workflow will need to be reworked (which isn't something dependabot would ever do).

As you're using a matrix, you might have considered:
https://github.com/actions/upload-artifact/blob/main/merge/README.md#examples

But the real culprit is that your matrix has 3 dimensions and your artifact upload/download only use two, as such this job/matrix element 

https://github.com/microsoft/cppwinrt/actions/runs/8570372218/job/23488155109

Performed:
```yml
Run actions/upload-artifact@v4
  with:
    name: msvc-build-x64-Debug-bin
```

While https://github.com/microsoft/cppwinrt/actions/runs/8570372218/job/23488155949#step:11:1 wanted to perform the same.

The "fix" would be to change:
https://github.com/microsoft/cppwinrt/blob/ea187691c1c0d508399c1b069635383082fd9e5b/.github/workflows/ci.yml#L76

```diff
- name: msvc-build-${{ matrix.arch }}-${{ matrix.config }}-bin 
+ name: msvc-build-${{ matrix.compiler }}-${{ matrix.arch }}-${{ matrix.config }}-bin 
```

(You'll have to make the change to each side of upload/download artifacts)

kennykerr:
Thanks! I'm not too familiar with the workflow, so I really appreciate you taking a look. Will give this a try.

jsoref:
I've put some untested changes into https://github.com/check-spelling-sandbox/cppwinrt/tree/unique-matrix-artifacts but it looks like it'd take me a bit to actually trigger stuff...

jsoref:
This run w/ the penciled steps seems to work (at least it gets past the first fault):
https://github.com/check-spelling-sandbox/cppwinrt/actions/runs/8572837925/job/23496284366

This run tries to use the v4 version of things (but I'm going to run into contention as I don't have an infinite supply of runners)
https://github.com/check-spelling-sandbox/cppwinrt/actions/runs/8572877150/job/23496405260

kennykerr:
Feel free to send a PR this way and we can run it.

PR #1410, opened 2024-04-05
Unique matrix artifacts

jsoref:
Fixes #1407
Fixes #1408

This ensures artifacts to be uploaded by matrix jobs are unique (which is more or less a requirement of GitHub's `actions/*-artifact@v4` api).


kennykerr (approved):
Amazing - thanks!

PR #1411, opened 2024-04-05
Use `actions/cache@v4`

jsoref:
`actions/cache@v3` runs on node16 which is deprecated.
Unlike `actions/*-artifact@v4`, `actions/cache@v4` is a drop in upgrade:
https://github.com/actions/cache?tab=readme-ov-file#v4


kennykerr (approved):

kennykerr:
I wonder why dependabot didn't pick this up...

Issue #1422, opened 2024-07-06
Bug: observable_map_base::call_changed is private


Patschkowski:
### Version

C++/WinRT v2.0.220110.5

### Summary

Dear all,

`winrt::observable_vector_base::call_changed` is declared protected, while `winrt::observable_map_base::call_changed` is declared private.

Propose to change the latter to the way it is declared in the former.

Kind regard,
Felix

### Expected behavior

`winrt::observable_map_base::call_changed` should be possible to access from a derived class.

### Actual behavior

Cannot access `winrt::observable_map_base::call_changed` from a derived class.

kennykerr:
@dunhor may be able to comment on visibility.

dunhor:
Going based off source history, it looks like `call_changed` has existed since the very first commits; I did not add them. That aside, I don't see any reason why `call_changed` can't be `protected` in both types.

kennykerr:
Ah, I may have written that after all. 😊

kennykerr:
Feel free to offer a contribution.

PR #1425, opened 2024-07-10
Changed visibility of `observable_map_base::call_changed` to `protected`

Patschkowski:
This matches the visibility of `observable_vector_base::call_changed` to enable derived classes to raise change notifications when the underlying container changes.

Fixes: #1422


kennykerr (approved):
LGTM

PR #1427, opened 2024-07-26
Add missing `<exception>` to base_includes.h

m417z:
Clang with `std=c++23` rightfully complains that `std::exception_ptr` isn't defined.


kennykerr (approved):
LGTM!

PR #1433, opened 2024-09-05
Add NuGet config

manodasanW:
Adding NuGet.config with public feed that upstreams to NuGet.org per guidance.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="CppWinRTDependencies" value="https://pkgs.dev.azure.com/shine-oss/CppWinRT/_packaging/CppWinRTDependencies/nuget/v3/index.json" />
+  </packageSources>
+</configuration>


sylveon:
What about external users? Since this is in the root of the repo, their Visual Studio will end up picking up this feed as well - is this intentional?

manodasanW:
@sylveon the feed is a public externally accessible feed, so external users should have access to it.  Let me know if you don't and I can look into that.

DefaultRyan:
It's also worth noting that all the CI tests passed, and they don't have any priveleged access to internal feeds - the CI tests can only access public feeds just like external users.

DefaultRyan (approved):

sylveon:
Ah, fair enough then

walbourn:
Note that another solution would be to not have a `nuget.config` at all and use in YAML:

```
  - task: NuGetCommand@2
    displayName: NuGet restore
    inputs:
      solution: cppwinrt.sln
      feedRestore: $(GUID_FEED)
      includeNuGetOrg: false
```
   
The `GUID_FEED` is set as a variable so it only gets used by ADO pipelines and not everyone else. GitHub Actions can just use nuget.org directly.

PR #1434, opened 2024-09-10
Update NuGetCommand to use nuget.config

manodasanW:
Previous PR adding nuget.config wasn't enough as a couple nuget restore tasks still used nuget.org.  Updating the tasks to use the nuget.config.

### DIFF
--- a/.pipelines/jobs/OneBranchBuild.yml
+++ b/.pipelines/jobs/OneBranchBuild.yml
@@ -51,6 +51,8 @@ jobs:
       inputs:
         command: 'restore'
         restoreSolution: '$(Build.SourcesDirectory)\cppwinrt.sln'
+        feedsToUse: config
+        nugetConfigPath: NuGet.config
 
     - task: VSBuild@1
       displayName: Build fast_fwd
@@ -72,6 +74,8 @@ jobs:
       inputs:
         command: 'restore'
         restoreSolution: '$(Build.SourcesDirectory)\natvis\cppwinrtvisualizer.sln'
+        feedsToUse: config
+        nugetConfigPath: NuGet.config
         
     - task: VSBuild@1
       displayName: Build Component visualizer
--- a/.pipelines/build.yml
+++ b/.pipelines/build.yml
@@ -49,6 +49,10 @@ jobs:
 
   - task: NuGetCommand@2
     displayName: NuGet restore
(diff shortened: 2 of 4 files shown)


DefaultRyan (approved):

.pipelines/jobs/OneBranchBuild.yml
```diff
@@ -51,6 +51,8 @@ jobs:
       inputs:
         command: 'restore'
         restoreSolution: '$(Build.SourcesDirectory)\cppwinrt.sln'
+        feedsToUse: config
```
walbourn, line 54: From my experience, a better solution is to not have the private ADO feed link in the public repo in a `nuget.config`.

Instead, delete the `nuget.config` from the feed, add a pipeline variable set in the ADO interface that points to the feed, and then have in the ADO YAML:

```
  - task: NuGetCommand@2
    inputs:
      command: 'restore'
      restoreSolution: '$(Build.SourcesDirectory)\cppwinrt.sln'
      feedRestore: $(GUID_FEED)
      includeNuGetOrg: false
```

The only time I've had to use a ``NuGet.config`` for just a source has been doing a custom to use ``nuget install``.

The 'trick' if there any is that the feed has to be described as GUID or rather two GUIDS with the projectid\feedid.

For the URL checked into `nuget.config` here, you'd set the ``GUID_FEED`` variable to ``0e1afd68-1a41-4bd2-9a93-ad91fb9c76d5/2cb8c784-833f-471a-a386-9be37cb4d900``

manodasanW: The feed is a public feed hosted on ADO.

walbourn: It doesn't need to be public, and again if someone submodules this repo and they follow security guidance, they should be using their own feed and NOT use the ADO feed hard-coded in this repo.

manodasanW: You do have a good point about submodules that will need to be evaluated.  And I do agree for public repos, it is not ideal, but the nuget.config is needed as part of one of the current [guidance](https://aka.ms/cfs/nuget).  If you or someone else has a repo that this repo is currently a submodule of, I can look at starting a thread with folks involved with the guidance to get further guidance on how to support it.

PR #1435, opened 2024-09-13
Clarify how InitializeComponent works in the template comment

nthorn:
The default template describes what *not* to do with `InitializeComponent`, but it doesn't describe how developers are supposed to interact with it. This change clarifies the comment so it's more obvious why the link is relevant to the developer.

### DIFF
--- a/vsix/ItemTemplates/BlankPage/BlankPage.h
+++ b/vsix/ItemTemplates/BlankPage/BlankPage.h
@@ -9,6 +9,7 @@ namespace winrt::$rootnamespace$::implementation
         $safeitemname$() 
         {
             // Xaml objects should not call InitializeComponent during construction.
+            // If a Xaml object needs to access a Xaml property during initialization, it should override InitializeComponent.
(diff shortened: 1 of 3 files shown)


Scottj1s (approved):

dmachaj:
As a suggestion, this pattern is also useful for cases where you need to subscribe to callbacks or produce weak references to hand out.  It is usually unsafe to do that during construction but it is safe after the constructor has completed without throwing.

nthorn:
> As a suggestion, this pattern is also useful for cases where you need to subscribe to callbacks or produce weak references to hand out. It is usually unsafe to do that during construction but it is safe after the constructor has completed without throwing.

That's a good callout. In an ideal world, I think it'd be good if the [documentation](https://github.com/microsoft/cppwinrt/tree/master/nuget#initializecomponent) covered that stuff, but I don't think it does. I'd offer to update it myself, but I think that may be an exercise better left to someone with more familiarity with the subject matter.

dmachaj (approved):

PR #1438, opened 2024-10-02
Silence clang-tidy "unused variable" warnings

dunhor:
Clang-tidy (and I'm pretty sure just Clang in general since this is a Clang diagnostic) warns that these variables are unused. This happens because the single threaded collections types use the following lock guard type:
```c++
struct nop_lock_guard {};
```
I could also fix this issue by adding a destructor to the type, however that may have consequences on code-gen. This seemed to be the less risky change with the tradeoff of touching more lines.

### DIFF
--- a/strings/base_collections_base.h
+++ b/strings/base_collections_base.h
@@ -108,7 +108,7 @@ WINRT_EXPORT namespace winrt
         auto First()
         {
             // NOTE: iterator's constructor requires shared access
-            auto guard = static_cast<D&>(*this).acquire_shared();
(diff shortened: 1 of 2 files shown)


sylveon:
Would doing `struct [[nodiscard]] nop_lock_guard {};` silence the warning as well?

dunhor:
> Would doing `struct [[nodiscard]] nop_lock_guard {};` silence the warning as well?

Looks like no: https://godbolt.org/z/YY9vEqh4s

kennykerr (approved):

dmachaj (approved):

PR #1442, opened 2024-10-28
Projected winrt::consume_ methods will nullptr crash if 

### LINKED ISSUE #1441: Bug: Projected winrt::consume_ methods will nullptr crash if the underlying QueryInterface call fails
### Version

2.0.24040

### Summary

The generated projection code for interfaces that the metadata declares as required for a runtimeclass assume that QueryInterface never fails.  Assuming the metadata is correct in the first place, this is a valid assumption for inproc calls.

However, for cross-process calls the QI can fail even if the metadata is correct and the class really does implement all of the required interfaces.  It can fail with E_ACCESSDENIED and a variety of other RPC error codes.

When this happens there is a nullptr crash in the generated consume method.  This can be very painful to debug because the HRESULT is lost by the time it crashes.

### Reproducible example
IDL file:
```
namespace CrashRepro
{
    // This class declares that it implements another interface but under the covers it actually does
    // not.  This allows us to test the behavior when QI's that should not fail, do fail.
    runtimeclass LiesAboutInheritance : Windows.Foundation.IStringable
    {
        void StubMethod();
    }
}
```

main.cpp:
```cpp
#include <winrt/CrashRepro.h>

namespace
{
    struct LiesAboutInheritance : public winrt::implements<LiesAboutInheritance, winrt::CrashRepro::ILiesAboutInheritance>
    {
        LiesAboutInheritance() = default;
        void StubMethod() {}
    };
}

int main()
{
    init_apartment();

    auto lies = winrt::make_self<LiesAboutInheritance>().as<winrt::CrashRepro::LiesAboutInheritance>();
    FAIL_FAST_IF_NULL(lies);
    lies.ToString();     // this line will nullptr crash
}
```

### Expected behavior
Ideally it would throw an invalid cast exception that can be caught/handled.  Less ideally it should fail with a clear signature.

### Actual behavior
There is a nullptr dereference crash.  It has unwound enough before crashing that the HRESULT is permanently lost so it is very difficult to debug.

### Additional comments

_No response_

dmachaj:
The generated projection code for interfaces that the metadata declares as required for a runtimeclass assume that QueryInterface never fails. Assuming the metadata is correct in the first place, this is a valid assumption for inproc calls.

However, for cross-process calls the QI can fail even if the metadata is correct and the class really does implement all of the required interfaces. It can fail with E_ACCESSDENIED and a variety of other RPC error codes.

When this happens there is a nullptr crash in the generated consume method. This can be very painful to debug because the HRESULT is lost by the time it crashes.

This set of changes does not fix the underlying issue.  The program will still crash.  However, it changes the way in which the program crashes so that it is easier to diagnose what went wrong.

The major difference is that the `struct require_one` cast operator now uses a new `as_or_failfast` method instead of `try_as`.  The new `as_or_failfast` method is identical to `try_as` except that it will detect failure and failfast when the QueryInterface call fails.  The code gen comes with a bit of plumbing to generate the new `as_or_failfast` method.

In local testing it was not sufficient to just call `RoFailFastWithErrorContext`.  I had to first call `RoCaptureErrorContext` so that there was stowed context to failfast with.  (It still terminated the process, but the exception record was missing the context).  There is no stub for `RoCaptureErrorContext` so I had to add one.  The macros are a bit unclear on that so I'm _guessing_ that the count is the number of bytes for the argument list.

I also added a new file of test code that exercises this code path.  The test_component IDL declares a runtimeclass that implements `IStringable`.  And then the implementation fails to implement `IStringable`.  When `ToString` is called on this object it hits the failure path.  The cppwinrt code gen will not allow this to happen so I had to directly use `winrt::implements`.

Test code for failfast code paths is tricky.  I used an overload to hook RoFailFastWithErrorContext to use a test stub that confirms it is called.  It doesn't stop execution so the call returns and the original nullptr crash hits the test suite.  I had to use some ugly __try/__except nonsense to eat the nullptr crash so that test execution can continue.  This is a travesty but it also seems to work reliably and I have no better idea as to how to test this effectively.

In the future it would be nice to make this code non-crashing.  There are concerns about binary size if this code path throws exceptions so this is left as an exercise for the future.
 
Besides the new test cases I also wrote a little console app that crashes this way.  I built and ran it using the latest stable cppwinrt as well as my private new one.  As expected the debugger blame is far more useful with these changes.

Before:
`FAILURE_BUCKET_ID:  ACCESS_VIOLATION_c0000005_CppwinrtQIFailureCrashRepro.exe!winrt::impl::consume_Windows_Foundation_IStringable_winrt::CrashRepro::LiesAboutInheritance_::ToString`

After:
`FAILURE_BUCKET_ID:  STOWED_EXCEPTION_80004002_CppwinrtQIFailureCrashRepro.exe!winrt::impl::require_one_winrt::CrashRepro::LiesAboutInheritance,winrt::Windows::Foundation::IStringable_::operator_winrt::Windows::Foundation::IStringable`

Note that the HRESULT ends up in the signature and the blame has the "requires LiesAboutInheritance to implement IStringable" as the method name.

Closes #1441

### DIFF
--- a/strings/base_meta.h
+++ b/strings/base_meta.h
@@ -159,7 +159,7 @@ namespace winrt::impl
     {
         operator I() const noexcept
         {
-            return static_cast<D const*>(this)->template try_as<I>();
+            return static_cast<D const*>(this)->template as_or_failfast<I>();
         }
     };
 
--- a/strings/base_windows.h
+++ b/strings/base_windows.h
@@ -131,6 +131,27 @@ namespace winrt::impl
         ptr->QueryInterface(guid_of<To>(), &result);
         return wrap_as_result<To>(result);
     }
+
+    template <typename To, typename From, std::enable_if_t<is_com_interface_v<To>, int> = 0>
+    com_ref<To> as_or_failfast(From* ptr) noexcept
+    {
+#ifdef WINRT_DIAGNOSTICS
+        get_diagnostics_info().add_query<To>();
+#endif
+
+        if (!ptr)
+        {
+            return nullptr;
+        }
+
+        void* result{};
+        hresult code = ptr->QueryInterface(guid_of<To>(), &result);
+        if (code < 0)
+        {
+            WINRT_IMPL_RoFailFastWithErrorContext(code);
+        }
+        return wrap_as_result<To>(result);
+    }
 }
 
 WINRT_EXPORT namespace winrt::Windows::Foundation
@@ -203,6 +224,12 @@ WINRT_EXPORT namespace winrt::Windows::Foundation
         auto try_as() const noexcept
         {
             return impl::try_as<To>(m_ptr);
+        }
+
+        template <typename To>
+        auto as_or_failfast() const noexcept
+        {
+            return impl::as_or_failfast<To>(m_ptr);
         }
 
         template <typename To>
--- a/strings/base_windows.h
+++ b/strings/base_windows.h
@@ -148,6 +148,10 @@ namespace winrt::impl
         hresult code = ptr->QueryInterface(guid_of<To>(), &result);
         if (code < 0)
         {
+            // Capture an error context immediately before the failfast call.  The failfast gives the best
+            // output when there is a "stowed" exception.  Capturing the error context will stow it enough
+            // for the failfast to find it and use it when creating an exception context record.
+            WINRT_IMPL_RoCaptureErrorContext(code);
             WINRT_IMPL_RoFailFastWithErrorContext(code);
         }
         return wrap_as_result<To>(result);
--- a/test/test/missing_required_interfaces.cpp
+++ b/test/test/missing_required_interfaces.cpp
@@ -1,0 +1,56 @@
+#include "pch.h"
+
+// Unset lean and mean so we can implement a type from the test_component namespace
+#undef WINRT_LEAN_AND_MEAN
+#include <winrt/test_component.h>
+
+namespace
+{
+    struct LiesAboutInheritance : public winrt::implements<LiesAboutInheritance, winrt::test_component::ILiesAboutInheritance>
+    {
+        LiesAboutInheritance() = default;
+        void StubMethod() {}
+    };
+}
+
+bool g_failfastCalled = false;
+
+void __stdcall WINRT_IMPL_RoFailFastWithErrorContext(int32_t) noexcept
+{
+    g_failfastCalled = true;
+}
+
+void DoTheUncatcheable(winrt::test_component::LiesAboutInheritance& lies) noexcept
+{
+    lies.ToString();
+}
+
+bool CatchTheUncatcheable(winrt::test_component::LiesAboutInheritance& lies) noexcept
+{
+    bool caughtCrash = false;
+    __try
+    {
+        // We verify that our replacement version of WINRT_IMPL_RoFailFastWithErrorContext is
+        // called.  In a real program that would exit the process.  But this is a test so it
+        // continues execution and hits a nullptr dereference instead.  Eat that known/expected
+        // error and allow execution to continue onto other test cases.
+        DoTheUncatcheable(lies);
+    }
+    __except (EXCEPTION_EXECUTE_HANDLER)
+    {
+        caughtCrash = true;
+    }
+    return caughtCrash;
+}
+
+TEST_CASE("missing_required_interfaces")
+{
+    auto lies = winrt::make_self<LiesAboutInheritance>().as<winrt::test_component::LiesAboutInheritance>();
+    REQUIRE(lies);
+    REQUIRE_NOTHROW(lies.StubMethod());
+
+    REQUIRE(!g_failfastCalled);
+    const bool caughtCrash = CatchTheUncatcheable(lies);
+    REQUIRE(caughtCrash);
+    REQUIRE(g_failfastCalled);
+}
--- a/strings/base_extern.h
+++ b/strings/base_extern.h
@@ -29,6 +29,7 @@ extern "C"
     int32_t __stdcall WINRT_IMPL_SetThreadpoolTimerEx(winrt::impl::ptp_timer, void*, uint32_t, uint32_t) noexcept WINRT_IMPL_LINK(SetThreadpoolTimerEx, 16);
     int32_t __stdcall WINRT_IMPL_SetThreadpoolWaitEx(winrt::impl::ptp_wait, void*, void*, void*) noexcept WINRT_IMPL_LINK(SetThreadpoolWaitEx, 16);
     int32_t __stdcall WINRT_IMPL_RoOriginateLanguageException(int32_t error, void* message, void* exception) noexcept WINRT_IMPL_LINK(RoOriginateLanguageException, 12);
+    int32_t __stdcall WINRT_IMPL_RoCaptureErrorContext(int32_t error) noexcept WINRT_IMPL_LINK(RoCaptureErrorContext, 4);
     void __stdcall WINRT_IMPL_RoFailFastWithErrorContext(int32_t) noexcept WINRT_IMPL_LINK(RoFailFastWithErrorContext, 4);
     int32_t __stdcall WINRT_IMPL_RoTransformError(int32_t, int32_t, void*) noexcept WINRT_IMPL_LINK(RoTransformError, 12);
 
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -788,6 +788,12 @@ namespace winrt::impl
             return m_inner.try_as<Qi>();
         }
 
+        template <typename Qi>
+        auto as_or_failfast() const noexcept
+        {
+            return m_inner.as_or_failfast<Qi>();
+        }
+
         explicit operator bool() const noexcept
         {
             return m_inner.operator bool();
--- a/test/test/test.vcxproj
+++ b/test/test/test.vcxproj
@@ -387,6 +387,7 @@
       <PrecompiledHeader>NotUsing</PrecompiledHeader>
     </ClCompile>
     <ClCompile Include="memory_buffer.cpp" />
+    <ClCompile Include="missing_required_interfaces.cpp" />
     <ClCompile Include="module_lock_dll.cpp">
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">NotUsing</PrecompiledHeader>
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">NotUsing</PrecompiledHeader>
--- a/test/test_component/test_component.idl
+++ b/test/test_component/test_component.idl
@@ -163,6 +163,13 @@ namespace test_component
         static void StaticMethodWithAsyncReturn();
     }
 
+    // This class declares that it implements another interface but under the covers it actually does
+    // not.  This allows us to test the behavior when QI's that should not fail, do fail.
+    runtimeclass LiesAboutInheritance : Windows.Foundation.IStringable
+    {
+        void StubMethod();
+    }
+
     namespace Structs
     {
         struct All


strings/base_meta.h
```diff
@@ -159,7 +159,7 @@ namespace winrt::impl
     {
         operator I() const noexcept
         {
-            return static_cast<D const*>(this)->template try_as<I>();
+            return static_cast<D const*>(this)->template as_or_failfast<I>();
```
jonwis, line 162: This feels like a pretty big change. I think this means code like `auto stringable = static_cast<IStringable>(lies)` will crash? That's a legit way of probing the type hierarchy, like this:

```c++
if (auto stringable = static_cast<IStringable>(lies)) {
    stringable.ToString();
}
```

... but yes it's strange that `static_cast<>` here can "fail."

See WINRT_IMPL_SHIM - that's the place that should get this fix.  If this was limited to the cast-shim helper on the projected runtimeclass type it might be better, since it's no "worse" than we have today (and in fact better since it buckets more nicely.)

Note @oldnewthing 's comment though - this means all the `foo.Bar(); foo.Bas()` that use the shim now get additional code generated to check the hresult and throw/failfast.

I suggest rebuilding something meaty internally with this change in place (Baker, etc.) and see what the codegen looks like using sizebench.

test/test/missing_required_interfaces.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+bool g_failfastCalled = false;
+
+void __stdcall WINRT_IMPL_RoFailFastWithErrorContext(int32_t) noexcept
+{
+    g_failfastCalled = true;
+}
+
+void DoTheUncatcheable(winrt::test_component::LiesAboutInheritance& lies) noexcept
```
arkrumbe, line 23: nit: spelling `uncatchable` in both methods

dmachaj: Now that this throws instead of failfasting that stuff is deleted, typo and all.

dmachaj:
As a binary size test case I built a private copy of WinUI3.  The binary with cppwinrt usage is Microsoft.UI.Xaml.Controls.dll.  The binary size for that actually went down (not up).  5,775,872 to 5,636,096 (-139,776 bytes aka -2.4%).

kennykerr:
> As a binary size test case I built a private copy of WinUI3. The binary with cppwinrt usage is Microsoft.UI.Xaml.Controls.dll. The binary size for that actually went down (not up). 5,775,872 to 5,636,096 (-139,776 bytes aka -2.4%).

Compared to what? The previous commit in the cppwinrt repo or some previous release?

dmachaj:
> > As a binary size test case I built a private copy of WinUI3. The binary with cppwinrt usage is Microsoft.UI.Xaml.Controls.dll. The binary size for that actually went down (not up). 5,775,872 to 5,636,096 (-139,776 bytes aka -2.4%).
> 
> Compared to what? The previous commit in the cppwinrt repo or some previous release?

Compared to `master`.  (Actually, not _literally_ master.  I had to deactivate the breaking changes from #1319 before I could compile XAML with a newer cppwinrt.).  The only difference between the two binaries under comparison is the changes in this PR.

kennykerr:
Technically the FH4 work that Raymond alluded to is about folding identical catch handlers rather than call sites. A reduction here for `Microsoft.UI.Xaml.Controls.dll` would seem to imply that code that was previous inlined is no longer.

dmachaj:
With the latest iteration (with the fix to keep the lifetime valid on the casted-to object) the binary size difference is close to zero.  With the changes in this PR the XAML controls DLL is 10KB smaller than before (well below 1% diff).

jonwis:
> Technically the FH4 work that Raymond alluded to is about folding identical catch handlers rather than call sites. A reduction here for `Microsoft.UI.Xaml.Controls.dll` would seem to imply that code that was previous inlined is no longer.

In prior investigations, I there are types that have pathological inline behavior. [Types like ValueSet](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.collections.valueset?view=winrt-26100), for instance, don't have any methods of their own - it's a flavored IMap<>.  Someone accidentally doing the obvious thing:

```c++
ValueSet v{};
v.Insert(L"kittens", winrt::box_value(7u));
v.Insert(L"puppies", winrt::box_value(2.0f));
v.Insert(L"muffins", winrt::box_value(L"blueberry"));
```

... ended up with three _inlined_ QueryInterface calls from `IValueSet` -> `IMap<String, Object>` in order to call `IMap<>::Insert`.  It's not clear why the optimizer/inliner chose to splat the sequence of "load vtable, load queryinterface, push address of IMap<>, call" instead of leaving that as "invoke consume_ValueSet_IMapInsert" method calls, but it was _very_ noticeable in the sizebench builds (building the same way as our release branch build builds.)

Any way to either merge those three QI sequences (tear off interface caching?) or at least not inline them (so the QI + call Insert is an invocable thing) would produce really great size enhancements. The code's already doing a bunch of "call" anyhow, we aren't saving anything by inlining the QI+Call into the parent frame.

jonwis:
> As a binary size test case I built a private copy of WinUI3. The binary with cppwinrt usage is Microsoft.UI.Xaml.Controls.dll. The binary size for that actually went down (not up). 5,775,872 to 5,636,096 (-139,776 bytes aka -2.4%).

Make sure to also run the validation with the "real build line" - things like PGO can be super duper smart and decide to either inline or fold based on "other data."  I tried fixing winrt::hstring to be less inlined only to find that the PGO optimizer was going to already de-inline all the calls on its own.

kennykerr:
Before I worked on FH4, I worked with MSVC on an optimization to treat certain calls to `QueryInterface` as [pure functions](https://en.wikipedia.org/wiki/Pure_function). This worked well for solving such problems but we could never get the optimizer to hoist these calls reliably. Eventually I started working on the fast abi, which achieved the same effect but sadly was never adopted. 

Ultimately, types like `ValueSet` are very poorly designed and should not be used. In the early days of cppwinrt I had diagnostics that would print out debug warnings at runtime for such pathological cases but it was deemed too judgy. 😉

strings/base_error.h
```diff
@@ -529,6 +529,28 @@ WINRT_EXPORT namespace winrt
         return pointer;
     }
 
+    template <typename T>
+    WINRT_IMPL_NOINLINE void check_cast_result(T* from WINRT_IMPL_SOURCE_LOCATION_ARGS)
```
kennykerr, line 533: Seems like this should be in the `impl` namespace.

strings/base_windows.h
```diff
@@ -128,7 +128,11 @@ namespace winrt::impl
         }
 
         void* result{};
-        ptr->QueryInterface(guid_of<To>(), &result);
+        hresult code = ptr->QueryInterface(guid_of<To>(), &result);
+        if (code < 0)
+        {
+            WINRT_IMPL_RoCaptureErrorContext(code);
```
kennykerr, line 134: Is this still needed now that `WINRT_IMPL_SHIM` isn't being used above?

cppwinrt/code_writers.h
```diff
@@ -1148,7 +1154,10 @@ namespace cppwinrt
         {
             format = R"(    template <typename D%> auto consume_%<D%>::%(%) const
     {%
-        check_hresult(WINRT_IMPL_SHIM(%)->%(%));%
+        const auto castedResult = static_cast<% const&>(static_cast<D const&>(*this));
+        const auto abiType = *(abi_t<%>**)&castedResult;
+        check_cast_result(abiType);
+        check_hresult(abiType->%(%));%
     }
```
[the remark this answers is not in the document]
kennykerr: You may also want to confirm (manually) that this still destroys the temporary. Its very easy to inject extra QIs or drop needed dtors in crazy code like this. I speak from experience. 😉

strings/base_windows.h
dmachaj: Yes it is needed.  There is no HRESULT return from this method so the COM error context is used to propagate the failure code upwards to `check_cast_result`.

cppwinrt/code_writers.h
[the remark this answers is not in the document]
dmachaj: I honestly think the verbosity is a good thing, not a bad thing.  I've been debugging and bottomed out at the `WINRT_IMPL_SHIM` macro many times and it has always confused me.  I had to stare at it for a good 20 minutes yesterday, with help from @jonwis, and it's still difficult for me to parse.  I think splitting out the two sets of casts into separate lines outside a macro makes the code much more explicit/debuggable for mortals.  It also makes space to set breakpoints.  I don't think that conciseness of generated code that humans don't edit is a top concern (within reason).

As for the temporary, what do you think the best way would be to check that?  This is keeping the winrt projection object alive through the end of the function so the lifetime should be fine (including destruction).  I had to expand this out because previously the temporary didn't outlive the macro so the calls after `WINRT_IMPL_SHIM` were using freed memory.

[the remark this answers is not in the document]
kennykerr: I'd just want to confirm that:
* you're not getting a `QueryInterface` in cases where one isn't needed
* you're getting the right number of `Release` calls

The former probably requires a breakpoint or `printf` inside the `QueryInterface` and the latter can (cheaply) be stress-tested in the scratch project with a loop and observing that memory isn't being leaked.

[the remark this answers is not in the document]
dmachaj: I added a little test program to the scratch project.  When it is waiting for input before exiting the memory usage is very low (1MB active memory in task manager; close to the lowest on my system).  I think that confirms there is no leak.

I set breakpoints in QI, AddRef, and Release.  When this loop is running there is one AddRef for the initial cast.  There is one Release when the consume_ method returns.  This matches expectations.  There is no QI call in this case because it is already the correct type.

```cpp
namespace
{
    struct BasicStringable : public winrt::implements<BasicStringable, winrt::Windows::Foundation::IStringable>
    {
        BasicStringable() = default;
        hstring ToString()
        {
            return L"A string!";
        }
    };
}

int main()
{
    winrt::init_apartment();

    auto stringable = winrt::make_self<BasicStringable>().as<winrt::Windows::Foundation::IStringable>();
    for (auto i = 0U; i < 1000000; ++i)
    {
        const auto result = stringable.ToString();
    }

    printf("Press any key to exit");
    int num{};
    scanf("%d", &num);
}
```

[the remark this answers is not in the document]
kennykerr: For the memory leak check, you'd want to move the "make" into the loop to put all the pressure on allocation rather than only reference counting. For example:

```C++
using namespace winrt;

struct Thing : implements<Thing, IClosable>
{
    void Close() {}
};

int main() {
    for (;;)
    {
        auto thing = make<Thing>();
        //detach_abi(thing);
    }
}
```

Then just run in the debugger and look for this flat graph:

image

If i looks more like this then you have a problem: 

image

You can uncomment the `detach_abi` call to simulate.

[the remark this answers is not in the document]
dmachaj: It is flat :)

image

That is with the make_self call moved inside the loop.  I have not been able to get the scratch project to build projections from an IDL file so I can't easily test the case where the consume_ call requires a QI.  I can't get the scratch project to build an IDL at all (I've been trying for about an hour).

strings/base_error.h
dmachaj: Moved to impl namespace.

cppwinrt/code_writers.h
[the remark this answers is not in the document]
kennykerr: Just rebuild the test projection with `build_projection.cmd` and then use a stock implementation like `Uri`.

kennykerr:
image

🤣🤣 vcxproj files are so awful.

cppwinrt/code_writers.h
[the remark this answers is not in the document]
dmachaj: I realized after entirely too long that I don't need the scratch project for this.  I already have a little throwaway console app that builds just fine.  I made the changes there.

I'm observing that for the case where there is a QI in this code path there is a spurious reference.  The QI adds one reference.  And then a second addref is also triggered.

image

The proper number of Release calls are made so this does not leak, at least.  (I'll need to compare against the baseline to see if this is new from my changes or not).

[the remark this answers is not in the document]
dmachaj: Alas, the baseline does not have that extra ref:

image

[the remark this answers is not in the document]
dmachaj: That was building unoptimized.  When I build release everything is aggressively inlined into the containing code and there does not seem to be extra references:

image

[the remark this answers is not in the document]
kennykerr: It should definitely be unchanged in unoptimized builds.

cppwinrt/code_writers.h
```diff
@@ -1131,15 +1131,21 @@ namespace cppwinrt
...
+        abiType->%(%);%
     }
 )";
             }
             else
             {
                 format = R"(    template <typename D%> auto consume_%<D%>::%(%) const noexcept
     {%
-        WINRT_VERIFY_(0, WINRT_IMPL_SHIM(%)->%(%));%
+        const auto castedResult = static_cast<% const&>(static_cast<D const&>(*this));
```
DefaultRyan, line 1145: I believe this creates a copy and refcount bump.
```suggestion
        auto const& castedResult = static_cast<% const&>(static_cast<D const&>(*this));
```

cppwinrt/code_writers.h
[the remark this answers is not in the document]
dmachaj: Thanks @DefaultRyan for pointing out that making castedResult a reference will fix the extra copy.  I have confirmed that in Debug builds the extra copy is gone.  And for Release there is no leak.

Debug:
image

Release:
image

cppwinrt/code_writers.h
dmachaj: Thanks for the fix 👍.

I did `const auto&` but let me know if you think `auto const&` more closely matches the style of this project.  I believe they are equally correct.

test/test/missing_required_interfaces.cpp
```diff
@@ -0,0 +1,25 @@
...
+TEST_CASE("missing_required_interfaces")
+{
+    auto lies = winrt::make_self<LiesAboutInheritance>().as<winrt::test_component::LiesAboutInheritance>();
+    REQUIRE(lies);
+    REQUIRE_NOTHROW(lies.StubMethod());
+
+    // The IStringable::ToString method does not exist on this type.  In previous versions of cppwinrt
+    // this line would crash with a nullptr deference.  It now throws an exception.
+    REQUIRE_THROWS_AS(lies.ToString(), winrt::hresult_error);
+}
```
DefaultRyan, line 25: Nit: missing newline

cppwinrt/code_writers.h
kennykerr: cppwinrt uses east const.

dmachaj: Updated to code gen east const.

cppwinrt/code_writers.h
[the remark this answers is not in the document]
dmachaj: Marking this thread as resolved.  I think the code gen is now confirmed as acceptable.  LMK if you disagree.

dmachaj:
With the latest iteration of code Microsoft.UI.Xaml.Controls.dll is 6,634,560 bytes (-141,312bytes, -2.4%, compared to baseline).  That is back to where it was with the failfast approach.

jonwis (approved):

cppwinrt/code_writers.h
```diff
@@ -1129,17 +1129,27 @@ namespace cppwinrt
...
+                //
+                // The `noexcept` versions will crash if check_cast_result throws but that is no different than previous
+                // behavior where it would not check the cast result and nullptr crash.  At least the exception will terminate
+                // immediately while preserving the error code and local variables.
                 format = R"(    template <typename D%> auto consume_%<D%>::%(%) const noexcept
     {%
-        WINRT_IMPL_SHIM(%)->%(%);%
+        auto const& castedResult = static_cast<% const&>(static_cast<D const&>(*this));
+        auto const abiType = *(abi_t<%>**)&castedResult;
+        check_cast_result(abiType);
```
kennykerr, line 1140: I would just point out that this check is done even when no QI has occurred. I think its reasonable as the check short-circuits quickly but its worth noting.

kennykerr (approved):

cppwinrt/code_writers.h
dmachaj: That's fair.  In the common case where the cast succeeds (or there was no cast at all) it is a single nullptr check that gets inlined into the calling code.  This seems like a reasonable cost.  If evidence arises that this is too much overhead then we can always revisit (perhaps with an `if constexpr` to detect if the QI is going to be needed?).

kennykerr: I would be surprised if a decent optimizer doesn't eliminate this.

strings/base_error.h
```diff
@@ -546,6 +546,28 @@ namespace winrt::impl
         }
         return result;
     }
+
+    template <typename T>
+    WINRT_IMPL_NOINLINE void check_cast_result(T* from WINRT_IMPL_SOURCE_LOCATION_ARGS)
```
oldnewthing, line 551: Doesn't need to be a template. Can just take a `void*`. This allows the different instantiations to be folded.

oldnewthing, line 551: Doesn't this have to be `WINRT_IMPL_NOINLINE inline` to avoid ODR problems?

strings/base_error.h
```diff
@@ -546,6 +546,28 @@ namespace winrt::impl
...
+                impl::bstr_handle description;
+                impl::bstr_handle restrictedDescription;
+                impl::bstr_handle capabilitySid;
+                if (restrictedError->GetErrorDetails(description.put(), &code, restrictedDescription.put(), capabilitySid.put()) == 0)
+                {
+                    throw hresult_error(code, take_ownership_from_abi WINRT_IMPL_SOURCE_LOCATION_FORWARD);
+                }
+            }
+        }
+    }
```
oldnewthing, line 570: I'm wondering if code gen would be better as
```cpp
WINRT_IMPL_NOINLINE [[noreturn]] void throw_failed_cast(hresult code WINRT_IMPL_SOURCE_LOCATION_ARGS)
{
    throw hresult_error(code, take_ownership_from_abi WINRT_IMPL_SOURCE_LOCATION_FORWARD);
}

inline void check_cast_result(hresult code WINRT_IMPL_SOURCE_LOCATION_ARGS)
{
    if (code < 0) throw_failed_cast(code WINRT_IMPL_SOURCE_LOCATION_FORWARD);
}
```
where we have a new try_as_reason that returns both a pointer and an hresult.
```cpp
void* result{};
hresult code = ptr->QueryInterface(guid_of<To>(), &result);
return { wrap_as_result<To>(result), code };
```

strings/base_error.h
dmachaj: What sort of ODR problems?  The only variance based on compiler options is the SOURCE_LOCATION macro and that already has global ODR guards.

The `noinline` is important because it is how the code bloat is avoided by having lots of exception throwing locations spewed out into containing code.

dmachaj: That's fair.  If I spin another PR I'll make that change.

PR #1444, opened 2024-11-01
WINRT_SOURCE_LOCATION has ODR checks that prevent mixing cpp17 and cpp20 static libraries

dmachaj:
Someone is trying to upgrade the cppwinrt version used by a large project that has many static libraries using cppwinrt.  Some binaries in that project are mixing static libs with different cpp language versions.  (This seems like not a great idea generally but it is the state of the world so to some degree we have to live with it).  In those binaries the ODR checks for cppwinrt source_location usage are breaking the build.  For good reason, it is not safe to mix this functionality across language versions.  This set of changes is aimed at making that upgrade process easier without losing any useful functionality.

We already have `winrt::impl::slim_source_location` which is a lot like `std::source_location`, minus always containing the very impactful FUNCTION data.  This type is powered by the same intrinsics as the STL version so it works as well as the STL library.  The existence of this class means that we can avoid the ODR violations by always using winrt::impl::slim_source_location.

Some new macros are used to control what goes into the constructor.  cpp20 code that does not suppress source_location will get valid source information passed in.  Code that is cpp17, or suppresses this feature, will have zero's passed in.  Furthermore, when compiling as _DEBUG this will also include the FUNCTION data, matching previous behavior.

The net result is that there is no more ODR violation because the function signatures are always the same.

For now I have elected not to clean up the `WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT` macro family.  There is only one option so there is no more ODR violation.  But removing that would churn a lot of code.  I could be convinced to remove it now but only if there is agreement it is worth it and that we won't try to re-add variance later.

`build_test_all.cmd` for both release and debug.  I also created a little project that mixes cpp17 and cpp20 libraries.  With the latest public release of cppwinrt this project does not build because of the ODR violations.  With these changes it builds and runs as expected.

build break with existing cppwinrt
```
3>Cpp17StaticLib.lib(Cpp17StaticLib.obj) : error LNK2038: mismatch detected for 'WINRT_SOURCE_LOCATION': value 'false' doesn't match value 'slim' in main.ob
```

cpp17 code
```cpp
void Cpp17StaticLib()
{
    try
    {
        winrt::check_hresult(E_FAIL);
    }
    catch (const winrt::hresult_error& e)
    {
        wprintf(L"C++17 Static Lib: %ws\n", e.message().c_str());
    }
}
```

cpp20 code
```cpp
void Cpp20StaticLib()
{
    try
    {
        winrt::check_hresult(E_FAIL);
    }
    catch (const winrt::hresult_error& e)
    {
        wprintf(L"C++20 Static Lib: %ws\n", e.message().c_str());
    }
}
```

console exe code
```cpp

void __stdcall MY_winrt_throw_hresult_handler(uint32_t lineNumber, char const* fileName, char const* functionName, void* returnAddress, winrt::hresult const result) noexcept
{
    wprintf(L"winrt::hresult_error thrown at line %u in %hs in %hs with error code 0x%08X\n", lineNumber, fileName, functionName, (uint32_t)result);
}

int main()
{
    winrt::init_apartment();

    winrt_throw_hresult_handler = MY_winrt_throw_hresult_handler;

    Cpp17StaticLib();
    Cpp20StaticLib();
}
```

Release output:
```
winrt::hresult_error thrown at line 0 in (null) in (null) with error code 0x80004005
winrt::hresult_error thrown at line 0 in (null) in (null) with error code 0x80004005
C++17 Static Lib: Unspecified error
winrt::hresult_error thrown at line 7 in %PATH%Cpp20StaticLib\Cpp20StaticLib.cpp in (null) with error code 0x80004005
winrt::hresult_error thrown at line 7 in %PATH%\Cpp20StaticLib\Cpp20StaticLib.cpp in (null) with error code 0x80004005
C++20 Static Lib: Unspecified error
```

Debug output:
```
winrt::hresult_error thrown at line 0 in (null) in (null) with error code 0x80004005
winrt::hresult_error thrown at line 0 in (null) in (null) with error code 0x80004005
C++17 Static Lib: Unspecified error
winrt::hresult_error thrown at line 7 in %PATH%\Cpp20StaticLib\Cpp20StaticLib.cpp in Cpp20StaticLib with error code 0x80004005
winrt::hresult_error thrown at line 7 in %PATH%\Cpp20StaticLib\Cpp20StaticLib.cpp in Cpp20StaticLib with error code 0x80004005
C++20 Static Lib: Unspecified error
```

### DIFF
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -105,7 +105,7 @@ typedef struct _GUID GUID;
 
 #if !defined(__cpp_lib_source_location) || defined(WINRT_NO_SOURCE_LOCATION)
 // Case1: cpp17 mode.  The source_location intrinsics are not available.
-// Cas2: The caller has disabled source_location support.  Ensure that there is no binary size overhead for line/file/function.
+// Case2: The caller has disabled source_location support.  Ensure that there is no binary size overhead for line/file/function.
 #define WINRT_IMPL_BUILTIN_LINE 0
 #define WINRT_IMPL_BUILTIN_FILE nullptr
 #define WINRT_IMPL_BUILTIN_FUNCTION nullptr
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -77,6 +77,12 @@ struct IUnknown;
 typedef struct _GUID GUID;
 #endif
 
+#if defined(__cpp_consteval)
+#define WINRT_IMPL_CONSTEVAL consteval
+#else
+#define WINRT_IMPL_CONSTEVAL
+#endif
+
 // std::source_location is a C++20 feature, which is above the C++17 feature floor for cppwinrt.  The source location needs
 // to be the calling code, not cppwinrt itself, so that it is useful to developers building on top of this library.  As a
 // result any public-facing method that can result in an error needs a default-constructed source_location argument.  Because
@@ -84,7 +90,17 @@ typedef struct _GUID GUID;
 //
 // Some projects may decide to disable std::source_location support to prevent source code information from ending up in their
 // release binaries, or to reduce binary size.  Defining WINRT_NO_SOURCE_LOCATION will prevent this feature from activating.
-#if defined(__cpp_lib_source_location) && !defined(WINRT_NO_SOURCE_LOCATION)
+#if !defined(WINRT_NO_SOURCE_LOCATION)
+
+// Some projects mix static libs compiled as cpp17 and cpp20 into the same binary.  The source_location ODR check creates
+// (legitimate) build breaks in those projects because it is not safe to mix the two.  To ease the compatibility burden on
+// these projects updating to a newer cppwinrt we can define the builtin's to be empty in cpp17 mode.  This does not provide
+// the correct data, but it does avoid the ODR violations.  winrt::impl::slim_source_location can serve both modes.
+#if !defined(__cpp_lib_source_location)
+#define __builtin_LINE() 0
+#define __builtin_FILE() nullptr
+#define WINRT_IMPL_EMPTY_SOURCE_LOCATION
+#endif
 
 namespace winrt::impl
 {
@@ -93,7 +109,7 @@ namespace winrt::impl
     // have heavy binary size overhead when templates cause many permutations to exist.
     struct slim_source_location
     {
-        [[nodiscard]] static consteval slim_source_location current(
+        [[nodiscard]] static WINRT_IMPL_CONSTEVAL slim_source_location current(
             const std::uint_least32_t line = __builtin_LINE(),
             const char* const file = __builtin_FILE()) noexcept
         {
@@ -136,7 +152,7 @@ namespace winrt::impl
 // usefulness of source_location while having a much smaller binary impact.
 //
 // When building _DEBUG binary size is not usually much of a concern, so we can use the full source_location type.
-#ifdef _DEBUG
+#if defined(_DEBUG) && !defined(WINRT_IMPL_EMPTY_SOURCE_LOCATION)
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT , std::source_location const& sourceInformation
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS , std::source_location const& sourceInformation = std::source_location::current()
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM std::source_location const& sourceInformation = std::source_location::current()
@@ -150,7 +166,7 @@ namespace winrt::impl
 #pragma detect_mismatch("WINRT_SOURCE_LOCATION", "true")
 #endif // _MSC_VER
 
-#else // !_DEBUG
+#else // !_DEBUG || WINRT_IMPL_EMPTY_SOURCE_LOCATION
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT , winrt::impl::slim_source_location const& sourceInformation
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS , winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()
@@ -164,7 +180,7 @@ namespace winrt::impl
 #pragma detect_mismatch("WINRT_SOURCE_LOCATION", "slim")
 #endif // _MSC_VER
 
-#endif // _DEBUG
+#endif // _DEBUG / WINRT_IMPL_EMPTY_SOURCE_LOCATION
 
 #else
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -102,6 +102,12 @@ typedef struct _GUID GUID;
 // To help with debugging the __builtin_FUNCTION() intrinsic will be used in _DEBUG builds.  This will provide a bit more diagnostic
 // value at the cost of binary size.  The assumption is that binary size is considered less important in debug builds so this tradeoff
 // is acceptable.
+//
+// The different behavior of the default parameters to winrt::impl::slim_sourc_location::current() is technically an ODR violation,
+// albeit a minor one.  There should be no serious consequence to this violation.  In practice it means that mixing cpp17/cpp20,
+// or mixing WINRT_NO_SOURCE_LOCATION with undefining it, will lead to inconsistent source location information.  It may be missing
+// when it is expected to be included, or it may be present when it is not expected.  The behavior will depend on the linker's choice
+// when there are multiple translation units with different options.  This violation is tracked by https://github.com/microsoft/cppwinrt/issues/1445.
 
 #if !defined(__cpp_lib_source_location) || defined(WINRT_NO_SOURCE_LOCATION)
 // Case1: cpp17 mode.  The source_location intrinsics are not available.
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -103,12 +103,8 @@ typedef struct _GUID GUID;
 // value at the cost of binary size.  The assumption is that binary size is considered less important in debug builds so this tradeoff
 // is acceptable.
 
-#if !defined(__cpp_lib_source_location)
+#if !defined(__cpp_lib_source_location) || defined(WINRT_NO_SOURCE_LOCATION)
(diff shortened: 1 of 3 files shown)


jonwis (approved):

strings/base_macros.h
```diff
@@ -93,19 +132,23 @@ namespace winrt::impl
     // have heavy binary size overhead when templates cause many permutations to exist.
     struct slim_source_location
     {
-        [[nodiscard]] static consteval slim_source_location current(
-            const std::uint_least32_t line = __builtin_LINE(),
-            const char* const file = __builtin_FILE()) noexcept
+        [[nodiscard]] static WINRT_IMPL_CONSTEVAL slim_source_location current(
```
dmachaj, line 135: As best as I can tell this approach resolves to the correct thing because the preprocessor is doing the work.  Even translation units in the same library are compatible with different flags.  I added another test case to my little app and it has the expected output:

(This code is in the cpp20 static lib alongside the previously shared sample)
```cpp
```
#define WINRT_NO_SOURCE_LOCATION
#include <winrt/base.h>

void Cpp20StaticLibSuppressSourceLocation()
{
    try
    {
        winrt::check_hresult(E_FAIL);
    }
    catch (const winrt::hresult_error& e)
    {
        wprintf(L"C++20 Static Lib (WINRT_NO_SOURCE_LOCATION): %ws\n", e.message().c_str());
    }
}
```
winrt::hresult_error thrown at line 8 in V:\T\davidmac.randomprojects\CppWinrtMixCpp17AndCpp20\Cpp20StaticLib\Cpp20StaticLib.cpp in Cpp20StaticLib with error code 0x80004005
winrt::hresult_error thrown at line 8 in V:\T\davidmac.randomprojects\CppWinrtMixCpp17AndCpp20\Cpp20StaticLib\Cpp20StaticLib.cpp in Cpp20StaticLib with error code 0x80004005
C++20 Static Lib: Unspecified error
winrt::hresult_error thrown at line 0 in (null) in (null) with error code 0x80004005
winrt::hresult_error thrown at line 0 in (null) in (null) with error code 0x80004005
C++20 Static Lib (WINRT_NO_SOURCE_LOCATION): Unspecified error
```

DefaultRyan: The C++ standard says ...

> There can be more than one definition in a program of [...] inline function [...] as long as all of the following is true:
> * each definition appears in a different translation unit
> * each definition consists of the same sequence of tokens
> * [... more rules]
> * the rules above apply to every default argument used in each definition

My understanding is that by conditionally changing the *default* arguments in the definition of `slim_source_location::current()`, we are introducing an ODR violation on that function - there are two definitions of the function with different default argument values.

It might be a relatively tame violation, because we're not changing the type of the params, nor which params have default values. In practice, it's probably common that we see those args evaluated using the declaration currently visible to that translation unit. But in the presence of things like whole-program optimization, I don't know for certain if the evaluation of default args can be deferred to link time, at which point we can't predict which version is used.

strings/base_macros.h
```diff
@@ -118,39 +161,18 @@ namespace winrt::impl
...
-#define WINRT_SOURCE_LOCATION_ACTIVE
-
-#ifdef _MSC_VER
-#pragma detect_mismatch("WINRT_SOURCE_LOCATION", "true")
-#endif // _MSC_VER
-
-#else // !_DEBUG
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_NO_DEFAULT , winrt::impl::slim_source_location const& sourceInformation
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS , winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()
 #define WINRT_IMPL_SOURCE_LOCATION_ARGS_SINGLE_PARAM winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()
```
DefaultRyan, line 178: Since this no longer evals to a no-op, we technically don't need the `_SINGLE_PARAM`, which only exists to remove the comma. We could eliminate that variant and explicitly supply the comma in the callsites. Not required, but something to consider.

DefaultRyan: Minimizing churn sounds fine to me.

strings/base_macros.h
dmachaj: I tried your suggestion of replacing `check_hresult` with `winrt::Windows::Foundation::Uri(L"InvalidUri");` and that does not behave well.  The cpp17 library wins the linker battle and the source location is always empty.

strings/base_macros.h
sylveon: I've never like these macros so I'd be in favor of removing them even if it causes a bit of churn.

ChrisGuzak (approved):
I'm in favor of the macro cleanup too.

strings/base_macros.h
```diff
@@ -77,14 +77,53 @@ struct IUnknown;
...
+#if !defined(__cpp_lib_source_location)
+// cpp17 mode.  The source_location intrinsics are not available.
+#define WINRT_IMPL_BUILTIN_LINE 0
+#define WINRT_IMPL_BUILTIN_FILE nullptr
+#define WINRT_IMPL_BUILTIN_FUNCTION nullptr
+#elif defined(WINRT_NO_SOURCE_LOCATION)
+// The caller has disabled source_location support.  Ensure that there is no binary size overhead for line/file/function.
+#define WINRT_IMPL_BUILTIN_LINE 0
+#define WINRT_IMPL_BUILTIN_FILE nullptr
+#define WINRT_IMPL_BUILTIN_FUNCTION nullptr
```
CaseyCarter, line 115: Combining these identical cases is clearer, IMO:
```suggestion
#if !defined(__cpp_lib_source_location) || defined(WINRT_NO_SOURCE_LOCATION)
// Either the source_location intrinsics are not available, or the user doesn't want them.
// Ensure that there is no binary size overhead for line/file/function.
#define WINRT_IMPL_BUILTIN_LINE 0
#define WINRT_IMPL_BUILTIN_FILE nullptr
#define WINRT_IMPL_BUILTIN_FUNCTION nullptr
```

strings/base_macros.h
```diff
@@ -77,14 +77,53 @@ struct IUnknown;
...
+// Debug builds include FUNCTION information, which has a heavy binary size impact.
+#define WINRT_IMPL_BUILTIN_LINE __builtin_LINE()
+#define WINRT_IMPL_BUILTIN_FILE __builtin_FILE()
+#define WINRT_IMPL_BUILTIN_FUNCTION __builtin_FUNCTION()
+#else
+// Release builds in cpp20 mode get file and line information.
+#define WINRT_IMPL_BUILTIN_LINE __builtin_LINE()
+#define WINRT_IMPL_BUILTIN_FILE __builtin_FILE()
+#define WINRT_IMPL_BUILTIN_FUNCTION nullptr
+#endif
```
CaseyCarter, line 126: Similarly, I think extracting the identical parts makes this easier to understand:
```suggestion
#else
#define WINRT_IMPL_BUILTIN_LINE __builtin_LINE()
#define WINRT_IMPL_BUILTIN_FILE __builtin_FILE()
#if _DEBUG
// Debug builds include FUNCTION information, which has a heavy binary size impact.
#define WINRT_IMPL_BUILTIN_FUNCTION __builtin_FUNCTION()
#else
// Release builds in cpp20 mode get only file and line information.
#define WINRT_IMPL_BUILTIN_FUNCTION nullptr
#endif
#endif
```

strings/base_macros.h
CaseyCarter: `current` doesn't violate ODR. The default arguments are considered to be separate definitions from the function definition. The function definition itself is the same in all TUs (after the change, to be clear) regardless of support for `source_location` or the definition of `WINRT_NO_SOURCE_LOCATION`. The difference in the default argument definitions _does_ affect the definitions of other functions that call `slim_source_location::current` in ways that use those default arguments. The effect is as if the tokens in the definition of the default argument appeared in the definition of the calling function (https://eel.is/c++draft/basic.def.odr#15.11). If some of those callers are defined in multiple TUs that differ on support for `source_location` or the definition of `WINRT_NO_SOURCE_LOCATION`, then those callers are violating the ODR.

Outside of the guarantees the Standard provides, I'll say this seems like a fairly benign ODR violation. I don't think it's unreasonable to remove the `detect_mismatch`, but even if you're incredibly conservative I'd suggest providing an escape hatch macro for users to define to disable the mismatch detection.

strings/base_macros.h
dmachaj: I disagree on this one (although the code is small in both cases so the impact is limited either way).  In my opinion a base condition with fallthrough to other conditions is going to be harder to understand than independent blocks that are self-contained.  If you are looking at _DEBUG mode, for example, then you don't need to look anywhere else besides that one case.

strings/base_macros.h
dmachaj: I think your assessment is in alignment with what I'm seeing in a local test app.

There is an ODR violation but it is not a dangerous one.  For a given `winrt::consume_X` generated function that has confliction definitions one will include source, the other will not.  The linker then chooses which is the winner and the resulting binary will either have, or not have, source information in that function based on the choice.  Regardless of the choice the `slim_source_location` class is the same and the ABI is the same too so there is no corruption or anything really bad like that.

I think the matrix is roughly as follows:
1) No ODR, everything is consistent.  This is the happy option.
2) Mixing source location options.  For every function that is actually called in a given translation unit:
2a) Linker chooses the one with source information.  Final binary has source info for that call no matter which compiland is calling it.
2b) Linker chooses the one without source information.  Finaly binary does not have source info for that call, no matter which compiland is calling it.
3) Functions that are only called from one compiland or the other will have the source information for that compiland. In other words:
3a) Only cpp17 code calls a given function.  That function will not have source info.
3b) Only cpp20 code calls a given function.  That function will have source info.

Overall this feels fine to me.  We really shouldn't be mixing static libs across language versions so in most cases this doesn't matter.  When it does matter, the source info may be missing in some cases, but at no point will this introduce crashes or other major problems.

strings/base_macros.h
dmachaj: I never liked them either, even though I added them.  By popular request they are now gone :).

strings/base_macros.h
CaseyCarter: I think the `// Release builds in cpp20 mode get file and line information.` comment is what initially confused me. "But they both have file and line info - what is this comment trying to tell me?" I think clarifying that it's _only_ file and line info in release mode would avoid that confusion.

dmachaj: I can try to clarify.

dmachaj: Updated the comments; please let me know if they are clear enough to you.  Or if you think further clarification is needed.

CaseyCarter (approved):

DefaultRyan (approved):

dikan-i:
When new version with this fix will be released ?

PR #1454, opened 2024-11-18
GitHub Actions workflow cannot build arm32 after updating Windows SDK

### LINKED ISSUE #1452: GitHub Actions workflow cannot build arm32 after updating Windows SDK min version
### Version

_No response_

### Summary

Here's a CI build where it is failing: https://github.com/microsoft/cppwinrt/actions/runs/11865784595/job/33073424298

It seems that I broke it earlier today when I upgraded the SDK version to something newer: https://github.com/microsoft/cppwinrt/pull/1451.  I'm not sure why that built just fine.  It should have broken the same way.

```
D:\a\cppwinrt\cppwinrt>call msbuild /m /p:Configuration=Release,Platform=arm,CppWinRTBuildVersion=3.0.0.0 cppwinrt.sln /t:fast_fwd 
MSBuild version 17.12.6+db5f6012c for .NET Framework
Build started 11/16/2024 2:49:51 AM.

     1>Project "D:\a\cppwinrt\cppwinrt\cppwinrt.sln" on node 1 (fast_fwd target(s)).
     1>ValidateSolutionConfiguration:
         Building solution configuration "Release|arm".
     1>Project "D:\a\cppwinrt\cppwinrt\cppwinrt.sln" (1) is building "D:\a\cppwinrt\cppwinrt\fast_fwd\fast_fwd.vcxproj" (2) on node 1 (default targets).
     2>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.Cpp.WindowsSDK.targets(49,5): error MSB8037: The Windows SDK version 10.0.26100.0 for Desktop C++ ARM Apps was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".  [D:\a\cppwinrt\cppwinrt\fast_fwd\fast_fwd.vcxproj]
     2>Done Building Project "D:\a\cppwinrt\cppwinrt\fast_fwd\fast_fwd.vcxproj" (default targets) -- FAILED.
     1>Done Building Project "D:\a\cppwinrt\cppwinrt\cppwinrt.sln" (fast_fwd target(s)) -- FAILED.
```

The build then exits with a failure at the end because this file is not found:
```
D:\a\cppwinrt\cppwinrt>nuget pack nuget\Microsoft.Windows.CppWinRT.nuspec -Properties target_version=3.0.0.0;cppwinrt_exe=D:\a\cppwinrt\cppwinrt\_build\x86\Release\cppwinrt.exe;cppwinrt_fast_fwd_x86=D:\a\cppwinrt\cppwinrt\_build\x86\Release\cppwinrt_fast_forwarder.lib;cppwinrt_fast_fwd_x64=D:\a\cppwinrt\cppwinrt\_build\x64\Release\cppwinrt_fast_forwarder.lib;cppwinrt_fast_fwd_arm=D:\a\cppwinrt\cppwinrt\_build\arm\Release\cppwinrt_fast_forwarder.lib;cppwinrt_fast_fwd_arm64=D:\a\cppwinrt\cppwinrt\_build\arm64\Release\cppwinrt_fast_forwarder.lib  
Attempting to build package from 'Microsoft.Windows.CppWinRT.nuspec'.
Could not find a part of the path 'D:\a\cppwinrt\cppwinrt\_build\arm\Release'.
Error: Output nuget package not found!
Error: Process completed with exit code 1.
```

The first question is whether there is any reason to maintain a native arm32 build of cppwinrt.  Windows does not support arm32 at all on Win11 24H2.  Previous Win11 versions only supported arm32 as WOW (not as a native architecture).

If we are ok with removing arm32 support then that is the easiest and most direct fix.  Otherwise I either need to adjust the SDK version or I need to generate a zero-byte stub file just to make the .nuspec happy.  I would prefer simply removing native arm32 support.

### Reproducible example

```cpp

```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional comments

_No response_

dmachaj:
Closes #1452

arm32 is not supported anymore by Windows.  There has not been a native arm32 OS since before Win11.  The wow32 backcompat on arm64 devices to run arm32 programs was removed with Win11 24H2.  Most notably, the ability to build arm32 with newer SDKs and toolsets has seemingly been removed.  This is now breaking the CI build.

The easiest fix is to simply remove arm32 support.

See: https://www.microsoft.com/en-us/windows/windows-11-specifications?r=1#table3

See: https://learn.microsoft.com/en-us/windows/arm/arm32-to-arm64
"Windows devices running on an Arm processor (for example, Snapdragon processors from Qualcomm) will no longer support AArch32 (Arm32). This change impacts Universal Windows Platform apps that presently target AArch32 (Arm32). [Support for 32-bit Arm versions of applications will be removed in a future release of Windows 11.](https://www.microsoft.com/windows/windows-11-specifications#table3). System binaries for ARM32 support (present in the sysarm32 folder) will also be removed."

See: https://blogs.windows.com/windows-insider/2023/07/12/announcing-windows-11-insider-preview-build-25905/
(July 2023)
"Starting in this Insider Preview build in the Canary Channel, we have removed support for Arm32 UWP applications from Windows on Arm, as documented here: Windows 11 Specs and System Requirements. After the OS upgrade, any installed Arm32 applications will no longer launch on your device. Note that this does not affect Arm64 applications."


nuget/Microsoft.Windows.CppWinRT.nuspec
```diff
@@ -20,7 +20,6 @@
     <file src="$cppwinrt_exe$" target="bin"/>
     <file src="$cppwinrt_fast_fwd_x86$" target="build\native\lib\win32"/>
     <file src="$cppwinrt_fast_fwd_x64$" target="build\native\lib\x64"/>
-    <file src="$cppwinrt_fast_fwd_arm$" target="build\native\lib\arm"/>
```
jonwis, line 23: Are there other source files to delete here?

jonwis (dismissed):

nuget/Microsoft.Windows.CppWinRT.nuspec
dmachaj: Good question.  I was focused on the nuget and its build breaks but it does seem like there is room for a deeper clean.  Pending.

jonwis:
Might find and reference the right support KB for removal, just for posterity - https://support.microsoft.com/en-us/topic/windows-11-version-24h2-update-history-0929c747-1815-4543-8461-0160d16f15e5 is the latest, in case that helps.

dmachaj:
> Might find and reference the right support KB for removal, just for posterity - https://support.microsoft.com/en-us/topic/windows-11-version-24h2-update-history-0929c747-1815-4543-8461-0160d16f15e5 is the latest, in case that helps.

Updated the description of the PR with some links

nuget/Microsoft.Windows.CppWinRT.nuspec
dmachaj: Wow, there was a lot to update to fully remove it.  I just pushed a large iteration that prunes a lot of things (and converts many others from arm32-only to arm64-only).

jonwis (approved):

sylveon (approved):
The fact this is over 2000 lines of vcxproj changes makes me sad. Really wish the project format was better.

dmachaj:
> The fact this is over 2000 lines of vcxproj changes makes me sad. Really wish the project format was better.

Yeahhhhhhh.  I can comprehend wanting the vcxprojs to be more independent (vs. centralizing in some far-flung props/targets files) but the duplication is pretty extreme.  Even within the vcxproj's there is a ton of identical content for each supported build flavor.  That could be refactored to be 5-10x less lines of XML but that will have to wait for another day.

tigerw:
I was relying on ARM32 for WP builds...

sylveon:
cppwinrt should still work with older ARM32 SDKs.

tigerw:
Oh my goodness it's sylveon himself! <3

I managed to get my project compiling by manually patching the generated base.h file but not sure what the `thunks.asm` file is for, do you know if it's important to re-add?

sylveon:
Ah, the thinks file was used for fastcall ABI iirc. It looks like o was wrong and arm32 support was as fully stripped tho, in which case you're probably better off just keeping an old version prior to this change.

PR #1461, opened 2024-12-06
cppwinrt code base should have clang-format style formatting enforced automatically

### LINKED ISSUE #1459: cppwinrt code base should have clang-tidy formatting enforced automatically
### Version

_No response_

### Summary

There is nothing enforcing coding style consistency so there is variance that creeps in over time (e.g. some files have tabs, some have spaces).  It would be nice to use automation to enforce at least a bare minimum of consistency so that there is one less thing to think about.

### Reproducible example

```cpp

```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional comments

_No response_

dmachaj:
Closes #1459

Over time a lot of style inconsistencies have snuck into this code base (including things like tabs vs. spaces within a single file).  These aren't in and of themselves a big deal but it is nice to apply tooling to this type of problem so that it becomes automatic and we don't have to worry about it during code reviews.

This set of changes follows the pattern established by WIL in https://github.com/microsoft/wil/pull/416, with some modifications.  The .clang-format file in use here is a copy of the WIL file, with some light modifications to reduce the number of lines churned.  Settings like "don't indent the first level of namespace" touched a massive fraction of lines of code so those were relaxed to more closely match the exsting patterns in this repo.  I also decided to embed the "check formatting during CI" directly in the YML instead of having a separate script for it.  I think that is specialized enough that it belongs in the YML whereas having multiple formatting batch files can be confusing.

The `build_test_all.cmd` script was modified to automatically run clang-format before building the code.  I think most/all contributors rely on that script so that should make formatting automatic and therefore not something that needs to be worried about at all.

The PR build pipeline was also modified to run `clang-format` before building and will fail if any of the modified files has changes from formatting.  This will insta-fail any PR builds that don't match formatting requirements.  Therefore any PRs that build successfully meet the style and nobody has to review for style.

By far the biggest set of changes come from the formatting.  Just about 100% of files have some sort of change made to them.  Many files have tiny/trivial changes.  Some had a lot of formatting changes.  I tried to tweak the configuration a bit to reduce the churn as much as possible while still becoming consistent across the repo (see https://clang.llvm.org/docs/ClangFormatStyleOptions.html for the full list of options).

We can debate style choices as part of this PR to try and find the ideal balance between desired style and lines churned.  I'm happy to tweak the .clang-format file based on feedback.

A very small set of locations had formatting disabled because I wasn't sure if it was doing the right thing.  We can add more suppressions if necessary.

`build_test_all.cmd` locally.


.clang-format
```diff
@@ -0,0 +1,236 @@
...
+  ActivatableClass,
+  ActivatableClassWithFactory,
+  ActivatableClassWithFactoryEx,
+  ActivatableStaticOnlyFactory,
+  ActivatableStaticOnlyFactoryEx,
+  CoCreatableClass,
+  CoCreatableClassWithFactory,
+  CoCreatableClassWithFactoryEx,
+
+  # Additional definitions for the WIL repo
```
dunhor, line 176: Probably not relevant to keep this, but at least probably want to change the comment

.clang-format
```diff
@@ -0,0 +1,236 @@
...
+  SECTION,
+]
+
+TypenameMacros: [
+  IFACEMETHOD,
+  STDMETHOD,
+  STDAPI_,
+]
+
+# Additional definitions for the WIL repo
```
dunhor, line 191: Same here & the below

.github/workflows/ci.yml
```diff
@@ -7,6 +7,40 @@ on:
...
+      )
+      call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+
+      call find_clang_format.cmd
+      if %ERRORLEVEL% neq 0 exit /b %ERRORLEVEL%
+
+      git clang-format origin/master --binary "%CLANG_FORMAT%" --style file -- cppwinrt/*.h cppwinrt/*.cpp fast_fwd/*.h fast_fwd/*.cpp prebuild/*.h prebuild/*.cpp scratch/*.h scratch/*.cpp strings/*.h strings/*.cpp test/*.h test/*.cpp vsix/*.h vsix/*.cpp
+
+      if %ERRORLEVEL% neq 0 (
+        echo ##vso[task.logissue type=error]ERROR: This branch contains changes that have not been formatted with 'clang-format'
```
dunhor: `echo ##vso[task.logissue type=error]` probably not valid for GitHub actions(?)

.clang-format
dmachaj: Yep, good call

.github/workflows/ci.yml
dmachaj: Agreed.  I think I've got it converted over to GH syntax now.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,38 @@ on:
       - master
 
 jobs:
+  check-formatting:
+    name: 'Check Formatting of Changes'
+    runs-on: windows-latest
+    steps:
+      - name: Check Formatting
+        run: |
+          if NOT EXIST "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat" (
```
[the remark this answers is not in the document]
dmachaj: Unsure.  This is the approach used by WIL, which seems to have been stable for at least a year without any breakage that I've seen.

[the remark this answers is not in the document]
dunhor: Just to comment for WIL, AFAIK running on `windows-2022` ensures VS 2022. IIRC I actually copied that portion of the CI yaml from C++/WinRT long ago. But it looks like other parts of the pipeline have:
```
	          $VSDevCmd = & "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere" -latest -find Common7\tools\VSDevCmd.bat
          if (!$VSDevCmd) { exit 1 }	          if (!$VSDevCmd) { exit 1 }
          echo "Using VSDevCmd: ${VSDevCmd}"	          echo "Using VSDevCmd: ${VSDevCmd}"
          Add-Content $env:GITHUB_ENV "VSDevCmd=$VSDevCmd"	          Add-Content $env:GITHUB_ENV "VSDevCmd=$VSDevCmd"
```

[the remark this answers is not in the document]
DefaultRyan: WindowsAppSDK uses [vswhere ](https://github.com/microsoft/vswhere)to look for the environment variable `$VCToolsInstallDir` so that it can then run `VsDevCmd.bat` to obtain a developer command prompt.

.clang-format
```diff
@@ -0,0 +1,236 @@
...
+WhitespaceSensitiveMacros: [
+  __WI_CLANG_DISABLE_WARNING,
+]
+
+# TODO: Only supported by clang-format version 17+
+# Macros:
+# - WI_NOEXCEPT=noexcept
```
DefaultRyan, line 198: Leftover WIL stuff.

.clang-format
```diff
@@ -0,0 +1,236 @@
...
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLambdasOnASingleLine: Empty
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: MultiLine
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
```
DefaultRyan, line 24: There's a fair bit of config here that we could eliminate by basing the style on Microsoft, rather than llvm. See getMicrosoftStyle: https://github.com/hdoc/llvm-project/blob/a38b25fa77bdf1437c690494ae6d61179b3bb4f8/clang/lib/Format/Format.cpp#L1622

cppwinrt/cmd_reader.h
```diff
@@ -520,18 +493,21 @@ namespace cppwinrt
 
         auto files(std::string_view const& name) const
         {
-            return files(name, [](auto&&) {return true; });
```
DefaultRyan, line 523: Not sure I love this change. Consider removing the override of `AllowShortLambdasOnASingleLine`.

cppwinrt/code_writers.h
```diff
@@ -474,25 +468,15 @@ namespace cppwinrt
...
-                bind<write_guid_comment>(guid));
+            w.write(format, type, bind<write_guid_value>(guid), bind<write_guid_comment>(guid));
         }
         else
         {
             auto format = R"(    template <%> inline constexpr guid guid_v<%>{ pinterface_guid<%>::value };
     template <%> inline constexpr guid generic_guid_v<%>{ % }; // %
 )";
 
-            w.write(format,
```
DefaultRyan, line 488: I find these `.write(...)` calls to be harder to match with the format string when the args get packed, and easier when they are one-per-line. But why is this happening even though `BinPackArguments: false` is in the format file?

cppwinrt/cmd_reader.h
dmachaj: Trying out setting it to `Inline` (per https://clang.llvm.org/docs/ClangFormatStyleOptions.html#allowshortlambdasonasingleline)

cppwinrt/code_writers.h
dmachaj: The column length cap is relatively long so many of these fit onto a single line now.

.github/workflows/ci.yml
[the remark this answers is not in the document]
dmachaj: Updated to use vswhere.

.clang-format
dmachaj: I wasn't aware of a Microsoft base style.  I declared a base style on that and deleted all definitions that are the same across both.

cppwinrt/code_writers.h
dmachaj: I let the default style shorten the column cap back to 120 which fixed _most_ of these.  I'll also check and see if there's a way to special-case by method name.

DefaultRyan: To be fair, when it's a shorter list, like 2-3 args, the readability on one line isn't bad. It was just a bit rough when I was having to count more than 5-6 args and they were starting to cram onto separate lines.

Finally figured out that BinPackArguments = false still allows packing the args, as long as they fit on a single line. So that probably explains it.

cppwinrt/cmd_reader.h
```diff
@@ -153,13 +145,13 @@ namespace cppwinrt
...
 
         if (0 != RegOpenKeyExW(
-            HKEY_LOCAL_MACHINE,
-            L"SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots",
-            0,
-            // https://task.ms/29349404 - The SDK sometimes stores the 64 bit location into KitsRoot10 which is wrong,
-            // this breaks 64-bit cppwinrt.exe, so work around this by forcing to use the WoW64 hive.
-            KEY_READ | KEY_WOW64_32KEY, 
-            &key))
+                     HKEY_LOCAL_MACHINE,
```
DefaultRyan, line 148: Not against it. But I found this big indent curious. My only idea is that it's applying a 4-space indent from the beginning of `RegOpenKeyExW`, instead of `0`.

strings/base_windows.h
```diff
@@ -302,7 +291,7 @@ WINRT_EXPORT namespace winrt
     }
 
     template <typename T, std::enable_if_t<!std::is_base_of_v<Windows::Foundation::IUnknown, T>, int> = 0>
-    auto put_abi(T& object) noexcept
+    auto put_abi(T & object) noexcept
```
DefaultRyan, line 294: What's happening here? Earlier in this same file, it left these reference parameters left-aligned. But here it suddenly switches to middle alignment?

dmachaj: Not sure about this one.  I tried setting `ReferenceAlignment: Left` and it still adds the space.

dmachaj: I think the templated T is what confuses it.  I see _many_ examples of changes in this PR where `var & i` is correctly adjusted to be `var& i`.  So this is not universally wrong.  There are just a small number of false positive type matches.

dmachaj:
I'm going to let this one expire.  I have moved on to other work and don't have the capacity to push on it further, and there was not a good consensus on the cost/benefit.

If people feel like this is desired after all then we can always dig this PR back up and try again later.

PR #1462, opened 2024-12-12
Make consume methods more amenable to optimization

DefaultRyan:
The temporary std::pair introduced with `try_as_with_reason` sometimes causes spurious temporary dtor calls, depending on the optimization flags used and/or the version of the compiler. This didn't result in an extra refcount bump/release, but it still resulted in generating spurious destructor calls to `winrt::Foundation::~IUnknown()` as a result of that temporary.

The temporary had already been moved-from into the local variables, so was always zero, but this code bloat, extra function call and branch, when multiplied over thousands of projected methods can still add up.

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -1137,7 +1137,8 @@ namespace cppwinrt
     {%
         if constexpr (!std::is_same_v<D, %>)
         {
-            auto const [_winrt_casted_result, _winrt_cast_result_code] = impl::try_as_with_reason<%, D const*>(static_cast<D const*>(this));
+            winrt::hresult _winrt_cast_result_code;
+            auto const _winrt_casted_result = impl::try_as_with_reason<%, D const*>(static_cast<D const*>(this), _winrt_cast_result_code);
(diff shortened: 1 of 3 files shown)


cppwinrt/code_writers.h
```diff
@@ -1137,7 +1137,8 @@ namespace cppwinrt
     {%
         if constexpr (!std::is_same_v<D, %>)
         {
-            auto const [_winrt_casted_result, _winrt_cast_result_code] = impl::try_as_with_reason<%, D const*>(static_cast<D const*>(this));
+            winrt::hresult _winrt_cast_result_code;
+            auto const _winrt_casted_result = impl::try_as_with_reason<%, D const*>(static_cast<D const*>(this), _winrt_cast_result_code);
```
dmachaj, line 1141: I'm not sure if it is still the recommended coding standard, but for a long time non-const references were disallowed and passing a pointer was preferred.  That way the call site is clearly mutating the variable because the & is visible.  In this case it is generated code so I guess it is [shrug].

dmachaj (approved):

PR #1463, opened 2024-12-12
Fix warnings in natvis project

DefaultRyan:
There are a slew of compilation warnings in the natvis project. This addresses them.

Most are coming from a the `<vsdebugeng.h>` header in the `Microsoft.VSSDK.Debugger.VSDebugEng.16.0.2012201-preview` package.

But there were a few where a global variable named `db` was getting hidden by a local `db` in a different scope. Let's just give the global a more unique name for now.

### DIFF
--- a/natvis/pch.h
+++ b/natvis/pch.h
@@ -4,7 +4,10 @@
 #define NOMINMAX
 
 #include <windows.h>
+#pragma warning(push)
+#pragma warning(disable : 4471)
 #include <vsdebugeng.h>
+#pragma warning(pop)
 #include <vsdebugeng.templates.h>
 #include <Dia2.h>
 #include "base_includes.h"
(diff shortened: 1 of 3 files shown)


natvis/pch.h
```diff
@@ -4,7 +4,10 @@
 #define NOMINMAX
 
 #include <windows.h>
+#pragma warning(push)
+#pragma warning(disable : 4471)
 #include <vsdebugeng.h>
```
dmachaj, line 9: Is there a way to pull a newer version of this nuget package?  I tried specifying the latest but couldn't `nuget pull` after that.

dmachaj (approved):

natvis/pch.h
DefaultRyan: I've tried before, and there's been some weirdness. But I've been making some visualizer improvements and hope to do some updating as part of that work! For now, though, I'm a big fan of keeping PRs tightly scoped when possible. :)

PR #1467, opened 2025-01-08
Add get_iids_tearoff() hook.

dlech:
Add a `get_iids_tearoff()` hook to complement the `query_interface_tearoff()` hook.

`query_interface_tearoff()` allows dynamically extending types with additional interfaces that can be accessed with the `QueryInterface()` method. However, these interfaces were not discoverable by the `GetIids()` method (and therefore not by the `winrt::get_interfaces()` function either).

The `get_iids_tearoff()` hook allows adding the GUIDs of interfaces to the array returned by `GetIids()`.

This also fixes a bug in the implementation of the `root_implements_type::is_composing` branch of `NonDelegatingGetIids()` where `*array` was updated causing the local iids to be unreachable by the caller and risking the caller reading past the end of the array.

### DIFF
--- a/strings/base_implements.h
+++ b/strings/base_implements.h
@@ -912,6 +912,11 @@ namespace winrt::impl
             return error_no_interface;
         }
 
+        virtual std::vector<guid> get_iids_tearoff() const noexcept
+        {
+            return {};
+        }
+
         root_implements() noexcept
         {
         }
@@ -1021,7 +1026,8 @@ namespace winrt::impl
         int32_t __stdcall NonDelegatingGetIids(uint32_t* count, guid** array) noexcept
         {
             auto const& local_iids = static_cast<D*>(this)->get_local_iids();
-            uint32_t const& local_count = local_iids.first;
+            auto tearoff_iids = get_iids_tearoff();
+            auto local_count = local_iids.first + tearoff_iids.size();
             if constexpr (root_implements_type::is_composing)
             {
                 if (local_count > 0)
@@ -1033,8 +1039,10 @@ namespace winrt::impl
                     {
                         return error_bad_alloc;
                     }
-                    *array = std::copy(local_iids.second, local_iids.second + local_count, *array);
-                    std::copy(inner_iids.cbegin(), inner_iids.cend(), *array);
+                    auto _array = *array;
+                    _array = std::copy(local_iids.second, local_iids.second + local_iids.first, _array);
+                    _array = std::copy(tearoff_iids.cbegin(), tearoff_iids.cend(), _array);
+                    std::copy(inner_iids.cbegin(), inner_iids.cend(), _array);
                 }
                 else
                 {
@@ -1051,7 +1059,9 @@ namespace winrt::impl
                     {
                         return error_bad_alloc;
                     }
-                    std::copy(local_iids.second, local_iids.second + local_count, *array);
+                    auto _array = *array;
+                    _array = std::copy(local_iids.second, local_iids.second + local_iids.first, _array);
+                    std::copy(tearoff_iids.cbegin(), tearoff_iids.cend(), _array);
                 }
                 else
                 {
--- a/test/old_tests/UnitTests/Composable.cpp
+++ b/test/old_tests/UnitTests/Composable.cpp
@@ -1,12 +1,14 @@
 #include "pch.h"
 #include "catch.hpp"
 #include "winrt/Composable.h"
+#include <windows.foundation.h>
 
 using namespace winrt;
 using namespace Windows::Foundation;
 using namespace Composable;
 
 using namespace std::string_view_literals;
+using hstring = ::winrt::hstring;
 
 namespace
 {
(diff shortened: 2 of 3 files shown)


strings/base_implements.h
```diff
@@ -1021,7 +1026,8 @@ namespace winrt::impl
         int32_t __stdcall NonDelegatingGetIids(uint32_t* count, guid** array) noexcept
         {
             auto const& local_iids = static_cast<D*>(this)->get_local_iids();
-            uint32_t const& local_count = local_iids.first;
+            auto tearoff_iids = get_iids_tearoff();
+            uint32_t const& local_count = local_iids.first + tearoff_iids.size();
```
dmachaj, line 1030: I could be mistaken but it seems like this shouldn't be a reference anymore.

strings/base_implements.h
```diff
@@ -1033,8 +1039,10 @@ namespace winrt::impl
                     {
                         return error_bad_alloc;
                     }
-                    *array = std::copy(local_iids.second, local_iids.second + local_count, *array);
-                    std::copy(inner_iids.cbegin(), inner_iids.cend(), *array);
+                    auto _array = *array;
+                    _array = std::copy(local_iids.second, local_iids.second + local_iids.first, _array);
+                    _array = std::copy(tearoff_iids.cbegin(), tearoff_iids.cend(), _array);
```
dmachaj, line 1044: Do these copies go to the unoccupied slots in the array or do they keep overwriting the first elements?  As best as I can tell `_array` is a pointer, not an iterator, so copy wouldn't be advancing it.

test/old_tests/UnitTests/Composable.cpp
```diff
@@ -167,4 +169,120 @@ TEST_CASE("Composable conversions")
...
+                return make_stringable(*this, L"ToString", result);
+            }
+
+            *result = nullptr;
+            return E_NOINTERFACE;
+        }
+
+        winrt::array_view<winrt::guid> get_iids_tearoff() const noexcept final
+        {
+            return winrt::array_view(tearoff_iids);
```
dmachaj, line 271: What is the behavior when the author makes a mistake an `get_iids_tearoff` duplicates entries that are already on the list?

test/old_tests/UnitTests/Composable.cpp
```diff
@@ -167,4 +169,120 @@ TEST_CASE("Composable conversions")
...
+        }
+    };
+
+	auto object = make<Tearoff>();
```
dmachaj, line 275: [nit] Mixing tabs and spaces on these lines

test/old_tests/UnitTests/Composable.cpp
```diff
@@ -167,4 +169,120 @@ TEST_CASE("Composable conversions")
...
+    auto ifaces = get_interfaces(object);
+
+    REQUIRE(object.as<IClosable>());
+    REQUIRE(object.as<IStringable>());
+	// IBaseOverrides is repeated twice for some reason, so actual size is 7 but there are only 6 unique interfaces
```
dmachaj, line 280: This seems odd and possibly worth understanding before proceeding.  Would you be willing to check the behavior of this test case with your `get_iid_tearoff` logic commented out?

test/test/tearoff.cpp
```diff
@@ -164,6 +164,13 @@ namespace
             *result = nullptr;
             return E_NOINTERFACE;
         }
+
+		inline static std::array<winrt::guid, 1> tearoff_iids = {winrt::guid_of<IPersist>()};
```
dmachaj, line 168: [nit] tabs vs spaces mixing in this file too

test/test/tearoff.cpp
```diff
@@ -195,6 +202,13 @@ namespace
 
             *result = nullptr;
             return E_NOINTERFACE;
+        }
+
+		inline static std::array<winrt::guid, 1> tearoff_iids{winrt::guid_of<winrt::IStringable>()};
+
+        winrt::array_view<winrt::guid> get_iids_tearoff() const noexcept final
+        {
+            return winrt::array_view(tearoff_iids);
```
dmachaj, line 211: It might be nice to have one of the test cases use an inline definition instead of deferring to a class field.  Something like `return winrt::array_view({winrt::guid_of<winrt::IStringable>()});` just to prove that it is possible.

test/test/tearoff.cpp
```diff
@@ -215,6 +229,11 @@ TEST_CASE("tearoff")
         REQUIRE(S_OK == persist->GetClassID(&result));
         REQUIRE(winrt::is_guid_of<IPersist>(result));
 
+        winrt::com_array<winrt::guid> iids = winrt::get_interfaces(closable);
```
dmachaj, line 232: This test file adds tearoff IIDs to both `RuntimeType` and `InteropType` but only has test cases for `InteropType`.  It would be nice to also check the IIDs of `RuntimeType`.

strings/base_implements.h
dlech: `std::copy` returns a pointer to the next empty element. So `_array` ensures we don't write over the beginning of `*array` with each successive copy.

test/old_tests/UnitTests/Composable.cpp
dlech: The existing code already can create duplicate entries in the final array that is returned by `GetIids()`. So I'm guessing it isn't a big deal. [The docs](https://learn.microsoft.com/en-us/windows/win32/api/inspectable/nf-inspectable-iinspectable-getiids) say that it returns:

> an array that contains an IID for each interface implemented by the current Windows Runtime object. The [IUnknown](https://learn.microsoft.com/en-us/windows/desktop/api/unknwn/nn-unknwn-iunknown) and [IInspectable](https://learn.microsoft.com/en-us/windows/desktop/api/inspectable/nn-inspectable-iinspectable) interfaces are excluded.

So other than some minor inefficiencies of copying more than we need and having to check a few more items in the array, it seems pretty harmless.

test/old_tests/UnitTests/Composable.cpp
dlech: It doesn't come from `get_iid_tearoff `. It is likely due to this being a composable type and the `get_local_iids()` of the outer class also picks up the overrideable interfaces from the inner class since overrideable interfaces get special handling.

test/test/tearoff.cpp
dlech: Wouldn't that result in a dangling pointer? `winrt::array_view` doesn't make a copy of the underlying array AFAIKT, so the returned array view would be pointing to stack allocated memory that is no longer valid from the function that just returned.

strings/base_implements.h
dmachaj: Ah, I missed where the _array pointer is updated to be the output of `std::copy`.

test/old_tests/UnitTests/Composable.cpp
dmachaj: Seems reasonable to me.

test/test/tearoff.cpp
dmachaj: I guess that would depend on whether it is stack allocated or compiletime.  I would expect an embedded array to be compiletime and therefore non-dangling because it is part of the DLL itself.

dlech: The [best info I could find](https://en.cppreference.com/w/cpp/language/list_initialization#List-initializing_std::initializer_list) says:

 > The backing array has the same lifetime as any other [temporary object](https://en.cppreference.com/w/cpp/language/lifetime#Temporary_object_lifetime), except that initializing an [std::initializer_list](https://en.cppreference.com/w/cpp/utility/initializer_list) object from the backing array extends the lifetime of the array exactly like [binding a reference to a temporary](https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary).

[And](https://en.cppreference.com/w/cpp/utility/initializer_list)

> An object of type std::initializer_list<T> is a lightweight proxy object that provides access to an array of objects of type const T (that may be allocated in read-only memory). 

So it sounds like it can't be relied upon.

Probably using `winrt::array_view` for a return value was a bad idea since it could lead to dangerous usage and I should pick something else.

dlech:
Updated:
* Use `std::vector` instead of `winrt::array_view` as return value to avoid memory safety pitfalls.
* Drop `const&` on `local_count`.
* Fix mixed tabs and spaces.
* Add assertions for `RuntimeType` test.

kennykerr:
I would caution against doing this. `GetIids` in general can and does cause code bloat due to all the static data tables it produces. @DefaultRyan did a bunch of work to mitigate the cost without turning it off entirely. 

I have never seen a practical use for `GetIids` beyond a very obscure fallback if-all-else-fails scenario in CLR debugging that is never actually used in practice. I always noop the implementation as a result. Far better to rely on metadata for "reflection" rather than run time reflection via `GetIids`. 😊

dlech:
> run time reflection

The use case for this is for the PyWinRT projection, hence the need for runtime reflection. Since Python is dynamic, we need a way to add interfaces at runtime, so can't rely on static data.  I can probably make something work without this if it isn't really useful for anyone else, but hooking into the existing code just makes that easier.

PR #1468, opened 2025-01-12
Fix reading past end of array.

dlech:
Fix a bug in the `NonDelegatingGetIids()` implementation that could cause reading past the end of the returned array.

`std::copy()` returns a pointer to the next element in the array after the last element copied. The output argument `*array` was being assinged to this pointer after the first copy, causing it to no longer point to the beginning of the array. If the caller tries to access the full array after this, it will read past the end of the array and will miss the first elements of the array.

To fix, introduce a new temporary `_array` variable to pass the result of the first copy as the starting point of the second copy. Also add a test that failed before the fix and passes after the fix.


dlech:
If we aren't going to merge #1467, this fixes a bug that was noted there.

strings/base_implements.h
```diff
@@ -1033,8 +1033,8 @@ namespace winrt::impl
                     {
                         return error_bad_alloc;
                     }
-                    *array = std::copy(local_iids.second, local_iids.second + local_count, *array);
-                    std::copy(inner_iids.cbegin(), inner_iids.cend(), *array);
+                    auto _array = std::copy(local_iids.second, local_iids.second + local_count, *array);
```
dmachaj, line 1036: [nit] I think the name `_array` with the leading underscore is misaligned with the naming pattern in this code base.  Something like `array_copy` or `local_array` would probably fit better.

test/old_tests/UnitTests/Composable.cpp
```diff
@@ -167,4 +167,20 @@ TEST_CASE("Composable conversions")
...
+TEST_CASE("Composable get_interfaces")
+{
+    struct Foo : Composable::BaseT<Foo, IStringable> {
+        hstring ToString() const { return L"Foo"; }
+    };
+
+    auto obj = make<Foo>();
+    auto iids = winrt::get_interfaces(obj);
+    // BaseOverrides IID gets repeated twice. There are only 4 unique interfaces.
+    REQUIRE(iids.size() >= 4);
```
dmachaj, line 181: Maybe its just me, but `>=4` seems overly conservative.  If there is a well-known repeated IID then can this assert the size is 5 instead?  Perhaps that's not ideal but it is at least testing the current behavior so that if it changes later the test will identify the behavior change.

dlech:
Updated:
* Renamed `_array` to `next`.
* Changed test `== 5`.

dmachaj (approved):

DefaultRyan:
Looks good to me.

PR #1472, opened 2025-01-31
Allow visualization of properties on generic types

### LINKED ISSUE #784: Watch does not work for IReference<TimeSpan>
If you are trying to view a Windows::Foundation::IReference<Windows::Foundation::TimeSpan> type in watch then the value is not displayed and debug window shows:

Natvis C++/WinRT: Could not find metadata for Windows.Foundation.IReference`1<Windows.Foundation.TimeSpan>

DefaultRyan:
We can now visualize properties on generic types. For example:
- `IReference<T>.Value`
- `IVector<T>.Size`
- `IKeyValuePair<T>.Key` and `IKeyValuePair<T>.Value`

This does the heavy lifting of parsing the generic type strings received from the debugger, synthesizing type signatures for generics, calculating the pinterface UUID, and retrieving the property.

Note that this change only attacks true properties. For collections, (e.g. Vector, Map), only `Size` is a property. Obtaining the contents requires non-property method calls. Fortunately, I have an idea for that which I'll add in a separate PR.

Fixes: #784

image
image
image


natvis/cppwinrt_visualizer.cpp
```diff
@@ -135,19 +139,98 @@ TypeDef FindType(DkmProcess* process, std::string_view const& typeName)
...
+    std::string_view partName = *it;
+    auto basic_type_pos = std::find_if(std::begin(elementNames), std::end(elementNames), [&partName](auto&& elem) { return elem.first == partName; });
+    if (basic_type_pos != std::end(elementNames))
+    {
+        return TypeSig{ basic_type_pos->second };
+    }
+    
+    TypeDef type = FindSimpleType(process, partName);
+    auto tickPos = partName.rfind('`');
+    if (tickPos == std::string_view::basic_string_view::npos)
```
Scottj1s, line 199: can you just say partName.npos (since it's a static member)?

natvis/cppwinrt_visualizer.cpp
```diff
@@ -135,19 +139,98 @@ TypeDef FindType(DkmProcess* process, std::string_view const& typeName)
...
+        {"UInt16", ElementType::U2},
+        {"UInt32", ElementType::U4},
+        {"UInt64", ElementType::U8},
+        {"Single", ElementType::R4},
+        {"Double", ElementType::R8},
+        {"String", ElementType::String},
+        {"Object", ElementType::Object}
+    };
+    std::string_view partName = *it;
+    auto basic_type_pos = std::find_if(std::begin(elementNames), std::end(elementNames), [&partName](auto&& elem) { return elem.first == partName; });
```
dmachaj, line 191: Does this need to be case insensitive?

Scottj1s (dismissed):

natvis/cppwinrt_visualizer.cpp
DefaultRyan: That's a good question, and I had to really think about this answer. From everything I've learned about WinRT over the years, the exact casing of the string returned by `GetRuntimeClassName` is not prescribed by the type system, just that _"In particular, IInspectable.GetRuntimeClassName enables an object's client to retrieve a WinRT typename that can be resolved in metadata to enable language projection."_ (source https://learn.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system)

That same WinRT type system doc linked above also calls out the names of these fundamental types, so these exact strings are standard, with the exception of `Object`. What isn't strictly specified in the standard is that an implementation's internal type for a generic must return a particular string from `GetRuntimeClassName`. 

With a concrete type that implements generic interfaces, like `Windows.Foundation.Collections.ValueSet`, it only needs to return that string. After looking up that string in the winmd, the relationship of the runtime class to the various generic interfaces it implements is all captured in the ECMA metadata.

But with these implementation-supplied objects, the class name is more of a convention, rather than a dictate of the type system, as far as I can tell. But what I do know is that I checked 3 of the big implementations that actually generate matching pinterface GUIDs and runtime class names, and they all use these same names. So, I think we're good here.

DefaultRyan: I did just notice that I missed `Char16` and `Guid`. I'm also going to ping the CsWinRT owners to verify they also return consistent strings for these cases.

Scottj1s (approved):

YexuanXiao:
Currently, the C++/WinRT extension provided by Visual Studio is still stuck at 2.0.22531.1. This feature has never been shipped to users. @DefaultRyan Can you update it?

PR #1473, opened 2025-02-05
Fix various build warnings

DefaultRyan:
Needed to fix a BinSkim issue flagging CFG on prebuild.exe. While I was at it, I fixed a few more warnings. A few were genuine code tweaks, a few were warning suppressions for stuff that's either by design or out of our control.

At this point, the official build pipelines are looking almost 100% clean. Might even be able to enable /WX and cause the build to break on new warnings.

### DIFF
--- a/cppwinrt/component_writers.h
+++ b/cppwinrt/component_writers.h
@@ -176,7 +176,7 @@ int32_t __stdcall WINRT_CanUnloadNow() noexcept
 {
 #ifdef _WRL_MODULE_H_
 #ifdef _MSC_VER
-#pragma warning(disable: 4324) // structure was padded due to alignment specifier
+#pragma warning(suppress: 4324) // structure was padded due to alignment specifier
 #endif
     if (!::Microsoft::WRL::Module<::Microsoft::WRL::InProc>::GetModule().Terminate())
     {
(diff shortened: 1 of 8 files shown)


prebuild/prebuild.vcxproj
```diff
@@ -88,6 +88,7 @@
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>..\cppwinrt</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+      <ControlFlowGuard>Guard</ControlFlowGuard>
```
sylveon, line 91: Build is failing on debug with this, conflicts with `/ZI`

cppwinrt/component_writers.h
```diff
@@ -175,6 +175,9 @@ void* __stdcall %_get_activation_factory([[maybe_unused]] std::wstring_view cons
 int32_t __stdcall WINRT_CanUnloadNow() noexcept
 {
 #ifdef _WRL_MODULE_H_
+#ifdef _MSC_VER
+#pragma warning(disable: 4324) // structure was padded due to alignment specifier
```
dmachaj, line 179: If possible it is preferrable to pragma warning push/pop so it is only disabled for the small bit of code.  If there is one line of warning then `pragma warning (suppress: 4324)` disables it for only the next line.

DefaultRyan: Good catch. This one was an oversight on my part. (Had to do a couple other push/disable/pop elsewhere because, compiler yuckiness, and got into a pattern, haha)

prebuild/prebuild.vcxproj
DefaultRyan: Thanks for catching. Fixed this and disabled CFG in the other Debug flavors.

dmachaj (approved):

sylveon (approved):

Issue #1481, opened 2025-03-24
Bug: COM ABI types are not declared with public LTO visibility


ZeroMemes:
### Version

N/A

### Summary

From the Clang docs on [LTO Visibility](https://clang.llvm.org/docs/LTOVisibility.html):

> Some ABIs provide the ability to define an abstract base class without visibility attributes in multiple linkage units and have virtual calls to derived classes in other linkage units work correctly. One example of this is COM on Windows platforms. If the ABI allows this, any base class used in this way must be defined with public LTO visibility. [...] *To specifically handle the COM case, classes with the `__declspec(uuid())` attribute receive public LTO visibility.*

Since the `__declspec(uuid())` attribute is not used in WinRT (in favor of its internal `winrt::impl::guid_v`), public LTO visibility is not assumed by Clang. Incorrect labeling of visibility results in misguided optimizations that can have unpredictable results on the program's runtime.

The example I've provided below was how I initially discovered this issue. Unfortunately, I have not been successful in creating a minimally reproducible example out of it (only going as far as to paste the entire function into a standalone app out of the much larger original source codebase). It was compiled with `clang-cl` 19.1.0 using `-O3 -flto -fwhole-program-vtables`. Removing the `fwhole-program-vtables` compiler option "resolved" the issue.

### Reproducible example

```cpp
winrt::fire_and_forget pickImage() {
    using namespace winrt;
    using namespace Windows::ApplicationModel::Core;
    using namespace Windows::UI::Core;
    using namespace Windows::Storage;
    using namespace Windows::Storage::Streams;
    using namespace Windows::Storage::Pickers;
    using namespace Windows::Graphics::Imaging;

    FileOpenPicker picker{};
    picker.ViewMode(PickerViewMode::Thumbnail);
    picker.FileTypeFilter().Append(L".png"); // <--- crash occurred here

    // Removing the code below this comment had no effect on the program crash
    if (auto file = co_await picker.PickSingleFileAsync()) {
        auto stream = co_await file.OpenAsync(Windows::Storage::FileAccessMode::Read);
        auto decoder = co_await BitmapDecoder::CreateAsync(stream);

        auto pixelData = co_await decoder.GetPixelDataAsync(
            BitmapPixelFormat::Rgba8,
            BitmapAlphaMode::Straight,
            BitmapTransform{},
            ExifOrientationMode::IgnoreExifOrientation,
            ColorManagementMode::DoNotColorManage
        );
[...]

PR #1482, opened 2025-03-24
Declare public LTO visibility for ABI type declarations

ZeroMemes:
It may also be appropriate to use `__attribute__ ((visibility ("default")))` on GCC, though I haven't looked into it.

Fixes: #1481


sylveon:
This PR seems fairly simple. Can we get someone to review it?

dmachaj (approved):

Issue #1486, opened 2025-04-24
MSVC warns about use of __has_attribute


dunhor:
### Summary

This was brought up on an internal channel. It seems as though #1482 has added use of `__has_attribute`, but doesn't properly guard its usage. The usage should probably be something more like:

```c++
#if defined(__clang__) && defined(__has_attribute)
#if __has_attribute(__lto_visibility_public__)
#define WINRT_IMPL_PUBLIC __attribute__((lto_visibility_public))
#endif
#endif

#ifndef WINRT_IMPL_PUBLIC
#define WINRT_IMPL_PUBLIC
#endif
```

It might also be worth building tests with `/WX` to catch such warnings earlier

### Reproducible example

```cpp

```

PR #1487, opened 2025-04-24
All cppwinrt vcxproj should build with /W4 and /WX (warnings turned up, treated as error).  Fix existing violations.

dmachaj:
Fixes: #1486

The most recent PR introduced a warning for non-Clang consumers.  Many projects build with warnings-as-errors so this is a build break for those projects.  Because cppwinrt is widely used by many such projects it should itself build with warnings as errors enabled so that this doesn't happen again in the future.  Plus it is general goodness to build with warnings cranked up and blocking.

I added /WX to every single vcxproj in this repo (which required a whole lot of redundant edits).  Many of them didn't have /W4 so I added it wherever it was missing.  The project files definitely need some refactoring to reduce duplication.  But for this change I decided to modify them as-is without also including a refactoring.

I also fixed the newly-blocking warnings.  Fortunately the project was basically clean and only the recent issue needed fixing.

I also removed a typo'd <TreatWarningsAsError> (warning is singular, not plural) that was trying to disable warnings in one specific project.  It has no warnings so we might as well baseline them as blocking.

It builds locally (x64 Debug+Release, only tested MSVC).


dmachaj:
`3>Class.cpp(522,10): error : unused function 'ValidateStaticEventAutoRevoke' [-Werror,-Wunused-function] [D:\a\cppwinrt\cppwinrt\test\test_component\test_component.vcxproj]`

I guess I need to figure out how to build with clang-cl.exe.  Last time I tried I wasn't able to get it working locally.

dunhor (approved):

dunhor:
> `3>Class.cpp(522,10): error : unused function 'ValidateStaticEventAutoRevoke' [-Werror,-Wunused-function] [D:\a\cppwinrt\cppwinrt\test\test_component\test_component.vcxproj]`
> 
> I guess I need to figure out how to build with clang-cl.exe. Last time I tried I wasn't able to get it working locally.

Probably because the function is defined in an anonymous namespace. Could just try moving it to the global namespace

dmachaj:
I was able to figure out how to get enough of clang-cl to build to reproduce the break:

```
msbuild /m /clp:ForceConsoleColor /p:Configuration=Debug,Platform=x64,CppWinRTBuildVersion=1.2.3.4,Clang=1,PlatformToolset=ClangCl cppwinrt.sln /t:fast_fwd
msbuild /m /clp:ForceConsoleColor /p:Configuration=Debug,Platform=x64,CppWinRTBuildVersion=1.2.3.4,Clang=1,PlatformToolset=ClangCl cppwinrt.sln /t:cppwinrt
msbuild /m /clp:ForceConsoleColor /p:Configuration=Debug,Platform=x64,CppWinRTBuildVersion=1.2.3.4,Clang=1,PlatformToolset=ClangCl cppwinrt.sln /t:test\test
```

I'm not sure if all 3 builds are necessary but that was enough to get the test\test directory to produce the same break as the CI build.

dmachaj:
For some reason the Clang warnings wouldn't all print out at once.  I had to keep fixing warnings and rebuilding until it was green.  There were a bunch of clang-only warnings that were previously unknown (mostly minor, all in test code).

There was also an x86-only signed/unsigned mismatch in the `format` code that I have attempted to fix.

test/test/return_params_abi.cpp
```diff
@@ -10,16 +10,6 @@ namespace winrt
 
 using namespace winrt;
 
-namespace
-{
-    hstring to_hstring(::IInspectable* raw)
```
dunhor, line 15: I assume the intent of this was "does it compile?" - so probably want to keep, but remove from anonymous namespace

Directory.Build.Props
```diff
@@ -49,6 +49,7 @@
   <ItemDefinitionGroup>
     <ClCompile>
       <WarningLevel>Level4</WarningLevel>
+      <TreatWarningAsError>true</TreatWarningAsError>
```
ChrisGuzak, line 52: I see there is a `Directory.Build.Props` at the root of the repo that has a `ClCompile` element. it this can go there and then these files can remove the `WarningLevle` and `TreadWarningAsErrors`

```xml
    <ClCompile>
      <WarningLevel>Level4</WarningLevel>
```

test/test/return_params_abi.cpp
dmachaj: That might be the case.  There was another like that which I preserved.  I misread this one as an unused helper instead of a "does it compile" test case.  I'll restore it with a suppression.

strings/base_activation.h
```diff
@@ -257,8 +257,14 @@ namespace winrt::impl
...
-
-        size_t& m_count;
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-private-field"
+#endif
+        size_t& m_count; // Field is unused when WINRT_NO_MODULE_LOCK is defined.
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
```
dunhor, line 267: So apparently you can use `[[maybe_unused]]` on a private data member: https://godbolt.org/z/Gq5qbKMKz. Looks a bit cleaner

strings/base_fast_forward.h
```diff
@@ -112,6 +116,11 @@ namespace winrt::impl
             return self->m_owner->GetTrustLevel(level);
         }
 
+
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wmicrosoft-cast"
```
dunhor, line 122: Is this just because it's converting a function pointer to `void*`? I'm curious if you can convert these to `void(*)()` - the function pointer equivalent of `void*` - and make things work equivalently. GitHub makes it a little tricky to see, but I'm _hoping_ these are in the `impl` namespace and therefore such a change won't be breaking.

Idk how much such a change might break internally, so it might not be desirable to fully explore right now, so perhaps as a follow-up.

strings/base_activation.h
dmachaj: Oh yeah, that's definitely better.  I tried "using" it in the `else` block above but the compiler was smart enough to optimize the NOOP usage out and still warn that it was unused.

strings/base_fast_forward.h
dmachaj: The warning was something along the lines of these casts only working on Microsoft toolchains.  cppwinrt can be used with different compilers, but as far as I know the outputs only ever run on Windows, so that is likely fine.  I would definitely prefer to leave this for a future checkin because I've already spent 10x longer on this PR than I expected going into it.

dunhor: > The warning was something along the lines of these casts only working on Microsoft toolchains

I'm _guessing_ that warning just means that MSVC allows the cast, not that it'll only work on Windows. But anyway, I agree that it's not necessary to address it right now.

test/test_cpp20/clang_only.cpp
```diff
@@ -0,0 +1,28 @@
...
+TEST_CASE("clang_library")
+{
+    // PickImage();
+
+    FileOpenPicker picker{};
+    picker.ViewMode(PickerViewMode::Thumbnail);
+    picker.FileTypeFilter().Append(L".png"); // <--- crash occurred here
+
+    REQUIRE(true);
+}
```
dunhor, line 26: Maybe add a comment what this is doing/testing & why there is code commented out

dunhor (dismissed):

test/test_cpp20/clang_only.cpp
dmachaj: [sigh] I totally did last night but my edits seem to have been lost.  Re-adding them.

dunhor (approved):

Issue #1499, opened 2025-08-05
Bug: `-Wnon-virtual-dtor` warn


qwertycxz:
### Version

v2.0.250303.1

### Summary

A warning could be raised from MinGW.

### Reproducible example

```cpp
#include <winrt/base.h>
```

### Expected behavior

No warnings.

### Actual behavior

warning: `struct winrt::impl::abi<winrt::Windows::Foundation::IUnknown>::type` has virtual functions and accessible non-virtual destructor [-Wnon-virtual-dtor]`

### Additional comments

A `virtual ~type() noexcept = default;` would fix that.

PR #1500, opened 2025-08-05
fix: `non-virtual-dtor` warning

qwertycxz:
This virtual destructor will resolve the warning of option `-Wnon-virtual-dtor` in MinGW.

Fixes: #1499


sylveon:
You are not meant to `delete` classes which derive from IUnknown. They automatically delete themselves once the ref count reaches 0. Adding the virtual destructor changes the offset of every function in interfaces that derive from IUnknown, breaking the COM ABI.

The correct resolution here is to suppress the warning, not declare the virtual dtor.

PR #1503, opened 2025-08-19
Update com_ptr dereference implementation to account for proyected types

MiguelBarro:
The current implementation:
```c++
    T& com_ptr<T>::operator*() const noexcept
    {
        return *m_ptr;
    }
```
cannot work for proyected types because `abi_t<T> != T`.
Note that proyections provide `abi_t` specializations. By default `abi_t<T> == T` and the operator works.

The fix takes into account two cases:
+ The `com_ptr<>` wraps a COM/WinRT interface, that is `T != abi_t<T>`.
    ```
    com_ptr
      ┕> m_ptr = impl::abi_t<T>*
    ``` 
   Because the projected wrapper layout is:
    ```
    IXXX
      ┕> m_ptr = impl::unknown_abi*
    ```
   Then a plain cast would do: `return *reinterpret_cast<IXXX*>(const_cast<com_ptr*>(this));`

+ The `com_ptr<>` wraps an implementation object, that is `T == abi_t<T>`. E.g. an `event_array`:
    ```
    com_ptr
      ┕> m_ptr = EventHandler<int>*
    ```
    return a reference to the member object: `return *const_cast<EventHandler<int>*>(m_ptr);`

A regression test is included.

### DIFF
--- a/strings/base_com_ptr.h
+++ b/strings/base_com_ptr.h
@@ -112,7 +112,10 @@ WINRT_EXPORT namespace winrt
 
         T& operator*() const noexcept
         {
-            return *m_ptr;
+            if constexpr (std::is_same_v<T, type>)
+                return *const_cast<T*>(m_ptr);
+            else
+                return *reinterpret_cast<T*>(const_cast<com_ptr*>(this));
         }
 
         type* get() const noexcept
--- a/test/old_tests/UnitTests/com_ptr.cpp
+++ b/test/old_tests/UnitTests/com_ptr.cpp
@@ -385,3 +385,27 @@ TEST_CASE("com_ptr, compare")
     REQUIRE(!(nullptr != c));
     REQUIRE(!(c != nullptr));
 }
+
+TEST_CASE("com_ptr, calls")
+{
+    hstring test_string;
+
+    // here com_ptr<Stringable>::type == Stringable
+    const wchar_t * grettings = L"Hello world!";
+    com_ptr<Stringable> impl = make_self<Stringable>(grettings);
+
+    // -> calls the projection
+    REQUIRE_NOTHROW(test_string = impl->ToString());
+    REQUIRE(grettings == test_string);
+
+    // here com_ptr<IStringable>::type == abi_t<IStringable>
+    com_ptr<IStringable> str(detach_abi(impl.as<IStringable>()), take_ownership_from_abi);
+
+    // -> calls the abi
+    REQUIRE(0 == str->ToString(put_abi(test_string)));
+    REQUIRE(grettings == test_string);
+
+    // * calls the projection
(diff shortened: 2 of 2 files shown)


sylveon:
You are not meant to wrap projection types in `com_ptr`. Projection types already implement reference counting semantics.

MiguelBarro:
`com_ptr` looks like the WinRT version of the ATL's [CComPtr](https://learn.microsoft.com/en-us/cpp/atl/reference/ccomptr-class?view=msvc-170) or the [_com_ptr_t](https://learn.microsoft.com/en-us/cpp/cpp/com-ptr-t-class?view=msvc-170) extension.
That is, it provides direct access to the underlying COM APIs.

The projection wrappers generated by `cppwinrt` provide more convenient modern C++ ones and exception handling of
errors.

It seems to me they seek different goals. Besides, note how `com_ptr` uses `abi_t<T>` for the underlying pointer.
It was designed with projected classes in mind too.

This pull request is a workaround for an issue I found using delegates.
The projected types have a constructor from `com_ptr`, for example:
```c++
    template <typename O, typename M>
    AsyncActionCompletedHandler::AsyncActionCompletedHandler(com_ptr<O>&& object, M method)
        : AsyncActionCompletedHandler([o = std::move(object), method](auto&&... args)
            { return ((*o).*(method))(args...); })
    {}
```
But it failed to compile. Eventually I discovered that the issue was that `O` was a projection type.
And that breaks `*o` because `O != abi_t<O>`.
That is frustrating for the user because `object` should be a valid argument.

A possible workaround would be make `cppwinrt` replace this lambda with another that takes this issue into account.
I tried using the [impl:com_ref](https://github.com/microsoft/cppwinrt/blob/a338c862c43e16ee8c3ba77bcae92d821be2deb5/strings/base_windows.h#L70) alias but as mentioned above `com_ptr` and the projections do not share APIs and `*o` is not valid for a projection
class.
I think fixing this issue in the `com_ptr` implementation is easier and would fix another similar scenarios not yet
uncovered.

kennykerr:
Is there a specific problem you're trying to solve?

sylveon:
Those handlers are meant to pass an implementation type, see [here](https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/weak-references#if-you-use-a-member-function-as-a-delegate).

I'm not so sure `com_ptr` was meant to be used this way with projection types, as you can spot the following line in `get_runtime_activation_factory_impl`: `com_ptr<abi_t<Windows::Foundation::IActivationFactory>> library_factory;`

kennykerr:
To be clear, I designed `com_ptr` for ABI types. Projected types handle this directly, as @sylveon mentioned.

Issue #1511, opened 2025-09-17
Bug:map_view_base Lookup failure case is several orders slower due to C++ exceptions


ChrisGuzak:
### Version

v2.0.250303.1

### Summary

I created a synthetic benchmark to test the perf of the failed case of Lookup(), that generates a C++ exception (impl below hresult_out_of_bounds()), and it reports the following when compared to the OS implementation of ValueSet. Lookup is called in TryLookup to simplify usage. 

```
CppWinRT Map
TryLookup fails
100000 iterations, 2021ms, 0ms per iteration

TryLookup succeeds
100000 iterations, 5ms, 0ms per iteration

ValueSet
TryLookup fails
100000 iterations, 2ms, 0ms per iteration

TryLookup succeeds
100000 iterations, 3ms, 0ms per iteration
```

**2021ms vs 2ms**

Antonio is working on fixing this in [this PR](https://github.com/microsoft/cppwinrt/pull/1512).

```cpp
...
    struct map_view_base : iterable_base<D, Windows::Foundation::Collections::IKeyValuePair<K, V>, Version>
    {
        V Lookup(K const& key) const
        {
...
            if (pair == static_cast<D const&>(*this).get_container().end())
            {
                throw hresult_out_of_bounds();
            }

```

### Reproducible example

```cpp

```

kennykerr:
Related workaround in Rust to avoid the cost of C++ exceptions: https://github.com/microsoft/windows-rs/pull/3476

ChrisGuzak:
thanks for pointing that out. I commented in that completed PR but repeating that here; making 2 calls is fine for in proc, but OOP APIs, where the cost per call is high, that is a problem. Antonio is making progress on a fix for this.

PR #1512, opened 2025-09-17
Avoid throwing/originating errors in expected scenarios

antmor:
Fixes: #1511  #1376 

### Problem Statement:
The winrt consumer IMapView projection exposes a TryLookup function. This function calls .Lookup across the ABI::IMapView, a `winrt::impl::consume_Windows_Foundation_Collections_IMap`, then instead of throwing an expected E_BOUNDS result, it returns an optional nullopt on the failure HRESULT. This is an optimization to avoid the possible RPC of doing both a HasKey and a Lookup. 
The issue is that the winrt producer of IMapView, does not have such an optimization. it exposes the ABI::IMapView as `winrt::impl::map_impl`, which wraps a winrt::IMapView implementation, without a TryLookup. Therefore `winrt::map_view_base::Lookup`, is forced to throw an error, which the `winrt::impl::map_impl`, will catch, then convert to an HRESULT. This throw/catch is pessimistic, as it's an extra throw, and could affect performance, as it forces a RoOriginate in order to print to the debugger. Thus it causes bad performance AND a bad developer/debugging experience. 

Similarly, when an async task is cancelled, the await_adapter is expected to throw,, but the origination causes extra debug output. 

### Solution
- support a TryLookup function in the `winrt::map_view_base`
- enlighten `winrt::impl::map_impl` to know about `TryLookup` and call it if it is present
- create an winrt::hresult override that can throw, but not rooriginate
- enable `cancellation_token` to choose whether to create a winrt::hresult_cancelled with OR without an origination
- added a .runsettings file to add the ability to run the tests from Visual Studio (using a Catch2ToVisualStudioTest adapter)

Verified:
ran build_test_all.cmd
Ran tests in Visual Studio:
image

PENDING: perf test


jonwis (approved):

strings/base_error.h
```diff
@@ -329,6 +336,13 @@ WINRT_EXPORT namespace winrt
         hresult_canceled(take_ownership_from_abi_t, winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()) noexcept : hresult_error(impl::error_canceled, take_ownership_from_abi, sourceInformation) {}
     };
 
+    struct non_originating_hresult_canceled : hresult_error
+    {
+        non_originating_hresult_canceled(winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()) noexcept : hresult_error(impl::error_canceled, hresult_error::avoid_originate, sourceInformation) {}
```
dmachaj, line 341: Instead of a different struct I'm wondering if it would be cleaner to continue using `hresult_canceled` and fill in the `hresult_error::avoid_originate` parameter when it is appropriate.

strings/base_error.h
```diff
@@ -332,17 +332,11 @@ WINRT_EXPORT namespace winrt
     struct hresult_canceled : hresult_error
     {
         hresult_canceled(winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()) noexcept : hresult_error(impl::error_canceled, sourceInformation) {}
+        hresult_canceled(hresult_error::avoid_originate_t) noexcept : hresult_error(impl::error_canceled, hresult_error::avoid_originate) {}
```
dmachaj, line 335: I was going to say that this doesn't have source location support.  But I guess if it isn't originating then that doesn't really matter.

dmachaj (dismissed):

jonwis (dismissed):

oldnewthing:
The PR title and description are extremely misleading. It's not “Avoid throwing/originating errors in expected scenarios” because straight Lookup failure is not an expected scenario.

Should be “Remove RoOriginate from Lookup failure (usually expected)”.

The PR description is too focused on TryLookup, which be a red herring for people investigating a Lookup regression.

Should be "We are removing RoOriginate from the internal implementation of Lookup, which benefits TryLookup performance" and then describe how you're doing it.

test/old_tests/UnitTests/TryLookup.cpp
```diff
@@ -143,4 +143,48 @@ TEST_CASE("TryLookup TryRemove error")
...
 
-}
\ No newline at end of file
+}
+
+TEST_CASE("trylookup_from_abi specialization")
+{
+    // A map that throws a specific error, used to verify various edge cases.
+    // and implements tryLookup, to take advantage of an optimization to avoid a throw. 
+    struct map_with_try_lookup : implements<map_with_try_lookup, IMapView<int, int>>
```
oldnewthing, line 152: Need a test with `IMapView<int, IInspectable>` where the value is present but the value is `nullptr`.

strings/base_meta.h
```diff
@@ -10,6 +10,10 @@ WINRT_EXPORT namespace winrt
     struct take_ownership_from_abi_t {};
     inline constexpr take_ownership_from_abi_t take_ownership_from_abi{};
 
+    // Map implementations can implement TryLookup with trylookup_from_abi_t as an optimization
+    struct trylookup_from_abi_t {};
+    inline constexpr trylookup_from_abi_t trylookup_from_abi{};
```
oldnewthing, line 15: I'm not convinced that we need this. Can we just call TryLookup if it is present and returns `std::optional<T>`?

strings/base_error.h
```diff
@@ -84,6 +84,9 @@ WINRT_EXPORT namespace winrt
 {
     struct hresult_error
     {
+        struct avoid_originate_t {};
```
oldnewthing, line 87: I think "suppress" is better than "avoid". Or even `no_originate_t`.

strings/base_error.h
```diff
@@ -325,6 +332,7 @@ WINRT_EXPORT namespace winrt
     struct hresult_canceled : hresult_error
     {
         hresult_canceled(winrt::impl::slim_source_location const& sourceInformation = winrt::impl::slim_source_location::current()) noexcept : hresult_error(impl::error_canceled, sourceInformation) {}
+        hresult_canceled(hresult_error::avoid_originate_t) noexcept : hresult_error(impl::error_canceled, hresult_error::avoid_originate) {}
```
oldnewthing, line 335: Shouldn't we add this to all of the hresult_* types? I don't know if `slim_source_location` works with this consolidation

```cpp
struct hresult_blah : hresult_error
{
    template<typename...Args>
    hresult_blah(Args&&...args) : hresult_error(impl::error_blah, std::forward<Args>(args)...) {}
};
```

strings/base_coroutine_foundation.h
```diff
@@ -351,6 +351,10 @@ namespace winrt::impl
             return m_promise->enable_cancellation_propagation(value);
         }
 
+        bool avoid_cancel_origination(bool value = true) const noexcept
```
oldnewthing, line 354: I think we should flip to be positively-named
```cpp
bool originate_on_cancel(bool value = true) const noexcept
{
    return !m_promise->enable_avoid_cancel_origination(!value);
};
```

test/old_tests/UnitTests/TryLookup.cpp
antmor: implemented a new test, had to make it an explicit optional trylookup

strings/base_meta.h
ChrisGuzak: seems very unlikely right?

strings/base_meta.h
```diff
@@ -298,4 +302,18 @@ namespace winrt::impl
...
     };
+
+    template <typename D, typename K>
+    struct has_TryLookup
+    {
+        template <typename U, typename = decltype(std::declval<U>().TryLookup(std::declval<K>(), trylookup_from_abi))> static constexpr bool get_value(int) { return true; }
+        template <typename> static constexpr bool get_value(...) { return false; }
+
+    public:
+
```
ChrisGuzak, line 313: nit, delete blank line if you want.

strings/base_meta.h
antmor: i agreed with his suggestion as since is not a "breaking" change, any code that compiled before this change, should compile after this change. So if a custom IMap has TryLookup(key, extra args) or bool TryLookup(key, outval), we won't break compilation of their code by merely updating their cppwinrt versions.

strings/base_error.h
antmor: i think it makes sense that if there's no origination, there's no logging. so i removed the error. As far as the greater consolidation, does that make sense to make that change in this PR? would it be too big a diff for this change?

dmachaj (approved):

test/old_tests/UnitTests/TryLookup.cpp
DefaultRyan: I'm trying to decipher if there is also a test case that verifies that a map with a TryLookup lacking `trylookup_from_abi_t` does *not* have its TryLookup called.

DefaultRyan (approved):

Issue #1244, opened 2022-12-09
Mark `winrt::resume_foreground` and `co_await <dispatcher>;` with `[[deprecated]]`


sylveon:
Now that the recommended way to move to the foreground is through the wil library, we should mark these functions as deprecated, and point users to wil. Eventually, one might consider removing them entirely. Related to #1123 

Sidenote: should we do the same to apartment_context too? I heard from somewhere that it was pretty slow and usage of the dispatchers is preferred nowadays.

JaiganeshKumaran:
I shouldn't need to install a whole other library for basic stuff. Windows App SDK should've provided co_await support by itself.

sylveon:
You can always manually call TryEnqueue.

kennykerr:
While there are good alternatives, I'm not sure its practical to deprecate these as there is a lot of existing code that uses them, and that code isn't necessarily wrong, and this would force all of that code to be rewritten or pick up a new dependency. I had hoped to remove these for a future v3 but that seems unlikely to happen any time soon.

sylveon:
The deprecated attribute just generates a compiler warning that can easily be ignored (or we could add a macro to disable the warning), and we can add a text blurb to the warning pointing to WIL. It also helps migration because there is an evident warning on all call sites. Code doesn't need to be rewritten until we actually remove it (if ever).

I still support porting the wil changes to cppwinrt and make supported extension points for WinUI 3 to hook into (no explicit support of WinUI 3 is required from our side). That way, we wouldn't have to remove it, require rewrites, or require adding a new dependency.

kennykerr:
Unfortunately, some of our internal builds turns warnings into errors and projects have no control over suppressing them. Primarily, this is how Windows is built and largely why we don't use the deprecated attribute for any code destined for the OS.

sylveon:
Couldn't they do, say,
```cpp
#define WINRT_SILENCE_DEPRECATED_RESUME_FOREGROUND
#include "winrt/base.h"
```

kennykerr:
Sure, but we have a rather enormous code base and it would have to be done for every project which is not practical. The idea of deprecating APIs has just never taken off for this reason. WinRT APIs actually include metadata indicating APIs that are deprecated but it's completely ignored for the same reason.

sylveon:
Understandable. This sounds like a not starter for now, would probably happen only if cppwinrt v3 ever realizes.

PR #1529, opened 2026-01-07
Remove implicit hstring constructor for string_view

### LINKED ISSUE #1527: Bug: `winrt::box_value(std::wstring_view(...))` may cause abort
### Version

_No response_

### Summary

I'm bitten by this for the second time, and it's not pleasant to debug. I have code like:
```cpp
Blabla(winrt::box_value(GetStringView(...)));
```
It seems that if `GetStringView` returns `std::wstring_view` which happens to end with a null terminator, it works, but if not, then `abort` is called and the app crashes.

Ideally it should just work correctly. Less ideally, it shouldn't compile without an explicit `hstring(...)` conversion. Even less ideally, it should always crash. But now, it crashes only in the less common cases, which bits me in the most inconvenient times.

I don't have an isolated repro right now, I'm working on a large codebase, let me know if you need one.

### Reproducible example

```cpp

```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional comments

_No response_

justanotheranonymoususer:
Fixes #1527

In C++, `string_view` has no notion of a null terminator, which is assumed by `create_string_reference`.


sylveon:
While I agree with the intent here, this is source breaking so unlikely to get merged.

I'm also not sure what's even the point of an overload taking an explicit size parameter. A constructor that takes in an explicit size suggests that the string doesn't need to be null-terminated, just like `std::wstring_view` did, so we're not gaining anything. Also param::hstring is meant to hide/abstract this fast-pass behavior from the user, and an explicit constructor call is notably not very hidden.

If a source-breaking change was possible, I'd just completely eliminate this overload. For now, I'd go with the suggestion from @DefaultRyan here to update `box_value` to take in `hstring` instead of `param::hstring`.

justanotheranonymoususer:
> this is source breaking

I don't know the extent, but as I explained, it's likely incorrect and a UB, which is not good to have in a library.

If it has extensive usage, perhaps it can be deprecated instead. LMK.

> A constructor that takes in an explicit size suggests that the string doesn't need to be null-terminated, just like std::wstring_view did, so we're not gaining anything

I added it to assist migration. I agree that it should also be documented to make absolutely clear that the requirement is that `value[size]` can be accessed and is zero.

> update `box_value` to take in `hstring` instead of `param::hstring`.

Well, it's not only that, example:
```cpp
unbox_value_or(boxed, ReturnsStringView());
```

DefaultRyan (changes requested):

justanotheranonymoususer:
Would you agree to make it
https://en.cppreference.com/w/cpp/language/attributes/deprecated ?

YexuanXiao:
The maintainers don't seem to consider using [[deprecated]], as seen at https://github.com/microsoft/cppwinrt/issues/1244. Honestly, I think even with v3, C++/WinRT needs to implement its own null-terminated string view, because C++29 will be the earliest version to have it, and raising C++/WinRT's minimum standard to C++29 would happen after v3 is released.

justanotheranonymoususer:
Waiting for reply

justanotheranonymoususer:
Keepalive

YexuanXiao:
It appears that C++/WinRT is currently avoiding any breaking changes. Meanwhile, Kenny mentioned that his focus is no longer on C++ tools. Rather than waiting for the nearly impossible C++/WinRT 3.0, it's better to start taking action now. I have forked C++/WinRT and am planning some ambitious improvements. If you're interested in it, you can contribute to my fork: https://github.com/YexuanXiao/cppwinrtplus.

YexuanXiao:
I have added C++ module support for C++/WinRT in my fork and published it on NuGet. Feel free to check out my repo.

PR #1530, opened 2026-01-08
box_value constructor: Replace param::hstring with hstring

justanotheranonymoususer:
Partial fix for https://github.com/microsoft/cppwinrt/issues/1527.

Other places remain unfixed, example:
```cpp
unbox_value_or(boxed, ReturnsStringView());
```

Or any other `param::hstring` usage.

### DIFF
--- a/strings/base_reference_produce.h
+++ b/strings/base_reference_produce.h
@@ -23,37 +23,37 @@ namespace winrt::impl
             return std::is_arithmetic_v<T> || std::is_enum_v<T>;
         }
 
-std::uint8_t GetUInt8() const
+        std::uint8_t GetUInt8() const
         {
             return to_scalar<std::uint8_t>();
         }
 
-std::int16_t GetInt16() const
+        std::int16_t GetInt16() const
         {
             return to_scalar<std::int16_t>();
         }
 
-std::uint16_t GetUInt16() const
+        std::uint16_t GetUInt16() const
         {
             return to_scalar<std::uint16_t>();
         }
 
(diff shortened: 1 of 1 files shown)


justanotheranonymoususer:
Keep, I still hope for someone to look at it

justanotheranonymoususer:
Keepalive

justanotheranonymoususer:
@DefaultRyan should I close this?

sylveon:
@DefaultRyan This PR is a good one - box_value will always do a copy of the hstring so we might as well skip the `param::hstring` here.

DefaultRyan:
This breaks boxing of string-like types such as `std::wstring` and `std::wstring_view` via conversion to `param::hstring`.

I think we can still do this with a little more finesse. One idea of the top of my head is to have an exact `hstring` overload, and an overload for string-like types that are convertible to `param::hstring`. For the second overload, we don't actually create a `param::hstring`, but explicitly construct a fully-fledged `hstring` instead.

justanotheranonymoususer:
I tried to fix, can you run tests?

DefaultRyan:
> I tried to fix, can you run tests?

Did you know you can run the tests locally? That break would have been caught there and saved you some time. :)

justanotheranonymoususer:
I was filling lucky. Should pass now, and I even added tests.

justanotheranonymoususer:
@DefaultRyan it's green now, are we good?

justanotheranonymoususer:
thx for approving, can you also merge?

Issue #766, opened 2020-10-14
Use after free when accessing captures in a dispatched fire_and_forget lambda


sylveon:
When dispatching a fire_and_forget lambda through a DispatcherQueue, we get a use-after-free:
```cpp
#include <DispatcherQueue.h>

#define __cpp_coroutines // temporary workaround to get coroutines working in C++20 mode
#include <winrt/base.h>
#include <winrt/Windows.Foundation.h>
#include <winrt/Windows.System.h>

#include <chrono>
#include <iostream>
#include <memory>

winrt::Windows::System::DispatcherQueueController CreateDispatcherController()
{
	const DispatcherQueueOptions options = {
		.dwSize = sizeof(options),
		.threadType = DQTYPE_THREAD_CURRENT,
		.apartmentType = DQTAT_COM_STA
	};

	winrt::com_ptr<ABI::Windows::System::IDispatcherQueueController> controller;
	winrt::check_hresult(CreateDispatcherQueueController(options, controller.put()));
	return { controller.detach(), winrt::take_ownership_from_abi };
}

int main()
{
	auto controller = CreateDispatcherController();

	auto veryimportant = std::make_unique<int>(1337);
	controller.DispatcherQueue().TryEnqueue([stillveryimportant = std::move(veryimportant)]() -> winrt::fire_and_forget
	{
		using namespace std::chrono_literals;

		co_await 1s;

		// use-after-free here
		std::cout << *stillveryimportant;
	});

	MSG msg;
	while (GetMessage(&msg, nullptr, 0, 0))
	{
		TranslateMessage(&msg);
		DispatchMessage(&msg);
	}
}
```

The lambda internal `this` is freed on the first coroutine suspension point, meaning accessing captured variables after this point will lead to undefined behavior and crashes.

DefaultRyan:
This is not a bug in C++/WinRT, but is a consequence of the way you are using the lambda in conjunction with coroutines. The temporary instance of the lambda (along with its captures) is destroyed after the call to `TryEnqueue` returns, which is going to be at the first suspension point.

coroutines create a frame that holds the lifetime of the function's parameters and local variables and keeps them alive across suspension points. A lambda is not a coroutine - it is a C++ object, with an overloaded `operator()` member function. So the lifetime extension afforded to a coroutine's parameters and locals is not afforded to the lambda object itself. And because a lambda's captures are simply member variables of the object, they also get destroyed.

See these links for more discussion:
[...]

PR #1553, opened 2026-03-23
Deducing this for delegates

YexuanXiao:
When using lambda coroutines as delegates, care must be taken that the lifetime of captured objects may be shorter than that of the coroutine, see https://github.com/microsoft/cppwinrt/issues/766#issuecomment-708713575. This issue has a better solution in C++23, which is to use explicit object parameters to copy the captures into the coroutine frame. However, this feature is currently not available in C++/WinRT.

The following code fails to compile with an error that the copy constructor has been deleted.

```cpp
using namespace winrt::Windows::Foundation::Collections;

int main()
{
	int called{};
	int sum{};

	auto vector = winrt::single_threaded_observable_vector<int>();
	auto revoker = vector.VectorChanged(winrt::auto_revoke, 
		[called, sum](this auto, IObservableVector<int>, IVectorChangedEventArgs)
		{
		});
}
```

The reason is that the final type `delegate` inherits from the handler (lambda), and [code_writers.h#L2619](https://github.com/microsoft/cppwinrt/blob/b9553a9f0faf9a1395ac98aea65716c6f44e9fca/cppwinrt/code_writers.h#L2619) uses the type `delegate` (`*this` in the `Invoke` member function of `delegate`) to call `operator()`, causing `this` (explicit parameter of lambda) to be deduced as `delegate`. Since `delegate` is non-copyable, `operator()` cannot be called. This change makes `Invoke` use the original lambda's type to call `operator()`, so `this` is deduced as the lambda's type, resolving the issue.

Because of the purpose is to copy the captured object into the coroutine frame, the slicing that occurs here is intentional, and there is no need to copy the complete `delegate` object.


run_tests.cmd
```diff
@@ -11,6 +11,7 @@ if "%target_configuration%"=="" set target_configuration=Debug
 call :run_test test
 call :run_test test_nocoro
 call :run_test test_cpp20
+call :run_test test_cpp23
```
sylveon: Move this one line below?

test/CMakeLists.txt
```diff
@@ -107,6 +107,7 @@ set(SKIP_LARGE_PCH FALSE CACHE BOOL "Skip building large precompiled headers.")
 add_subdirectory(test)
 add_subdirectory(test_nocoro)
 add_subdirectory(test_cpp20)
+add_subdirectory(test_cpp23)
```
sylveon: And here.

YexuanXiao:
I found that when running the CI on my own fork, Clang 20 target x86 generated incorrect code, causing the test to fail. The characteristic of this bug is that in the constructor of observed_sender at L101, `this` is observed to be X, while in the assignment operator at L108, `this` is observed to be X+16 (in bytes). Therefore, this is undoubtedly a Clang bug rather than a test bug. As a result, I will disable the test `observablevector_vectorchanged` for Clang target x86.

YexuanXiao:
@DefaultRyan Could you merge this patch? It's frustrating that C++/WinRT can't work with this cool C++23 feature.

PR #1555, opened 2026-03-30
User/chriwall/deprecated

Dreynor87:
What's this all about?
This implements the deprecated attribute as described here: https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.deprecatedattribute?view=winrt-26100
Along with the DeprecationType enum as described here: https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.deprecationtype?view=winrt-26100

Fixes: WinRT Deprecation, that right now does not do anything.


sylveon:
Is it normal for _build to be there?

Dreynor87:
> Is it normal for _build to be there?
Thanks for noticing pointing out. I missed the extra files. I have removed them and added this directory to the gitignore so people don't accidentally put them back in.

DefaultRyan:
@Dreynor87 The CI run is showing some errors. Could you look into fixing those?

Dreynor87:
> @Dreynor87 The CI run is showing some errors. Could you look into fixing those?

I had the Agent look at those... As well as fix the wrap/pop stuff that was asked about.

Edit: another failure, but new. Agent looking again.

Edit again: After 1 more round, the agent got something that builds. I can't say I understand everything going on here, and agent made it all for me, mostly during FHL week, but if you understand it, it will make things in WindowsAppRuntime better.

cppwinrt/helpers.h
```diff
@@ -242,6 +242,58 @@ namespace cppwinrt
...
+        }
+        return result;
+    }
+
+    template <typename T>
+    auto get_deprecated_message(T const& row)
+    {
+        auto attr = get_attribute(row, "Windows.Foundation.Metadata", "DeprecatedAttribute");
+        return escape_cpp_string_literal(get_attribute_value<std::string_view>(attr, 0));
+    }
```
riverar, line 279: Should probably do nothing if there is no deprecated attribute.

cppwinrt/helpers.h
```diff
@@ -242,6 +242,58 @@ namespace cppwinrt
...
+            return false;
+        }
+        auto args = attr.Value().FixedArgs();
+        if (args.size() >= 2)
+        {
+            // DeprecationType enum: Deprecate=0, Remove=1
+            auto val = std::get<ElemSig>(args[1].value);
+            auto enum_val = std::get<ElemSig::EnumValue>(val.value);
+            // Compare the underlying integer value (1 == Remove)
+            return std::visit([](auto v) -> bool { return static_cast<int32_t>(v) == 1; }, enum_val.value);
```
riverar: This should probably use `get_integer_attribute` instead.

riverar (changes requested):
Added a few nits. However, I'm struggling to understand the fundamental value of this attribute.

`Deprecate(0)` will just warn me forever until all the Windows targets in-use are upgraded (~5-10 years) and breaks `/WX` builds on upgrade even though the API works fine and always will.

`Remove(1)` breaks source compat for no real reason. The ABI is still preserved, the app would run correctly, but now established codebases fail CI just because they updated cppwinrt. Or if you do have plans to break ABI, then just remove the item to begin with?

What's the intended outcome here?

Dreynor87:
> Added a few nits. However, I'm struggling to understand the fundamental value of this attribute.
> 
> `Deprecate(0)` will just warn me forever until all the Windows targets in-use are upgraded (~5-10 years) and breaks `/WX` builds on upgrade even though the API works fine and always will.
> 
> `Remove(1)` breaks source compat for no real reason. The ABI is still preserved, the app would run correctly, but now established codebases fail CI just because they updated cppwinrt. Or if you do have plans to break ABI, then just remove the item to begin with?
> 
> What's the intended outcome here?

WindowsAppRuntime, that uses WinRT, has a desire to deprecate APIs in 2 passes. 

1st pass, we would mark that the API is deprecated, but only give a warning to developers. We will then, in the warning message give them when the API will be "removed", and what they should use instead. The API will still be implemented, and callable. This is primarily kept this way for 2 reasons. 1. We want to give time for developers to respond to an API going away. 2. We have a guarantee that anything built with Major version X can run with Major version X + or - 1, as long as only the lower Major version APIs are called, and those APIs were not deprecated in the lower Major version.

2nd pass, we would then change the DeprecationType to Removed. This will now actually "remove". That remove cannot change the vTable(that would break all sorts of things beyond just this API going away), but the API should no longer show up in the header files when a developer is writing code, and the API should fail immediately if someone tries to call it.

We figure that some of this could be done in a WindowsAppRuntime specific way, but this feels like something any WinRT developer would like, so we are trying to do this the right way and make it a part of WinRT. As mentioned in my first comment, this was already all described here:https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.deprecatedattribute?view=winrt-26100 and here: https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.deprecationtype?view=winrt-26100

It just was not implemented.

YexuanXiao:
I am not opposed to using AI, but I am a bit particular. The new code has some inconsistencies in style, such as `if (is_removed(type)) return;`, which is not the style used in the current code. Secondly, `#pragma warning` is only supported by MSVC, so it should be guarded. I think you should carefully review it yourself, rather than leaving it entirely to the AI.

riverar:
> 1st pass, we would mark that the API is deprecated, but only give a warning to developers. We will then, in the warning message give them when the API will be "removed", and what they should use instead. The API will still be implemented, and callable. This is primarily kept this way for 2 reasons. 1. We want to give time for developers to respond to an API going away. 2. We have a guarantee that anything built with Major version X can run with Major version X + or - 1, as long as only the lower Major version APIs are called, and those APIs were not deprecated in the lower Major version.

OK, but lets consider a real developer scenario. I'm supporting Windows 11 and the extended lifetime SKUs for Contoso Inc. I upgrade cppwinrt and CI breaks. I now have a choice: Remove `/WX` and be told about a deprecated API for 10 years or suppress this warning (and very likely forget to remove it later) to restore `/WX` functionality.

> 2nd pass, we would then change the DeprecationType to Removed. This will now actually "remove". That remove cannot change the vTable(that would break all sorts of things beyond just this API going away), but the API should no longer show up in the header files when a developer is writing code, and the API should fail immediately if someone tries to call it.

If the API is going to break, then compat is out the window and you might as well just delete the API right? What's the point of this dance?

Dreynor87:
> OK, but lets consider a real developer scenario. I'm supporting Windows 11 and the extended lifetime SKUs for Contoso Inc. I upgrade cppwinrt and CI breaks. I now have a choice: Remove `/WX` and be told about a deprecated API for 10 years or suppress this warning (and very likely forget to remove it later) to restore `/WX` functionality.

As of now, WindowsAppRuntime (and likely the rest of the world) is not using the Deprecated Attribute because even though it is defined in WinRT, it is not implemented anywhere except for a partial implementation in CSWinRT. The hope is that if you are actively developing things, you would get this CPPWinRT update long before WindowsAppRuntime, or other projects, are actually going to use these attributes in the IDL. Then, when you get a new set of winmds from WindowsAppRuntime, or another project, and that has deprecated an API, you would then make a choice of changing your API right away, or supressing that warning and creating a bug to fix it later. If you forget to do that until the next Major version, when it is removed, then you will notice the second the suppressed warning turns into an error.


> If the API is going to break, then compat is out the window and you might as well just delete the API right? What's the point of this dance?

I am not a CPPWinRT person, but talking with some of those people, they all mentioned that if you were to actually remove the API from an Interface (removing an entire interface from a runtime class is fine...) then it causes vTable lookup issues for anyone who built with the old IDL files/winmds, and run on a version that does not have that API. @Scottj1s is the one who mentioned this as a potential issue.

YexuanXiao:
Libraries written by third-party developers may deprecate or remove APIs, so cppwinrt should add this feature. However, the code format needs to be corrected.

Dreynor87:
> Libraries written by third-party developers may deprecate or remove APIs, so cppwinrt should add this feature. However, the code format needs to be corrected.

I will get the formatting fixed up in the next iteration.

cppwinrt/helpers.h
Dreynor87: I tried to do that, and it broke the build, so I reverted it.

YexuanXiao:
Just representing my own habits: I’m not a fan of complex test scripts. I prefer simple and clear ones. For example, if we want to test deprecated properties, we can set unique messages for each case and directly verify whether these messages appear once in the generated files. If we want to test removed properties, we should directly verify what should and shouldn’t appear. At the same time, add a positive test to ensure the generated files are usable. Current test scripts have too much unnecessary noise, and I don’t think we need to verify failed builds.

PR #1559, opened 2026-04-05
Add [[nodiscard]] to overloads of static events that return winrt::event_token

YexuanXiao:
Some events in WinRT are static, such as [Clipboard.ContentChanged](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.datatransfer.clipboard.contentchanged?view=winrt-26100).For these events, the only way to release a delegate is through `winrt::event_token`, because the delegate is not stored in any user-controllable object. Therefore, if the `winrt::event_token` returned by these events is discarded, the delegate will leak. This PR adds the `[[nodiscard]]` attribute to the overloads that return `winrt::event_token` for static events to indicate that their return value should not be ignored, just like the overloads that return a revoker.

```cpp
struct Clipboard
{
    [[nodiscard]] static auto ContentChanged(EventHandler<IInspectable> const& handler);
};
```

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -3157,11 +3157,12 @@ struct WINRT_IMPL_EMPTY_BASES produce_dispatch_to_overridable<T, D, %>
             method_signature signature{ method };
             auto method_name = get_name(method);
             auto async_types_guard = w.push_async_types(signature.is_async());
+            auto is_event = is_add_overload(method);
 
             if (is_opt_type)
             {
                 w.write("        %static % %(%);\n",
-                    is_get_overload(method) ? "[[nodiscard]] " : "",
+                    is_get_overload(method) || is_event ? "[[nodiscard]] " : "",
                     signature.return_signature(),
                     method_name,
                     bind<write_consume_params>(signature));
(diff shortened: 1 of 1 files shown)


sylveon:
Users may want to permanently subscribe to an event for the lifetime of their program, in which case discarding the event token seems intentional.

YexuanXiao:
It does not affect non-static events, and their overloads that return winrt::event_token still do not have this attribute. Non-static events can be released through the object they are associated with, so discarding the return value does not cause a leak, whereas static events do not have such an object. The impact of this patch is very small; if users wish to discard the return value, they can use (void)expr.

sylveon:
I think we should only add [[nodiscard]] if dropping the value is always considered an error

YexuanXiao:
I'm not entirely sure which static events might need to be subscribed to permanently, but at least the Clipboard [example](https://github.com/microsoft/Windows-universal-samples/blob/082195895276903b6630d5cb4d03c9d365ec210c/Samples/Clipboard/cppwinrt/HistoryAndRoaming.cpp#L34-L45) performs proper cleanup and I don't believe that well-written code should allow such a leak.

PR #1578, opened 2026-04-28
Update dev build cycle time

jonwis:
Noodling on some quality-of-life enhancements for dev builds.

* Enable running from a "plain" cmd window - finds `VsDevCmd` using `vswhere.exe`
* Combine msbuild invocations into one for multithreaded build
* Track errors in each invocation and stop if anything broke

### DIFF
--- a/cppwinrt/cppwinrt.vcxproj
+++ b/cppwinrt/cppwinrt.vcxproj
@@ -311,4 +311,9 @@
     </PropertyGroup>
     <Error Condition="!Exists('..\packages\Microsoft.Windows.WinMD.1.0.210629.2\build\native\Microsoft.Windows.WinMD.props')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\Microsoft.Windows.WinMD.1.0.210629.2\build\native\Microsoft.Windows.WinMD.props'))" />
   </Target>
+  <ItemGroup>
+    <ProjectReference Include="..\fast_fwd\fast_fwd.vcxproj">
+      <Project>{A63B3AD1-AB7B-461E-9FFF-2447F5BCD459}</Project>
+    </ProjectReference>
+  </ItemGroup>
 </Project>
--- a/ensure_vsbuild.cmd
+++ b/ensure_vsbuild.cmd
@@ -1,0 +1,70 @@
+@echo off
+
+set "target_platform=%~1"
+if "%target_platform%"=="" set "target_platform=x64"
+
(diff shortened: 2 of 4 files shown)


jonwis:
(Ignore the branch name, it's the third cppwinrt branch I have locally...)

cppwinrt/cppwinrt.vcxproj
```diff
@@ -311,4 +311,9 @@
     </PropertyGroup>
     <Error Condition="!Exists('..\packages\Microsoft.Windows.WinMD.1.0.210629.2\build\native\Microsoft.Windows.WinMD.props')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\Microsoft.Windows.WinMD.1.0.210629.2\build\native\Microsoft.Windows.WinMD.props'))" />
   </Target>
+  <ItemGroup>
```
DefaultRyan, line 314: What's the purpose of this ProjectReference? Should it instead be a build dependency in the solution?

ensure_vsbuild.cmd
```diff
@@ -0,0 +1,70 @@
+@echo off
+
+set "target_platform=%~1"
+if "%target_platform%"=="" set "target_platform=x64"
+
+set "vsdevcmd_arch="
+if /I "%target_platform%"=="x86" set "vsdevcmd_arch=x86"
```
dunhor, line 7: If host architecture is x64, it's probably best to use x64_x86 to use the x64 compiler and get the advantage of a larger address space.

And if you want to go the extra mile, you should probably do the same for arm64 so that you can cross-compile arm64 on Windows (though this may not be super useful given that it's called from the 'build_test_all' script, so the testing part wouldn't work out so well.

cppwinrt/cppwinrt.vcxproj
jonwis: In other solutions, I've been finding it better to use a <ProjectReference/> so I can say things like `msbuild foo.vcxproj` and have it figure out the dependency chain for me.

ensure_vsbuild.cmd
jonwis: Sure, I'll check it out.  I agree that the 'run tests cross arch' probably isn't going to work out a certain matrix.

dunhor: Though I do realize now this is calling `VsDevCmd` instead of `vcvarsall`, so maybe it's handled differently there.

PR #1579, opened 2026-05-09
Reduce non-default-interface costs for complicated runtimeclasses

jonwis:
# Super Summary

"Thunked" or "flattened" runtimeclasses decrease binary size and increase throughput by reducing the effort required to call non-default interface members on runtimeclass types.  When calling a method on a non-default interface, a place holder vtable proxy swaps itself for the "real" interface obtained from the default interface. Those "real" interfaces are cached inside the runtimeclass instance until the class is destroyed.

Certain kinds of types are excluded - classes with fast_abi metadata or which are unsealed. These limitations may be addressed at a later time.

# Use

1. Use `cppwinrt.exe` in the usual way
2. For NuGet consumers, the `*.asm` thunk files are pre-compiled as libraries similar to the fast_abi.lib per-architecture files
3. Rebuild clean

This feature is turned on with the `-thunked_classes` switch, which is off by default.

# Background

A type like `ValueSet` implements a number of interfaces.  Most developers write code like this:

```c++
PropertySet v;
v.Insert(L"kittens", box_value(true));
v.Insert(L"puppies", box_value(false));
for (auto& [k, v] : v) {
    tell_people(k, v.as<bool>());
}
```

Under the covers, each call to `.Insert` and the use of `.First` behind range-based `for` is a sequence of:

1. QueryInterface from IPropertySet (the default marker interface)
4. Call `IMap<String, Object>::Insert`
5. Release the temporary from step 1

This produces a nontrivial amount of code for composite objects, those with interfaces with `requires`, and those that have gone through multiple versioning cycles.  While QI and Release are fast-ish, they're still cacheline flushes (interlocked operations) and code the processor has to use. Windows heavily relies on runtimeclass contract versioning, producing types that grow in complexity over time.

# Mechanism

See `docs/runtimeclass-caching.md` for the full details. Projected runtimeclasses contain up to 8 "cached slots" for their non-default interfaces. Each slot acts like another COM instance - a pointer to a thunk vtable, plus a pointer to the parent class type combined with a slot index. 

When any method on the thunk vtable is invoked, the stub calls out to the parent class's cache to resolve its slot.  "Resolve" means calling `defaultInterface->QueryInterface(iid, &result)`, then using racy-init to swap the result of the QI into the projected type's interface pointers. Future calls on the same runtimeclass for the same interface then go through the "real" interface instance, rather than the thunk stub.

# Notable Changes

This mechanism is a departure from the `sizeof(runtimeclass)==sizeof(void*)` model. A runtimeclass instance is no longer just a container for a `winrt::com_ptr<abi_t<TDefaultInterface>>` ... runtimeclass instances grow in size to contain up to 8 slots, along with the increased code required to make full-value copies.  `ValueSet v = otherValueSet` copies only the default interface, causing a re-query for all the slots.

Many C++/WinRT functions that expected this pointer-is-the-default-interface model changed to accomodate


strings/cached_thunks_x64.asm
```diff
@@ -0,0 +1,136 @@
...
+    test    rax, rax
+    jz      resolve_failed
+
+    ; rax = resolved IFoo*. Load method ptr into rax for CFG validation.
+    mov     rcx, rax                ; rcx = new this (resolved IFoo*)
+    mov     r11, [rax]              ; r11 = resolved vtable
+    mov     rax, [r11 + r10 * 8]   ; rax = method at vtable[slot]
+
+    ; Verify indirect call target (preserves all GPRs except rax/flags)
+    call    [__guard_check_icall_fptr]
```
oldnewthing, line 56: I don't think you can issue call instructions on a non-paragraph aligned stack.

.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         run: |
           $target_configuration = "${{ matrix.config }}"
           $target_platform = "${{ matrix.arch }}"
-          & "_build\$target_platform\$target_configuration\cppwinrt.exe" -in local -out _build\$target_platform\$target_configuration -verbose
+          & "_build\$target_platform\$target_configuration\cppwinrt.exe" -in local -out _build\$target_platform\$target_configuration -verbose -flatten_classes
```
dmachaj, line 91: This should probably be a matrix config instead of all runs.

strings/base_meta.h
```diff
@@ -184,10 +184,50 @@ namespace winrt::impl
     struct WINRT_IMPL_EMPTY_BASES base : base_one<D, I>...
     {};
 
+    // ========================================================================
+    // Traits for detecting thunked runtimeclasses (C++17-compatible)
```
dmachaj, line 188: I think you should have a pragma ODR mismatch detector.  Mixing flat and non-flat ABI when linking is probably going to be catastrophic.

strings/base_thunked_runtimeclass.h
```diff
@@ -0,0 +1,436 @@
...
+
+    // ========================================================================
+    // Typed template: adds interface accessors, wires up lifecycle
+    // ========================================================================
+
+    template <typename IDefault, typename... I>
+    struct thunked_runtimeclass : thunked_runtimeclass_base
+    {
+        static constexpr size_t N = sizeof...(I);
+        static constexpr bool use_tagged = N <= 8;
```
dmachaj: How was the number 8 chosen for cache slots?  Is that guided by something like PropertySet so that it fits (or mostly fits)?

Also should this be a constexpr define somewhere so that if it needs tweaking there is only one line to update?

strings/cached_thunks_x64.asm
```diff
@@ -0,0 +1,136 @@
...
+
+; Emit 256 stubs (slots 0-255)
+counter = 0
+REPT 256
+    WINRT_CACHED_THUNK %counter
+    counter = counter + 1
+ENDM
+
+; ============================================================================
+; Vtable array: slots 0-2 are no-op IUnknown, slots 3-255 are resolve stubs
```
dmachaj: Where is the test case that inherits from 255 COM interfaces from a single runtimeclass 😅

test/test_cachedrtc/thunked_dispatch.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+__declspec(noinline) bool thunked_propertyset_haskey(PropertySet const& ps, hstring const& key)
+{
+    return ps.HasKey(key);
+}
+
+TEST_CASE("thunked_dispatch")
+{
+    PropertySet ps;
+    REQUIRE(ps);
```
dmachaj, line 46: Is there a way to static assert that the thunks are active on `ps` and that you aren't accidentally testing the previous approach?

test/Directory.Build.targets
```diff
@@ -0,0 +1,20 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Enable MASM assembler support -->
+  <Import Project="$(VCTargetsPath)\BuildCustomizations\masm.props" />
+
+  <!-- Link thunk stubs into all test binaries that use the cppwinrt projection -->
+  <ItemGroup>
+    <MASM Include="$(MSBuildThisFileDirectory)..\strings\cached_thunks_x64.asm" Condition="'$(Platform)'=='x64'" />
+    <MASM Include="$(MSBuildThisFileDirectory)..\strings\cached_thunks_x86.asm" Condition="'$(Platform)'=='Win32'" />
```
dmachaj, line 20: What about ARM64?

strings/base_activation.h
```diff
@@ -540,9 +540,18 @@ WINRT_EXPORT namespace winrt
...
             T ActivateInstance() const
             {
-                IInspectable instance;
-                check_hresult((*(impl::abi_t<IActivationFactory>**)this)->ActivateInstance(put_abi(instance)));
-                return instance.try_as<T>();
+                if constexpr (impl::has_thunked_cache_v<T>)
+                {
+                    void* result{};
+                    check_hresult((*(impl::abi_t<IActivationFactory>**)this)->ActivateInstance(&result));
+                    return{ result, take_ownership_from_abi };
```
sylveon: Doesn't this drop a typecheck that `instance.try_as` had?

sylveon:
Why was this solution preferred over expanding fast ABI to more types in `Windows.*`, like the example `PropertySet`? This is projection specific while fast ABI would benefit every projection. I'm also not sure it's the greatest idea to automatically enable this for existing projects, as there are significant implications over making a type suddenly 8 times its original size when it previously guaranteed to be pointer sized - as showed by all the cppwinrt changes needed.

I'd suggest making it off by default for existing projects and instead turning it on in the new project templates, if this solution is pursued.

strings/cached_thunks_x64.asm
jonwis: Hmm. Seems to work; this is basically cloning the fast_abi model. I'll check again.

strings/base_meta.h
jonwis: Updated.

strings/base_thunked_runtimeclass.h
jonwis: There's a note in the implementation docs; the lower 3 bits of the "pointer to parent object" are an index. Turns out that 8 covers most things (except PackageManager and Package) in the public API surface. There's _probably_ a way to eliminate this index entirely if we don't mind a little subtraction/division (subtraction shift? Hmm) in this path.  I'll see what can be done to eliminate this quirk.

strings/cached_thunks_x64.asm
jonwis: This is "count of methods on an interface" not "count of interfaces" ... I should add a validation, though, that we aren't trying to project more than this. The FastABI path has a similar feature, I'll clone it here.  (Internally, the "fast NDR proxy" work in combase/rpcrt4 assumes COM types won't have more than 4k methods, or it falls back to the slow path.)

test/test_cachedrtc/thunked_dispatch.cpp
jonwis: Interesting, yes, I'll add an enforcement here.

test/Directory.Build.targets
jonwis: Oops, missed. Will re-add.

strings/base_activation.h
jonwis: Hmm, yes - https://learn.microsoft.com/en-us/windows/win32/api/activation/nf-activation-iactivationfactory-activateinstance - does not require that the returned IInspectable was specifically from the default interface of the type, so (in theory) `WackyFactory` could return an `IInspectable` whose `QI` went and grabbed a static instance of an `IWacky`, rather than directly returning that `IWacky`.  I _think_ this can be removed now anyhow; there were some gyrations in this path, and "get the IInspectable, QI it to the cached runtimeclass object" should just work.  I'll see which other custom "if thunked cache" ones can be removed.

jonwis:
> Why was this solution preferred over expanding fast ABI to more types in `Windows.*`, like the example `PropertySet`? This is projection specific while fast ABI would benefit every projection. I'm also not sure it's the greatest idea to automatically enable this for existing projects, as there are significant implications over making a type suddenly 8 times its original size when it previously guaranteed to be pointer sized - as showed by all the cppwinrt changes needed.
> 
> I'd suggest making it off by default for existing projects and instead turning it on in the new project templates, if this solution is pursued.

Looks like FastABI works only for exclusiveto interfaces on a runtimeclass at the moment. It requires both the implementation and the projection to agree on fastABI-ness.

The main offenders that I see in terms of verbose codegen are things like PropertySet, ValueSet, Json*, and a set of internal-only types that are defined in terms of ABI with a deep versioning tree.  Composable types - `Composition*` and XAML's `UIElement` tree also hit the same "every method is a qi+call+release" sequence. In some methods, this traffic is something like 20% of the size of a binary (including all the unwind code to handle exceptions before the release.)

This approach is add-on to any type, and doesn't require updates to the implementation at all, and is ready to use.

Increasing the RTC size is definitely a concern. Previous attempts included a model where these "cached" types went into some child scope, like `winrt::Namespace::TypeName::cached` (as a using), or `winrt::cached<winrt::Namespace::TypeName>` as a specialization, etc. It retains the original layout, with an opt-in.  Even more options include projection-by-name, where you'd have a `cached_runtimeclasses.txt` passed to cppwinrt containing the types for which to emit a `fast<T>` wrapper (to reduce code spray.)

I've added a switch so it's not the default, and I'll remove it from the targets so you have to say `<CppWinRTCachedRuntimeclass>true` to enable.

strings/cached_thunks_x64.asm
riverar, line 56: Does this work in `/guard:cf-` scenarios?

riverar: Never mind, it works.

jonwis:
Should this just be a "mode" of fastabi, like "best effort fastabi"? If you want better fastabi, use the [FastABI] attribute and structure your implementation a certain way. If you don't, you get this "pretty OK" version.

@sylveon I've also been trying to figure out how to address the "sizeof(runtimeclass) != sizeof(void*)" concerns... the only thing that comes to mind is having the "real" cache be stored in a `std::shared_ptr<>` and the runtimeclass provides value-like behavior (the `foo.Bar(5)` instead of `foo->Bar(5)`) ... it does mean an additional external reference count and an external heap allocation, but preserves the _literal_ size of the projected types.  Thoughts?

Some nondefault interfaces probably don't need to be cached as they are rarely _reused_ on a type:

* ICloseable - closing something more than once is strange
* IStringable - not super common in the codebase, but also often a "use once to extract content" kind of thing

I'll see if I can get Tempo to hand out the set of non-exclusiveto-interfaces in common use and see what others stick out.

sylveon:
It's okay if the change is intrusive (like by increasing the size), but it needs to be opt-in so that developers can review and ensure their code is able to work with the change before flipping it on, instead of it getting flipped on automatically by a dependency update and cause hard to track issues.

YexuanXiao:
I truly hope that the optimizations for QueryInterface, AddRef, and Release are implemented by the compiler.

jonwis:
> I truly hope that the optimizations for QueryInterface, AddRef, and Release are implemented by the compiler.

Unfortunately, this is a design consequence of WinRT and something all projections have to work around.  When we decided that WinRT interfaces may only inherit from `IInspectable` and that a `runtimeclass` can be updated by adding a new interface with the new methods on it, the result is that a `runtimeclass` like `Windows.UI.Composition.Spotlight` is composed of ~8-10 interfaces.

In another world, we would have allowed `ICompositionObject4 : ICompositionObject3` derivation, so the number of interfaces is related to the number of base types.  We could have said that `ISpotlight : ICompositionObject` to reduce the number slightly.  However, since `Spotlight` and `CompositionObject` have their own versioning stories, you end up with at least `ISpotlightN : ISpotlightNMinusOne` and `ICompositionObject : ICompositionObjectNMinusOne`.

Further, C++/WinRT's excellent decision to be "value like" with dottable members means to simulate the same C++ derivation pattern we have all the current base conversion types, so when you down (up?) pass a `SpotLight` to member that takes `CompositionObject const&` that's a QI from `ISpotLight` to `ICompositionObject`.  All of this is hidden in the C++/WinRT codegen layers (the produce/consume pattern, the `operator I()`, etc.)

I've tried and discarded several options to make this better.  Really the only things I can come up with are radical departures from where we are today, like generating a `pimpl` style system that provides a "flattened vtable with proper inheritance" for version-related types, or this "swap the vtables so you pay one qi/release pair per composed interface."

FWIW, building a Lottie sample (https://github.com/CommunityToolkit/Lottie-Windows) with the updated cppwinrt in flattened-RTCs mode gets me a 5% binary savings, and Lottie is a pathologically bad environment for WinRT use.

sylveon:
> The main offenders that I see in terms of verbose codegen are things like PropertySet, ValueSet, Json*, and a set of internal-only types that are defined in terms of ABI with a deep versioning tree. Composable types - Composition* and XAML's UIElement tree also hit the same "every method is a qi+call+release" sequence. In some methods, this traffic is something like 20% of the size of a binary (including all the unwind code to handle exceptions before the release.)

These are all [exclusiveto] interfaces tho, right? Perfect candidate for FastAbi. I think it would make more sense to try and enable fast ABI there first and see if this is still needed after?

PR #1584, opened 2026-05-22
Massive speedup to file_equal and file_to_string

DefaultRyan:
I was testing some quality of life improvements, and I saw that in incremental situations when there are already headers in the output folder, cppwinrt.exe takes twice as long to run!

Before writing pending output, it reads the existing file in and compares it, skipping the write if the contents are identical. The comparison itself is fast enough, using `std::equal` which boils down to `memcmp`, but the read is using `std::ifstream::operator<<` to read a `std::string`, and this operation processes the file contents a single character at a time, resulting in tons of overhead.

With this change, we now do a single bulk read of the file contents. Additionally, before we read the contents, we check the file size in the filesystem *before* reading the contents, allowing us to short-circuit even faster.

A before/after comparison on a Release x64 run of regenerating the local SDK projection without cleaning the output folder first, shows a 50% speedup in wall clock time.

Here's the results from an instrumented profile run in Visual Studio.
Before:
image
image

After:
image
image

### DIFF
--- a/cppwinrt/text_writer.h
+++ b/cppwinrt/text_writer.h
@@ -13,8 +13,12 @@ namespace cppwinrt
 {
     inline std::string file_to_string(std::string const& filename)
     {
-        std::ifstream file(filename, std::ios::binary);
-        return static_cast<std::stringstream const&>(std::stringstream() << file.rdbuf()).str();
+        std::ifstream file(filename, std::ios::binary | std::ios::ate);
+        const auto size = file.tellg();
+        file.seekg(0);
+        std::string result(size, '\0');
+        file.read(result.data(), size);
+        return result;
     }
 
     template <typename T>
@@ -218,17 +222,12 @@ namespace cppwinrt
 
         bool file_equal(std::string const& filename) const
         {
-            if (!std::filesystem::exists(filename))
+            if (!std::filesystem::exists(filename) || std::filesystem::file_size(filename) != m_first.size() + m_second.size())
             {
                 return false;
             }
 
             auto file = file_to_string(filename);
-
-            if (file.size() != m_first.size() + m_second.size())
-            {
-                return false;
-            }
 
             if (!std::equal(m_first.begin(), m_first.end(), file.begin(), file.begin() + m_first.size()))
             {
--- a/cppwinrt/text_writer.h
+++ b/cppwinrt/text_writer.h
@@ -14,10 +14,24 @@ namespace cppwinrt
     inline std::string file_to_string(std::string const& filename)
     {
         std::ifstream file(filename, std::ios::binary | std::ios::ate);
-        const auto size = file.tellg();
+        if (!file) { return{}; }
+
+        const auto stream_size = file.tellg();
+        if (stream_size == std::ifstream::pos_type(-1))
+        {
+            return {};
+        }
+
         file.seekg(0);
+
+        auto size = static_cast<std::size_t>(stream_size);
         std::string result(size, '\0');
         file.read(result.data(), size);
+        if (!file)
+        {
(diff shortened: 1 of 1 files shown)


cppwinrt/text_writer.h
```diff
@@ -13,8 +13,12 @@ namespace cppwinrt
...
     inline std::string file_to_string(std::string const& filename)
     {
-        std::ifstream file(filename, std::ios::binary);
-        return static_cast<std::stringstream const&>(std::stringstream() << file.rdbuf()).str();
+        std::ifstream file(filename, std::ios::binary | std::ios::ate);
+        const auto size = file.tellg();
+        file.seekg(0);
+        std::string result(size, '\0');
+        file.read(result.data(), size);
+        return result;
```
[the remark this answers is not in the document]
lhecker: FWIW I benchmarked it and found that dropping down to ReadFile further doubles the speed.

jonwis (dismissed):
Fantastic! But do note @lhecker 's comment. I know we try and be crossplat-ready here, but 2x again is nothing to sneeze at... in aifabric builds, cppwinrt runtime is definitely a factor across all the projects that use it.

oldnewthing:
We can do what `cmd_reader.h` does and put the Windows-optimized version under `#if defined(_WIN32) || defined(_WIN64)`.

DefaultRyan:
@jonwis, @lhecker  
I'm open to measure the realistic impact of a Windows-specific code path. But we're already close to diminishing returns on this particular code. As it stands right now, this brings the time spent in `file_equal()` down to less than 2% of the total run time of cppwinrt.exe in the case where we completely skip the file write because the pending output perfectly matches the existing output. Probably on the order of 100 ms or less on average for generating a full SDK/platform projection?

There is probably more impactful possibilities for speeding up cppwinrt.exe.

For example, we are spending nearly 10% of our processing time on repeated (~1M) `std::map` lookups. An educated guess is that much of this is from repeatedly resolving a winmd's `TypeRef` namespace+name into a `TypeDef`, likely in `find_required()`.

YexuanXiao:
`std::fstream` is basically a wrapper around `FILE*` that provides iostream-style interfaces. Better performance can be achieved by directly using `std::fopen`/`fclose`/`fseek`/`fread`.

YexuanXiao:
C++23 adds `std::flat_map`, which is well-suited for scenarios where lookups far outnumber insertions, and it has been implemented by the STL [shipped with MSVC 14.51](https://github.com/microsoft/STL/wiki/Changelog#msvc-build-tools-1451). I expect it to solve the issue of `std::map` lookups taking too much time.

cppwinrt/text_writer.h
```diff
@@ -218,18 +236,15 @@ namespace cppwinrt
 
         bool file_equal(std::string const& filename) const
         {
-            if (!std::filesystem::exists(filename))
+            // Non-throwing file_size returns uintmax_t(-1) on errors, which shouldn't ever be the size of m_first or m_second
+            std::error_code ec;
+            if (std::filesystem::file_size(filename, ec) != m_first.size() + m_second.size())
```
DrusTheAxe, line 241: Should be

```
const auto size{ std::filesystem::file_size(filename, ec) };
if (ec || (size != m_first.size() + m_second.size())
```

?

https://en.cppreference.com/cpp/filesystem/file_size
(see code at bottom of Example)

DrusTheAxe (approved):
Neat! See comment for minor nit

cppwinrt/text_writer.h
sylveon: Not needed, the non throwing overload returns -1 on error, which will make the comparison fail.

manodasanW (approved):

PR #1586, opened 2026-05-27
Improve and ship module docs. Add missing properties to VS project properties page

### LINKED ISSUE #1583: Bug: Project properties page is not updated with cppwinrt 3
### Version

3.0.260520.1

### Summary

Image

The project property page for C++/WinRT should be updated to reflect the new module support `CppWinRTBuildModule` `CppWinRTModuleInclude` `CppWinRTModuleExclude` `CppWinRTConsumeModule`

### Reproducible example

```cpp

```

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional comments

The project settings seems to be [defined here](https://github.com/microsoft/cppwinrt/blob/master/nuget/CppWinrtRules.Project.xml)

DefaultRyan:
Fleshed out the module adoption guidance based on the excercise of prototyping Windows Terminal to use C++20 modules. Let's ship the readme.md and modules.md in the Nuget package.

Also, add the module-related project properties to the VS properties page

Fixes: #1583

### DIFF
--- a/nuget/CppWinrtRules.Project.xml
+++ b/nuget/CppWinrtRules.Project.xml
@@ -3,6 +3,7 @@
 
   <Rule.Categories>
     <Category Name="General" DisplayName="General"/>
+    <Category Name="Modules" DisplayName="C++20 Modules"/>
   </Rule.Categories>
 
   <Rule.DataSource>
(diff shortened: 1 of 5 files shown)


YexuanXiao:
Currently, C++/WinRT lacks a switch to define `WINRT_ENABLE_LEGACY_COM`, and this should be added.

sylveon:
I don't think this is particularly needed, or that the macro is needed at all. Legacy COM functionality is only visible to the user if they've dragged in windows.h (or unknwn.h) via other means.

manodasanW (approved):

leejy12:
> Also, add the module-related project properties to the VS properties page

Has this been shipped?

PR #1588, opened 2026-05-27
Precalculate pinterface guids for known generic instantiations

DefaultRyan:
C++/WinRT provides very powerful constexpr magic to allow users to calculate names and guids of parameterized types without requiring them to declare these instantiations in MIDL.

The catch is that the guid work is expensive because it involves, among other things, a SHA-1 hash computation. This happens at compile-time, so there is no runtime impact, but it does impact build times. The impact is particularly noticeable when many such pinterface specializations are needed, leading to build-time workarounds like https://github.com/microsoft/microsoft-ui-xaml/pull/209/changes

With this change, we resursively walk types in the namespace, looking for generic type instantiations in implmented/base interfaces, as well as method params and return types. We precalculate the guid of these generic types from inside cppwinrt.exe and emit them in the impl/<name>.0.h namespace headers.

This can't possibly capture every possible generic instantiation, but at least we can streamline the types we find in the winmd. I also stamp out `IReference` pinterface specializations for the "standard" types.

For example, `Windows.Foundation.Collections.IPropertySet` will now trigger explicit specializations of `winrt::impl::pinterface_guid` for the following WFC types:
- `IMap<hstring, IInspectable>` (interface impl of `IPropertySet`)
- `IObservableMap<hstring, IInspectable>` (interface impl of `IPropertySet`)
- `IIterable<IKeyValuePair<hstring, IInspectable>>` (interface impl of `IPropertySet`)
- `IKeyValuePair<hstring, IInspectable>` (part of the above `IIterable` instantiation, but also referenced in other methods)
- `IMapView<hstring, IInspectable>` (from `IMap::GetView()`)
- `IIterator<IKeyValuePair<hstring, IInspectable>>` (from `IIterable::First()`)
- `MapChangedEventHandler<hstring, IInspectable>` (from `IObservableMap::MapChanged`)
- `IMapChangedEventArgs<hstring>` (from `MapChangedEventHandler`)

These type combinations may be encountered in multiple different namespaces, and unlike vanilla inline variables and functions, C++ compilers don't like it when you provide multiple explicit template specializations of the same type, even when those specializations are identical, so I'm guarding them in preprocessor guards, like so:

```cpp
#ifndef WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__Collections__IKeyValuePair_hstring__hstring_
#define WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__Collections__IKeyValuePair_hstring__hstring_
    template <> struct pinterface_guid<winrt::Windows::Foundation::Collections::IKeyValuePair<hstring, hstring>>
    {
        static constexpr bool precomputed = true;
        static constexpr guid value{ 0x60310303,0x49C5,0x52E6,{ 0xAB,0xC6,0xA9,0xB3,0x6E,0xCC,0xC7,0x16 } }; // 60310303-49C5-52E6-ABC6-A9B36ECCC716
    };
#endif // WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__Collections__IKeyValuePair_hstring__hstring_
```

### Benchmark results
I created a benchmark file that gets `guid_of<T>()` on 60 or so generics that get precomputed pinterface guids (`IRerefence<T>` of built-in types, aforementioned pinterfaces from `IPropertyValue`, etc), and built it 10 times with these changes, as well as with the current master baseline. I also timed how long cppwinrt.exe took, to measure the runtime impact of precalculating these guids.

| **Metric** | **BEFORE (master)** | **AFTER (this PR)** | **Delta** |
| -- | -- | -- | -- |
| cppwinrt.exe warm avg | 0.796s | 0.882s | +0.086s (+11%) |
| cl.exe warm avg | 5.166s | 3.806s | -1.360s (-26%) |
| total warm avg | 5.962s | 4.688s | -1.274s (-21%) |

I did some profiling, and somewhat surprisingly, the actual guid calculation is barely a blip in the increased cppwinrt.exe. The additional time is dominated by the additional metadata walk, specifically, getting signature blobs to obtain generic type instantiations, and params/returns for method signatures, because those tend to return a temporary `std::vector` of stuff instead of the lighter-weight table traversals. There's some opportunity to optimize the winmd reader here, along with making type resolution (aka `find_required`) faster than the current `std::map` approach.


cppwinrt/helpers.h
```diff
@@ -1145,4 +1154,362 @@ namespace cppwinrt
...
+
+    // Add well-known IReference<T> and IReferenceArray<T> instantiations for standard primitive
+    // and Windows.Foundation struct types. These are emitted into Windows.Foundation.0.h so that
+    // consumers who only include Windows.Foundation.h (e.g. for boxing) still get precomputed GUIDs
+    // without requiring a header from a namespace that happens to use IReference<int32_t> etc.
+    //
+    // The pinterface signature format is: pinterface({open-type-guid};{arg-signature})
+    // These signatures are part of the WinRT ABI specification and are immutable.
+    static void add_well_known_ireference_instantiations(
+        std::map<std::string, generic_inst_info>& instantiations)
```
jonwis, line 1357: Aren't all these in the winmds as interface attributes already? midlrt should be precomputing even the "no specified `[uuid()]`" interfaces?

cppwinrt/winmd_signature.h
```diff
@@ -0,0 +1,383 @@
...
+        impl::write_hex(p, g.Data4[4]);
+        impl::write_hex(p, g.Data4[5]);
+        impl::write_hex(p, g.Data4[6]);
+        impl::write_hex(p, g.Data4[7]);
+        *p++ = '}';
+        return { buf, static_cast<std::size_t>(p - buf) };
+    }
+
+    // ---- SHA1 computation ----
+    struct sha1_context
```
jonwis, line 103: Is there a platform-standard crypto engine we should be using here instead of handrolling?

strings/base_reference_produce.h
```diff
@@ -393,6 +393,209 @@ WINRT_EXPORT namespace winrt::impl
...
+    template <> struct pinterface_guid<Windows::Foundation::IReference<char16_t>>
+    { static constexpr bool precomputed = true; static constexpr guid value{ 0xFB393EF3,0xBBAC,0x5BD5,{ 0x91,0x44,0x84,0xF2,0x35,0x76,0xF4,0x15 } }; };
+#endif
+#ifndef WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReference_bool_
+#define WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReference_bool_
+    template <> struct pinterface_guid<Windows::Foundation::IReference<bool>>
+    { static constexpr bool precomputed = true; static constexpr guid value{ 0x3C00FD60,0x2950,0x5939,{ 0xA2,0x1A,0x2D,0x12,0xC5,0xA0,0x1B,0x8A } }; };
+#endif
+#ifndef WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReference_winrt__hstring_
+#define WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReference_winrt__hstring_
```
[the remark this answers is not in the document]
DefaultRyan: Already fixed and refactored

strings/base_reference_produce.h
```diff
@@ -393,6 +393,209 @@ WINRT_EXPORT namespace winrt::impl
...
+    template <> struct pinterface_guid<Windows::Foundation::IReferenceArray<char16_t>>
+    { static constexpr bool precomputed = true; static constexpr guid value{ 0xA4095AAB,0xEB7D,0x5782,{ 0x8F,0xAD,0x16,0x09,0xDE,0xA2,0x49,0xAD } }; };
+#endif
+#ifndef WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReferenceArray_bool_
+#define WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReferenceArray_bool_
+    template <> struct pinterface_guid<Windows::Foundation::IReferenceArray<bool>>
+    { static constexpr bool precomputed = true; static constexpr guid value{ 0xE8E72666,0x48CC,0x593F,{ 0xBA,0x85,0x26,0x63,0x49,0x69,0x56,0xE3 } }; };
+#endif
+#ifndef WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReferenceArray_winrt__hstring_
+#define WINRT_IMPL_PINTERFACE_GUID_winrt__Windows__Foundation__IReferenceArray_winrt__hstring_
```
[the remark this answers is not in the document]
DefaultRyan: Already fixed and refactored

cppwinrt/helpers.h
DefaultRyan: You're talking about the 2 guids under iref_guid and iref_arr_guid? Yes, I suppose they are. Since I was trying to unconditionally insert these instantiations even though they aren't mentioned, I guess I got a little greedy and hardcoded these interfaces so I wouldn't have to synthesize a `GenericTypeInstSig` and look up the various bits. But, yes, it's probably better to use the actual attributes from `IReference` in the winmd, and synthesize the `GenericTypeInstSig` so it shared the rest of the computation code.

But if you're talking about the pinterface guids for the various specializations, those aren't in the winmd. Yes, midlrt precomputes the guids for certain specializations and emits them in the header files, but that data does not go into the winmd.

cppwinrt/winmd_signature.h
DefaultRyan: I'll check it out. We're already handrolling in the constexpr compile-time code, so we've already broken the seal, but it's probably better to not have two separate implementations.

I know we have [Windows.Security.Cryptography.Core.HashAlgorithmProvider](https://learn.microsoft.com/en-us/uwp/api/windows.security.cryptography.core.hashalgorithmprovider?view=winrt-28000). but I assume that won't work on non-Windows platforms, and I don't know if that API is available on all the Windows editions we need to run on. If this were app code, I'd be all over that, but I'm not too keen on converting stuff back and forth from a dynamically allocated `IBuffer` in this perf-sensitive code.

There is [BCryptHashData](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcrypthashdata), but has the same concerns around availability and cross-plat.

Outside of that, there aren't any truly *standard* implementations, but a fewseem to enjoy a degree of widespread use and seem to be somewhat trusted: OpenSSL, Crypto++, and Botan, with a heavy preference toward OpenSSL for the "official-ness". We aren't doing anything actually sensitive, just the SHA-1 hash (which will already raise some security warnings we'll need to deal with), but better safe than sorry if we're going to take on a third-party dependency.

Neither appear to have an *official* Nuget package (there are some that claim to have those packages, but with questionable provenance), so those are out. There are vcpkg bundles of each, though, so that could be an option, but cppwinrt has never had vcpkg support, so that's a whole other lift to add it to the build, CI, and official pipelines.

Sounds like something for "future us" to figure out...

DefaultRyan:
OMG, when I had the GitHub agent try to fix a build break, it completely nuked the PR title and description. Going to find the old one and change it back.

sylveon:
You can grab the previous revision from here:
image

cppwinrt/helpers.h
DefaultRyan: @jonwis The IReference logic is now "less special". I'm using the same GenericTypeIntSig path as the other specializations we discover in the winmds.

cppwinrt/winmd_signature.h
DefaultRyan: I think adding vcpkg hooks to the build is out of scope at the moment, but I want to take an IOU on that work, and use it to find an off-the-shelf SHA-1 implementation.

PR #1599, opened 2026-06-17
Fix race conditions in cancellable_promise

Clyraz:
The check in the revoke_canceller() method always succeeded on the second attempt, which led to segmentation faults if the cancellation callback ran for an extended period of time.


Clyraz:

ChrisGuzak (commented):
test cases?

Clyraz:
Closing in favour of #1609

Issue #1416, opened 2024-05-14
Bug: Calling PropertyValue.Create to handle boxing is significantly slower than allocating an IReference directly


sylveon:
### Version

2.0.240405.15

### Summary

This takes 30ms to execute:
```cpp
winrt::Windows::Foundation::IInspectable ref = nullptr;
for (int i = 0; i < 200000; i++)
    ref = winrt::box_value(i + 1);
```

This takes 8ms to execute:
```cpp
winrt::Windows::Foundation::IInspectable ref = nullptr;
for (int i = 0; i < 200000; i++)
    ref = winrt::make<winrt::impl::reference<int>>(i + 1);
```

It seems weird for `box_value` to be using PropertyValue.CreateInt32 when just directly allocating a wrapper is significantly faster.

### Reproducible example

```cpp
int main()
{
    winrt::init_apartment();
    const auto before = std::chrono::high_resolution_clock::now();

    winrt::Windows::Foundation::IInspectable ref = nullptr;
    for (int i = 0; i < 200000; i++)
        ref = winrt::box_value(i + 1);

    const auto after = std::chrono::high_resolution_clock::now();
    const auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(after - before);
    std::cout << duration.count();
}
```

### Expected behavior

I expected boxing to be fast.

### Actual behavior

It's kinda slow.

### Additional comments

Spotted this while looking into https://github.com/microsoft/microsoft-ui-xaml/discussions/9154

WPF users are expecting WinUI 3 to match WPF performance - and setting 200k DPs takes 21ms with WPF, but 130ms with WinUI 3. We ought to try and match that speed where possible, and this should help by cutting middlemans, at least in C++.

JaiganeshKumaran:
Reälly reminds me of those StructHelper types, which offered often basic functionality, doäble by yourself, with much greater overhead.

kennykerr:
As I recall, the `PropertyValue` implementations offer marshal-by-value semantics which the C++/WinRT fallback does not. This is/was required for some APIs depending on boxing. We can't change this now without the risk of breaking such callers. In practice, it's mostly only interesting to Xaml but there are other APIs that depend on this.

JaiganeshKumaran:
> marshal-by-value semantics

@kennykerr Can't we implement this ourselves?

sylveon:
From a cursory search, CsWinRT seems to allocate its own IReference (which works fine with XAML).

[...]

PR #1608, opened 2026-07-17
C++/WinRT ABI interop improvements

jonwis:
## Why

Analysis of the Rust (windows-rs) projection plus code analysis of Windows components that use C++/WinRT showed improvements available to C++/WinRT on two common interop paths — windows-rs boxes scalars and arrays in-process and builds `HSTRING` literals at compile time.

## What this PR changes

| Path | Change | Opt-in? |
| :-- | :-- | :-- |
| String | `_hs` compile-time `HSTRING` literal | opt-in |
| Reference | In-process scalar boxing (u8–u64, float, double, bool, char16, hstring, guid, DateTime, TimeSpan, Point) and array boxing (`IReferenceArray<T>`), marshaling by value through combase `PropertyValue` | transparent |

**Reference boxing** is the largest piece. `box_value` on a scalar or array previously hopped into combase to allocate a general `IPropertyValue` carrying the full discriminated-union machinery. This branch points the scalar and array `reference_traits` at the in-proc `impl::reference<T>` / `impl::reference_array<T>` that already back `IReference<T>` / `IReferenceArray<T>`, and makes those a correct `IPropertyValue` (right `PropertyType`, fixed `IsNumericScalar`, typed getters). A shared `reference_producer` CRTP base carries the whole `IPropertyValue` surface and the `IMarshal` tearoff once; each leaf supplies only its storage and how a getter reads it.

To preserve cross-process behavior, stock element types are marked `non_agile` and supply `IMarshal` lazily from `query_interface_tearoff` by building the equivalent combase `PropertyValue` — so the value still marshals **by value**, and the combase hop is paid only on marshal, never on box/unbox. Elements whose projected layout differs from the ABI (`hstring`, `DateTime`, `TimeSpan` arrays) and the non-stock element types (`IInspectable`, `Size`, `Rect`) continue to route through combase. This mirrors windows-rs' `StockReference`.

Two deltas vs combase, both matching windows-rs: `GetRuntimeClassName` now returns the `IReference<T>` / `IReferenceArray<T>` runtime class name rather than `Windows.Foundation.PropertyValue`, and cross-process the value marshals by value rather than as a proxy.

## Results

Warmed, order-controlled benchmark of Rust (windows-rs), stock cppwinrt, and `cppwinrt-new` (this branch). Rust is Rust→Rust; cppwinrt columns are C++→C++. Average of two reversed-order samples, 4M measured + 2M warmup iterations. Relative results are stable across multiple attempts.

| Path | Rust | cppwinrt (stock) | cppwinrt-new | new vs stock | new vs Rust |
| :-- | --: | --: | --: | :-- | :-- |
| Int32          |    14 |      18 |      10 | 1.8× faster | tie |
| String         |    14 |      20 |       8 | 2.5× faster | tie |
| Reference      |   435 |   2,065 |     541 | **3.8× faster** | Rust ~1.2× |

`Reference` retains a ~1.2× Rust lead that is structural, not the box shape: cppwinrt's per-object module-lock accounting (a `lock`-prefixed RMW on a process-global at each box's birth/death) plus the CRT `malloc`/`free` layer vs windows-rs' direct `HeapAlloc`.

## Testing

New tests: `test/test/reference_boxing.cpp` (scalar and array `PropertyType`, numeric conversion, round-trip, array-getter throw behavior, marshal-by-value via `GetUnmarshalClass` parity with combase, and cross-STA agile identity through the Global Interface Table) and `test/test_cpp20/hstring_literal.cpp`. Built Release/x64; `reference_boxing` passes all 61 assertions.

## References

- windows-rs fork with the benchmark harness and parity changes: [jonwis/windows-rs `jonwis/perf-bench`](https://github.com/jonwis/windows-rs/tree/jonwis/perf-bench/crates/samples/lang_perf)
- windows-rs compare: [master…jonwis:windows-rs:jonwis/perf-bench](https://github.com/microsoft/windows-rs/compare/master...jonwis:windows-rs:jonwis/perf-bench?expand=1)

(via Copilot)


oldnewthing:
> Iterator batching does have a slight change in semantics.

Another semantic change is that if the iterator is invalidated (e.g., because it is an iterator over a view that has been mutated), we don't throw `hresult_changed_state()` until the next time we go to fetch a batch. Old code threw at next iteration.

Therefore old code never operated on stale data. (Throws before stale data can be returned.) New code could operate on stale data for the remainder of the batch.

YexuanXiao:
Is it possible to add an opt-in `winrt::buffered_view<T, N>` that buffers elements instead of altering the current iterator? It can avoid various issues with iterators.

strings/base_iterator.h
```diff
@@ -126,8 +155,46 @@ WINRT_EXPORT namespace winrt::impl
...
+        using buffer_type = std::conditional_t<can_batch, std::array<value_type, block>, no_buffer>;
+
+        reference fetch(std::uint32_t const index) const
+        {
+            if constexpr (can_batch)
+            {
+                if (index < m_buffer_base || index >= m_buffer_base + m_buffer_size)
+                {
+                    m_buffer_base = index;
+                    m_buffer_size = m_collection->GetMany(index, m_buffer);
```
oldnewthing: Need to reorder these assignments. Otherwise, if GetMany throws E_BOUNDS, we will propagate the exception but we've already updated the m_buffer_base, so the next query for the same out-of-bounds element will succeed and return a cached element from some earlier batch.

strings/base_iterator.h
```diff
@@ -141,23 +208,115 @@ WINRT_EXPORT namespace winrt::impl
...
+            auto previous = *this;
+            ++*this;
+            return previous;
+        }
 
-        if (!result.HasCurrent())
+        bool operator==(buffered_iterator const& other) const noexcept
         {
-            return {};
+            return (m_size == 0) && (other.m_size == 0);
```
oldnewthing: This does lead to the weird scenario where a non-end iterator is not equal to itself! That might break some algorithms?

Maybe
```cpp
return m_size == other.m_size;
```
?
This gets us end==end as well as it==it.

Alternatively, have fill set m_iterator=null if the size is zero, and then this becomes m_iterator==other.m_iterator, which matches the behavior of the old iterator.

strings/base_iterator.h
```diff
@@ -141,23 +208,115 @@ WINRT_EXPORT namespace winrt::impl
...
+    private:
+
+        void fill()
+        {
+            m_index = 0;
+            m_size = m_iterator ? m_iterator.GetMany(m_buffer) : 0;
+        }
+
+        Iterator m_iterator{ nullptr };
+        std::array<value_type, buffer_capacity> m_buffer;
```
oldnewthing: I wonder what algorithms we will tank now that iterator assignment is not cheap.

strings/base_iterator.h
```diff
@@ -126,8 +155,46 @@ WINRT_EXPORT namespace winrt::impl
...
+            }
+
+            return m_collection->GetAt(index);
+        }
+
         T const* m_collection = nullptr;
         std::uint32_t m_index = 0;
+        mutable std::uint32_t m_buffer_base = 0;
+        mutable std::uint32_t m_buffer_size = 0;
+        mutable buffer_type m_buffer{};
```
oldnewthing: ``[[no_unique_address]]` so that `no_buffer` can disappear?

strings/base_coroutine_foundation.h
```diff
@@ -884,5 +964,16 @@ WINRT_EXPORT namespace winrt
...
+    template <typename TResult>
+    Windows::Foundation::IAsyncOperation<std::decay_t<TResult>> make_ready(TResult&& value)
+    {
+        return make<impl::ready_async_operation<std::decay_t<TResult>>>(std::forward<TResult>(value));
+    }
+
+    inline Windows::Foundation::IAsyncAction make_ready() noexcept
+    {
+        return make<impl::ready_async_action>();
+    }
```
oldnewthing: Do we also need WithProgress variants?

Also consider the versions here, which supports both with- and without progress , as well as `failed_async` for failed coroutines. https://devblogs.microsoft.com/oldnewthing/20240718-00/?p=109977

strings/base_reference_produce.h
```diff
@@ -2,7 +2,33 @@
...
+        std::is_same_v<T, Windows::Foundation::TimeSpan> || std::is_same_v<T, Windows::Foundation::Point>;
+
+    // Stock scalar references are marked non_agile so that our query_interface_tearoff supplies
+    // IMarshal (delegating to combase PropertyValue for by-value marshaling) instead of the default
+    // free-threaded-marshaler that marshals by reference. All other T keep the default agile shape.
+    template <typename T>
+    using reference_base_t = std::conditional_t<
+        is_stock_reference_v<T>,
+        implements<reference<T>, Windows::Foundation::IReference<T>, Windows::Foundation::IPropertyValue, non_agile>,
+        implements<reference<T>, Windows::Foundation::IReference<T>, Windows::Foundation::IPropertyValue>>;
```
oldnewthing: Or
```cpp
    using reference_base_t = implements<reference<T>, Windows::Foundation::IReference<T>, Windows::Foundation::IPropertyValue,
        std::conditional_t<is_stock_reference_v<T>, non_agile, std::monostate>>;
```

If a type parameter to `implements<>` is not a WinRT interface, not a COM interface, and not a marker class, then it is ignored.

sylveon:
> ```cpp
> // Previously:
> for (auto const& v : foo.ItemVector()) { // Call to v.First(), then j = i.Current()
>    /* ... */
>    // j is destroyed before i.MoveNext() & another i.Current()
> }
> ```

Slight nit, it actually does v.GetAt if it's available on ItemVector

sylveon:
Also, fixes #1416 :)

strings/base_coroutine_foundation.h
```diff
@@ -837,6 +837,86 @@ WINRT_IMPL_STD_EXPORT namespace std
     };
 }
 
+WINRT_EXPORT namespace winrt::impl
+{
+    template <typename Derived, typename AsyncInterface>
+    struct ready_async_base : implements<Derived, AsyncInterface, Windows::Foundation::IAsyncInfo>
```
sylveon: Couldn't this go in `wil`? There has been a prior desire to move the coroutine/etc stuff to wil and keep cppwinrt focused on purely projection, is this still true?

strings/base_iterator.h
```diff
@@ -126,8 +155,46 @@ WINRT_EXPORT namespace winrt::impl
...
 
+        // Batched forward traversal for random-access (GetAt-capable) collections that also expose
+        // GetMany. Instead of one GetAt ABI call per element, `fetch` fills a small block with a single
+        // GetMany call and serves in-window reads from it, so range-for (sequential ++ then *) crosses
+        // the ABI ~once per `block` elements. Random access still works: an out-of-window index
+        // re-anchors the block there, and an at/after-end index defers to GetAt so the component's
+        // bounds behavior (E_BOUNDS) is preserved. Elements are copied out -- no move-out -- so a given
+        // index may be read repeatedly, as the random-access contract requires. Collections without
+        // GetMany, or whose element type is not default-constructible, fall back to plain GetAt.
+        static constexpr bool can_batch = std::is_default_constructible_v<value_type> && has_GetMany_indexed<T>::value;
```
sylveon: Has there been any consideration on how this makes `fast_iterator` not cheap? They take far more space and require several AddRef to copy around. Would this slow down e.g. `<algorithm>` or `<ranges>`?

strings/base_iterator.h
```diff
@@ -141,23 +208,115 @@ WINRT_EXPORT namespace winrt::impl
...
+                return iterator_type{};
+            }
+
+            return result;
+        }
     }
 
     template <typename T, std::enable_if_t<!has_GetAt<T>::value, int> = 0>
-    auto get_end_iterator([[maybe_unused]] T const& collection) noexcept -> decltype(collection.First())
+    auto get_end_iterator([[maybe_unused]] T const& collection)
```
sylveon: Where did noexcept go?

strings/base_coroutine_foundation.h
jonwis: Sure. No particular reason to have it here.

strings/base_iterator.h
jonwis: True; if we go after `wil::buffered_iterator<>` ... maybe that should be an input_iterator so it's clear you're only supposed to move it forward.

jonwis:
> Is it possible to add an opt-in `winrt::buffered_view<T, N>` that buffers elements instead of altering the current iterator? It can avoid various issues with iterators.

I like this, @sylveon ... it does mean existing code won't be sped up, but that's probably OK.  And it also means it could move into WIL as you suggest. We can write code-review rules to request people use it. It also means you have to be aware of the "the collection changed under you" as @oldnewthing  suggests.

> > Iterator batching does have a slight change in semantics.
> 
> Another semantic change is that if the iterator is invalidated (e.g., because it is an iterator over a view that has been mutated), we don't throw `hresult_changed_state()` until the next time we go to fetch a batch. Old code threw at next iteration.
> 
> Therefore old code never operated on stale data. (Throws before stale data can be returned.) New code could operate on stale data for the remainder of the batch.

@oldnewthing  how do you feel about this compared to `wil::to_vector`? I think I'm leaning towards the "don't mess with the existing iterators" path above. Maybe the answer then is this disappears entirely, and the guidance is "if you are going to iterate the entire vector, just snarf it all with wil::to_vector<> and be done." ?

strings/base_iterator.h
sylveon: Yeah, I can imagine someone using a reverse iterator over fast_iterator wouldn't be too happy about this.

YexuanXiao:
A independent `winrt::buffered_view` can use stack memory, which saves memory compared to `wil::to_vector`. Its iterator can be a bidirectional iterator.

jonwis:
Pulled the buffered range-for iterator and the `make_ready` / already-completed async helpers out of this PR and reverted both headers back to baseline. They'll return shortly as a `wil/cppwinrt.h` implementation instead of living in cppwinrt core (the async side following Raymond's already-completed-async series). This PR now just carries the reference-boxing and `_hs` literal fast paths. Resolving the iterator and async review threads accordingly.

strings/base_reference_produce.h
jonwis: Done in 6495e489 — collapsed to a single `implements<>` with a conditional trailing marker. I used the inert `impl::marker` base instead of `std::monostate` purely to avoid pulling `<variant>` into the base headers; same effect since `implements<>` ignores non-interface, non-marker params. Also added a GIT-based test that boxes a scalar in one STA and fetches it from another, asserting identity is preserved (the IAgileObject in-proc fast path).

sylveon:
Think you forgot to push

strings/base_reference_produce.h
```diff
@@ -58,15 +102,15 @@ WINRT_EXPORT namespace winrt::impl
...
+        char16_t GetChar16() { if constexpr (std::is_same_v<T, char16_t>) { return m_value; } else { throw hresult_not_implemented(); } }
+        bool GetBoolean() { if constexpr (std::is_same_v<T, bool>) { return m_value; } else { throw hresult_not_implemented(); } }
+        hstring GetString() { if constexpr (std::is_same_v<T, hstring>) { return m_value; } else { throw hresult_not_implemented(); } }
+        guid GetGuid() { if constexpr (std::is_same_v<T, guid>) { return m_value; } else { throw hresult_not_implemented(); } }
+        Windows::Foundation::DateTime GetDateTime() { if constexpr (std::is_same_v<T, Windows::Foundation::DateTime>) { return m_value; } else { throw hresult_not_implemented(); } }
+        Windows::Foundation::TimeSpan GetTimeSpan() { if constexpr (std::is_same_v<T, Windows::Foundation::TimeSpan>) { return m_value; } else { throw hresult_not_implemented(); } }
+        Windows::Foundation::Point GetPoint() { if constexpr (std::is_same_v<T, Windows::Foundation::Point>) { return m_value; } else { throw hresult_not_implemented(); } }
         Windows::Foundation::Size GetSize() { throw hresult_not_implemented(); }
```
sylveon: Is it intentional that these types are not getting the fast path?

strings/base_string.h
```diff
@@ -155,6 +155,24 @@ WINRT_EXPORT namespace winrt::impl
...
+        static constexpr std::size_t size = N;
+        wchar_t value[N];
+
+        constexpr hstring_literal_storage(wchar_t const (&str)[N]) noexcept
+        {
+            for (std::size_t i = 0; i != N; ++i)
+            {
+                value[i] = str[i];
+            }
+        }
```
sylveon, line 171: static_assert for null termination?

build_test_all.cmd
```diff
@@ -4,38 +4,44 @@ set target_platform=%1
...
+call msbuild /m /p:Configuration=%target_configuration%,Platform=%target_platform%,CppWinRTBuildVersion=%target_version%%compiler_props% cppwinrt.sln /t:test\test_cpp20_no_sourcelocation
+call msbuild /m /p:Configuration=%target_configuration%,Platform=%target_platform%,CppWinRTBuildVersion=%target_version%%compiler_props% cppwinrt.sln /t:test\test_fast
+call msbuild /m /p:Configuration=%target_configuration%,Platform=%target_platform%,CppWinRTBuildVersion=%target_version%%compiler_props% cppwinrt.sln /t:test\test_slow
+call msbuild /m /p:Configuration=%target_configuration%,Platform=%target_platform%,CppWinRTBuildVersion=%target_version%%compiler_props% cppwinrt.sln /t:test\test_module_lock_custom
+call msbuild /m /p:Configuration=%target_configuration%,Platform=%target_platform%,CppWinRTBuildVersion=%target_version%%compiler_props% cppwinrt.sln /t:test\test_module_lock_none
+call msbuild /m /p:Configuration=%target_configuration%,Platform=%target_platform%,CppWinRTBuildVersion=%target_version%%compiler_props% cppwinrt.sln /t:test\test_module_lock_none
```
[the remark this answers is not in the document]
sylveon: Copilot appears right here. fix this as a drive-by?

strings/base_string.h
```diff
@@ -437,6 +479,44 @@ WINRT_EXPORT namespace winrt
...
 }
 
+#if defined(__cpp_nontype_template_args) && __cpp_nontype_template_args >= 201911L
+
+WINRT_EXPORT namespace winrt::impl
+{
+    // The fast-pass header for each distinct literal, materialized once as a compile-time
+    // constant so `_hs` is a constant expression rather than per-call work.
+    template <hstring_literal_storage Literal>
+    inline constexpr hstring_header hstring_literal_header
```
sylveon, line 487: Is it not possible to reuse param::hstring for this?

jonwis: Looks like no; we have to define the header somewhere - here it's in static const data so once ever per literal.  Using `winrt::param::hstring` instead param::hstring which is currently pointer-sized to hold the HSTRING. Hence the explicit conversion op.  If you've got a solution that removes this type I'm all ears. :)

strings/base_string.h
jonwis: Assert is now in the literal. There's no requirement that this type be null-terminated.

strings/base_string.h
sylveon: Hmmmm, there could maybe be a consteval constructor for param::hstring that just takes a pointer to hstring_header, which would make everyone benefit from this and not just explicit usages of the literal (and the literal could just do the same I think)

jonwis:
Filed the two pieces pulled out of this PR as WIL feature requests, per the discussion about keeping cppwinrt projection-only:

- Ready-made completed async (`make_ready`): microsoft/wil#663
- Buffered range-for over `IVector`/`IIterable` via `GetMany`: microsoft/wil#664

Both link back the prototype commits from this branch as a starting point.

strings/base_reference_produce.h
jonwis: Fixed!

DHowett:
> so the value still marshals **by value**, 
> ...
> Two deltas vs combase, ... cross-process the value marshals by value rather than as a proxy.


"still marshals" is contradictory with "rather than as a proxy" being considered a "delta" from combase.

Which is it?

sylveon:
Both. The IReference is still marshaled by value like it would be with PropertyValue::CreateXxx(). It is not marshaled by proxy like it would normally be with a custom IReference.

ChrisGuzak (approved):

PR #1615, opened 2026-08-03
Optimize `to_string`

YexuanXiao:
C++23 added `resize_and_overwrite` to avoid the problem of initializing the entire string to zero during `resize`. `winrt::to_hstring(std::string_view)` can take advantage of this feature.

### DIFF
--- a/strings/base_string.h
+++ b/strings/base_string.h
@@ -717,7 +717,7 @@ WINRT_EXPORT namespace winrt
     	result.resize_and_overwrite(size, [&](char* buffer, std::size_t) -> std::size_t
     		{
     			auto bytes_written = WINRT_IMPL_WideCharToMultiByte(65001 /*CP_UTF8*/, 0, value.data(), static_cast<std::int32_t>(value.size()), buffer, size, nullptr, nullptr);
-                WINRT_VERIFY_(size, bytes_written);
+    			WINRT_VERIFY_(size, bytes_written);
     			return bytes_written == size ? size : 0;
     		});
 #else
--- a/strings/base_string.h
+++ b/strings/base_string.h
@@ -720,8 +720,8 @@ WINRT_EXPORT namespace winrt
     			return size;
     		});
 #else
-    	std::string result(size, '?');
-    	WINRT_VERIFY_(size, WINRT_IMPL_WideCharToMultiByte(65001 /*CP_UTF8*/, 0, value.data(), static_cast<std::int32_t>(value.size()), result.data(), size, nullptr, nullptr));
+        std::string result(size, '?');
+        WINRT_VERIFY_(size, WINRT_IMPL_WideCharToMultiByte(65001 /*CP_UTF8*/, 0, value.data(), static_cast<std::int32_t>(value.size()), result.data(), size, nullptr, nullptr));
 #endif
         return result;
     }
--- a/strings/base_string.h
+++ b/strings/base_string.h
@@ -674,14 +674,7 @@ WINRT_EXPORT namespace winrt
     template <typename T, std::enable_if_t<std::is_same_v<T, bool>, int> = 0>
     hstring to_hstring(T const value)
     {
-        if (value)
-        {
-            return hstring{ L"true" };
-        }
-        else
-        {
-            return hstring{ L"false" };
-        }
+        return hstring{ value ? L"true" : L"false" };
     }
 
     inline hstring to_hstring(guid const& value)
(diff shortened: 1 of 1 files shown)


leejy12:
> Additionally, using value ? L"true" : L"false" instead of returning in two separate if branches allows the compiler to optimize better, avoiding the generation of two constructor calls, see https://godbolt.org/z/xeY5K4vMh.

I don't think this is a meaningful optimization. In your Godbolt example, `::to_hstring()` -- your optimized function -- includes a call to `wcslen` (to determine the integer argument to `winrt::impl::precreate_hstring_on_heap`).

When I ran the benchmarks, the optimized function was marginally slower.

<details>
<summary> Benchmark code </summary>

```cpp
#include <benchmark/benchmark.h>
#include <type_traits>
#include <winrt/base.h>

template <typename T, std::enable_if_t<std::is_same_v<T, bool>, int> = 0>
winrt::hstring to_hstring_optimized(T const value)
{
    return winrt::hstring{value ? L"true" : L"false"};
}

template <typename T, std::enable_if_t<std::is_same_v<T, bool>, int> = 0>
winrt::hstring to_hstring_original(T const value)
{
    if (value)
    {
        return winrt::hstring{L"true"};
    }
    else
    {
        return winrt::hstring{L"false"};
    }
}

namespace
{

void benchmark_to_hstring_optimized(benchmark::State &state)
{
    const bool value = state.range(0) != 0;

    for (auto _ : state)
    {
        auto result = ::to_hstring_optimized(value);
        benchmark::DoNotOptimize(result);
    }
}

void benchmark_to_hstring_original(benchmark::State &state)
{
    const bool value = state.range(0) != 0;

    for (auto _ : state)
    {
        auto result = ::to_hstring_original(value);
        benchmark::DoNotOptimize(result);
    }
}

BENCHMARK(benchmark_to_hstring_original)->ArgName("value")->Arg(false)->Arg(true);
BENCHMARK(benchmark_to_hstring_optimized)->ArgName("value")->Arg(false)->Arg(true);

} // namespace

BENCHMARK_MAIN();
```
</details>

<details>
<summary>Benchmark results</summary>

```
D:\test\bm-1615> .\build\Release\bm_1615.exe       
2026-08-07T01:08:52-05:00
Running D:\test\bm-1615\build\Release\bm_1615.exe
Run on (12 X 2976 MHz CPU s)
CPU Caches:
  L1 Instruction 192 KiB (x12)
  L1 Data 96 KiB (x12)
  L2 Unified 12288 KiB (x3)
---------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations
---------------------------------------------------------------------------------
benchmark_to_hstring_original/value:0        26.5 ns         25.1 ns     28000000
benchmark_to_hstring_original/value:1        26.2 ns         23.4 ns     28000000
benchmark_to_hstring_optimized/value:0       29.0 ns         28.3 ns     24888889
benchmark_to_hstring_optimized/value:1       28.9 ns         28.5 ns     28000000
```

</details>

YexuanXiao:
> I don't think this is a meaningful optimization. In your Godbolt example, `::to_hstring()` -- your optimized function -- includes a call to `wcslen` (to determine the integer argument to `winrt::impl::precreate_hstring_on_heap`).

You're right, I will revert the change. Thank you.

strings/base_string.h
```diff
@@ -719,8 +719,18 @@ WINRT_EXPORT namespace winrt
             return{};
         }
 
+#if defined(__cpp_lib_string_resize_and_overwrite) && __cpp_lib_string_resize_and_overwrite >= 202110L
+    	std::string result;
+    	result.resize_and_overwrite(size, [&](char* buffer, std::size_t) -> std::size_t
+    		{
+    			auto bytes_written = WINRT_IMPL_WideCharToMultiByte(65001 /*CP_UTF8*/, 0, value.data(), static_cast<std::int32_t>(value.size()), buffer, size, nullptr, nullptr);
+    			WINRT_VERIFY_(size, bytes_written);
+    			return bytes_written == size ? size : 0;
```
DefaultRyan: Nit: This slightly changes the specified behavior from the original. Now we truncate to zero if written differs from size, whereas before we would simply accept the result.

In practice, assuming WideCharToMultiByte behaves sanely, written should always equal the value of size (obtained from the first call to WCTMB). I think we are committing to assuming this sane behavior, so let's keep the defensive guardrails consistent, and go ahead and return bytes_written directly. The WINRT_VERIFY_ sanity check is fine to keep as is to warn if something weird happened.

strings/base_string.h
```diff
@@ -719,8 +719,18 @@ WINRT_EXPORT namespace winrt
             return{};
         }
 
+#if defined(__cpp_lib_string_resize_and_overwrite) && __cpp_lib_string_resize_and_overwrite >= 202110L
+    	std::string result;
+    	result.resize_and_overwrite(size, [&](char* buffer, std::size_t) -> std::size_t
```
DefaultRyan: Nit: Feels slightly dirty to capture `size` instead of using the lambda's param. Perhaps this is simplified if we are selective about our lambda captures and capture only `&value`.

PR #1619, opened 2026-08-26
Silence experimental coroutine deprecation

ChrisGuzak:
## Summary

Define MSVC's opt-out macro for the deprecated <experimental/coroutine> header used by the repository's C++17 /await:strict configuration.

This keeps existing C++17 coroutine builds working with current MSVC toolsets without changing runtime behavior.


sylveon:
/await:strict doesn't use experimental coroutines. This seems like the wrong fix?

YexuanXiao:
/await works with \<experimental/coroutine\>, and /await:strict works with \<coroutine\>. C++20 mode does not require either.

DefaultRyan:
Need a more thorough repro case for what this purports to fix. C++/WinRT no longer includes the deprecated experimental coroutine headers.

Issue #1620, opened 2026-09-01
Bug: headers cannot be compiled at high warning levels (MSVC /Wall, clang)


DanielJump:
### Version

master (856eaff). Measurements below were taken against the generated projection in `windowssdkcppwinrt.10.21320.2`, with the `strings/` fixes validated at master.

### Summary

C++/WinRT headers cannot be compiled at high warning levels. Consumers are forced to wrap every include in `#pragma warning` blocks. We repackage cppwinrt for the Office monorepo and currently inject a 19-warning push/disable prologue into every generated header as a post-generate step, which is unsatisfying and hides real diagnostics in adjacent code.

I went through that whole suppression list to work out which entries are genuine defects in cppwinrt rather than noise. Most of them are, and all of them are fixable. This issue is to track that; the pull requests are linked at the bottom.

**MSVC, measured on the repro below at `/Wall`:**

| Code | x64 | x86 | Cause |
|---|---|---|---|
| C5246 | 114 | 114 | `std::array` aggregate init needs two brace levels. The `guid` constructor alone accounts for 57 of these, one per generated guid |
| C4365 | 21 | 31 | integral promotion, plus `uint32_t` to `ptrdiff_t` iterator arithmetic that only shows up on 32-bit |
| C4946 | 16 | 16 | `reinterpret_cast` between related classes. Every abi type derives from `unknown_abi`, so `static_cast` is correct |
| C4265 | 15 | 15 | virtual functions with an accessible non-virtual destructor |
| C4826 | 0 | 1 | `reinterpret_cast` of a pointer to `int64_t` sign-extends |

Six further codes we suppress (C4101, C4189, C4456, C4457, C4459, C5205) never fired in any configuration I tried, and C4458 is already suppressed internally by `base_delegate.h`. C6101 fires under `/analyze` but all instances are in the Windows SDK's `WindowsNumerics.inl`, not in cppwinrt.

**clang (clang-cl 22.1.2), counting only the warnings we suppress:**

| Warning | Count | Notes |
|---|---|---|
| `-Wnon-virtual-dtor` | 184 | see #1499 |
| `-Wunused-variable` | 60 | already fixed at master by `[[maybe_unused]]`; only affects older packages |
| `-Wnontrivial-memcall` | 19 | already fixed at master by making the `base_array.h` guard clang-aware |
| `-Wmissing-braces` | 12 | all in the SDK's `DirectXMathMatrix.inl`, not cppwinrt |
[...]

Issue #1585, opened 2026-05-24
Using import winrt_base generates Microsoft Code Analysis warnings in the consumer project for base.h


vbaderks:
### Version

3.0.260520.1

### Summary

I switch yesterday my open source project from using my own C++ module wrapper of winrt to the v3 implementation. In my release build configuration I have Microsoft Code Analysis enabled with a custom ruleset.
To be able to use v3 I had to disable a couple of analysis rules to prevent warnings reported on base.h as my project uses the option warnings are errors.

It is not 100% clear to me why MSVC reports these warnings from base.h in my consuming project as for my project base.h is external code. I tried several ways to silence these warnings but at the end was forced to disable these warnings in my custom ruleset file.

I used to create a module from winRT v2 using my own winrt.ixx file:

```
module;
#include <Unknwn.h> // Required to enable classic COM support in WinRT.
#pragma warning(push)
#pragma warning(disable : 4265)  // : '' : class has virtual functions, but its non-trivial destructor is not virtual
#pragma warning(disable : 4946)  // reinterpret_cast used between related classes
#pragma warning(disable : 5204)  // class has virtual functions, but its trivial destructor is not virtual
#pragma warning(disable : 5260)  // the constant variable has external\internal linkage
#pragma warning(disable : 26432) // If you define or delete any default operation in the type define or delete them.
#include <winrt/base.h>
#pragma warning(pop)

export module winrt;

export namespace winrt {
using winrt::check_hresult;
using winrt::check_win32;
// ... etc.
}
```

The CA warnings that I needed to disable were:
```
- C26436: The type X with a virtual function needs either public virtual or protected non-virtual destructor (c.35).
- C26432: f you define or delete any default operation in the type define or delete them all (c.21).
- C26466: Don't use static_cast downcasts. A cast from a polymorphic type should use dynamic_cast (type.2).
- C26493: Don't use C-style casts (type.4).
- C26473: Don't cast between pointer types where the source type and the target type are the same (type.1).
- C26496: The variable 'hr' does not change after construction, mark it as const (con.4).
- C26497: You can attempt to make '' constexpr 
- C26409: Avoid calling new and delete explicitly, use std::make_unique<T> instead (r.11).
- C26447: The function is declared 'noexcept' but calls function 'AddRef()' which may throw exceptions (f.6).
- C26465: Don't use const_cast to cast away const or volatile. const_cast is not required; constness or volatility is not being 
```

### Reproducible example

```cpp
[...]

PR #1623, opened 2026-09-01
Suppress non-virtual destructor warnings in the generated headers

DanielJump:
Part of #1620.

`-Wnon-virtual-dtor` (clang) and C4265/C5204 (MSVC) fire on the ABI interface structs, the
delegate structs, and the `produce<D, I>` vtable adapters. That is about 200 warnings in a
single translation unit that exercises `implements<>`, delegates, events, and the collection
helpers.

They are false positives. The only refcounted, heap-allocated type here is `root_implements`,
and it already has a virtual destructor. `producer`, `produce<D, I>`, and the ABI structs are
non-owning subobjects -- each one's address is simply the COM interface pointer handed out for
that interface -- and none of them is ever deleted. The warning is reporting a delete that
cannot occur.

This emits a scoped push/disable/pop around the body of each generated header, matching the
guarded style already used in `base_delegate.h`, `base_implements.h`, `base_activation.h`, and
`base_error.h`. It is popped at the end of every header, so consumer code outside these headers
keeps the warning.

Measured against the shipped headers with the same pragmas applied:

| | before | after |
|---|---|---|
| clang `-Wnon-virtual-dtor` | 200 | 0 |
| MSVC `/Wall` C4265 (x64) | 15 | 0 |
| MSVC `/Wall` C4265 (x86) | 15 | 0 |
| errors | 0 | 0 |

The earlier revision of this PR gave those types protected destructors instead. That was
replaced per review feedback: suppression leaves the types untouched, so there is no question
of affecting an ABI type or forbidding a derivation that compiles today.

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -81,8 +81,45 @@ namespace cppwinrt
         w.write(format);
     }
 
+    // The ABI interface structs, the delegate structs, and the produce<D, I> vtable adapters
+    // all have virtual functions and non-virtual destructors, which trips C4265/C5204 on MSVC
+    // and -Wnon-virtual-dtor on clang. None of those types is ever deleted. The only
+    // refcounted, heap-allocated type is root_implements, and it already has a virtual
+    // destructor; everything below it is a non-owning subobject whose address is simply the
+    // COM interface pointer. The warning is therefore reporting a delete that cannot occur.
+    // Scoped with push/pop so that consumer code outside these headers is unaffected.
+    static void write_diagnostic_push(writer& w)
+    {
+        auto format = R"(#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable: 4265) // class has virtual functions, but destructor is not virtual
+#pragma warning(disable: 5204) // class has virtual functions, but its trivial destructor is not virtual
+#endif // _MSC_VER
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnon-virtual-dtor"
+#endif // __clang__
+)";
+
+        w.write(format);
+    }
+
+    static void write_diagnostic_pop(writer& w)
+    {
+        auto format = R"(#ifdef __clang__
+#pragma clang diagnostic pop
+#endif // __clang__
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif // _MSC_VER
+)";
+
+        w.write(format);
+    }
+
     static void write_close_file_guard(writer& w)
     {
+        write_diagnostic_pop(w);
         write_endif(w);
     }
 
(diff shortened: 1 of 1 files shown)


sylveon:
I think it would make more sense to suppress here than mess with the vtable layout of an ABI type. I'm also not sure final is correct, doesn't cppwinrt derive from those when implementing an interface?

DanielJump:
Good calls -- I have rewritten this to do what you suggested.

## final

You are right. To be precise about the scope, `final` was only on `produce<D, I>`, never on the
ABI structs; what cppwinrt derives from when implementing an interface is `produce_base<D, I>`
(about 7,700 sites in the Windows projection), and that had a protected destructor rather than
`final`. Nothing in this repo or in the generated headers derives from `produce<D, I>` itself.

But that is not a sufficient argument, because third-party code is not bound by what this repo
does. `struct Whatever : produce<D, IThing> {}` compiles today and would have become:

```
error C3246: 'Whatever': cannot inherit from 'produce<MyD,IThing>' as it has been declared as 'final'
```

Such a class is fragile regardless, since `shim()` reinterpret_casts `this` back to
`producer<D, I>*` and assumes it is that type's by-value vtable member. But "it compiles today"
is reason enough not to break it in a header with this much reach. Dropped.

## Vtable layout

One note for the record, since I think the diff may have read as adding virtual destructors: the
destructors were non-virtual, so they added no vtable slot. MSVC `/d1reportSingleClassLayout` on
a 3-method COM-style interface, before and after adding `protected: ~I() = default;`:

```
class IBefore  size(8):              class IAfter  size(8):
 0 | {vfptr}                          0 | {vfptr}
 0 | &IBefore::QueryInterface          0 | &IAfter::QueryInterface
 1 | &IBefore::AddRef                  1 | &IAfter::AddRef
 2 | &IBefore::Release                 2 | &IAfter::Release
```

What does change the layout is `virtual ~type()`, which is what #1499 proposes: 4 slots instead
of 3, and in a derived interface the destructor slot displaces the derived methods. Avoiding that
was the reason for using access rather than virtuality. Moot now, but worth recording.

## The rewrite

This PR is now a scoped push/disable/pop around the body of each generated header, in the same
guarded style already used in `base_delegate.h`, `base_implements.h`, `base_activation.h`, and
`base_error.h`. One file, +38 lines, no type touched. clang `-Wnon-virtual-dtor` 200 -> 0, MSVC
`/Wall` C4265 15 -> 0 on both x64 and x86, no errors.

I kept the reasoning in the code comment rather than only muting the warning, because the reason
seems worth recording: the only refcounted, heap-allocated type is `root_implements` and it
already has a virtual destructor, while `producer`, `produce`, and the ABI structs are non-owning
subobjects that are never deleted. So this is a false positive rather than a papered-over hazard,
which also gives #1499 and #1585 an answer.

Unrelated, but spotted while looking for precedent: the four disables in `base_macros.h` (5046,
4268, 4499, 4630) have no push/pop, so they leak into consumer translation units for the rest of
the file. Happy to send a separate PR wrapping those if that is welcome.

#1621 and #1622 are independent of this one, and are genuine conversion and brace-elision bugs
rather than false positives.

PR #1625, opened 2026-09-01
Remove obsolete warning disables from base_macros.h

### LINKED ISSUE #1624: base_macros.h disables warnings without push/pop, leaking them into consumer translation units
### Description

`strings/base_macros.h` disables four MSVC warnings with bare `#pragma warning(disable: ...)` and no matching `push` / `pop`:

```cpp
#ifdef _MSC_VER
// Note: this is a workaround for a false-positive warning produced by the Visual C++ 15.9 compiler.
#pragma warning(disable : 5046)

// Note: this is a workaround for a false-positive warning produced by the Visual C++ 16.3 compiler.
#pragma warning(disable : 4268)

// C++ module warnings by /W4
#pragma warning(disable : 4499)
#pragma warning(disable : 4630)
#endif // _MSC_VER
```

`base.h` includes `base_macros.h` near the top and never restores the warning state. C5046, C4268, C4499 and C4630 therefore stay disabled for the **remainder of every translation unit** that includes `<winrt/base.h>`, which includes all of the consumer's own code that follows the include.

The disables are legitimate - they cover declarations that C++/WinRT itself makes. The bug is only that their lifetime is unbounded.

### Repro

```cpp
// leak.cpp
namespace { struct S { int x; }; }
S f();
int main() { f(); return 0; }
```

```
cl /c /std:c++20 /permissive- /W4 leak.cpp
leak.cpp(3): warning C5046: 'f': Symbol involving type with internal linkage not defined
```

Add the include and the warning disappears, even though the offending code is unchanged and is entirely outside C++/WinRT:

```cpp
#include <winrt/base.h>
namespace { struct S { int x; }; }
S f();
int main() { f(); return 0; }
```

```
cl /c /std:c++20 /permissive- /W4 /I<sdk> leak.cpp
(no diagnostics)
```

Reproduced with both `/W4` and `/Wall` on MSVC x64.

### This is the only unscoped suppression in the tree

Auditing every warning pragma under `strings/`:

| header | MSVC push / pop / disable | clang push / pop / ignored |
| --- | --- | --- |
| base_activation.h | 1 / 1 / 1 | 1 / 1 / 1 |
| base_composable.h | 1 / 1 / 1 | - |
| base_delegate.h | 1 / 1 / 1 | - |
| base_error.h | - | 1 / 1 / 1 |
| base_fast_forward.h | - | 1 / 1 / 1 |
| base_implements.h | 1 / 1 / 1 | 1 / 1 / 1 |
| **base_macros.h** | **0 / 0 / 4** | - |

Every other suppression in the project is already correctly scoped, so this looks like an oversight rather than a deliberate choice.

### Suggested fix

Open a `#pragma warning(push)` before the `base_macros.h` include and pop it at the end of `base.h`. The disables keep covering everything C++/WinRT declares, and the warning state is handed back to the consumer unchanged.

I verified that this leaves the set of warnings reported from inside the C++/WinRT headers completely unchanged (a `/Wall` build of a translation unit exercising collections, coroutines, delegates, `implements`, and classic COM reports an identical 166 warnings before and after), while restoring C5046 in the consumer code above.

Fixed in #1625.

### Scope note

`base_macros.h` is also root-included by the generated component `.g.h` files (under `WINRT_IMPORT_MODULE`) and by the global module fragment of each generated `.ixx`. Those paths are cppwinrt's own module builds rather than ordinary consumer translation units, and C4499 / C4630 are specifically wanted there, so I left them alone. Happy to extend the fix if you would prefer it applied uniformly.

### Relationship to #1623

#1623 adds a file-level diagnostic scope to every generated header, which would incidentally contain these disables inside `base.h` as a side effect. I am raising and fixing this separately because it is a distinct bug with its own repro, and because it should be fixed whether or not #1623 is taken.

DanielJump:
Fixes #1624.

`base_macros.h` globally disables C5046, C4268, C4499 and C4630, so every translation unit that includes `<winrt/base.h>` also loses those diagnostics in its own code.

The first two pragmas are documented in the source as workarounds for Visual C++ 15.9 and 16.3. The latter two were added for early C++ module support. Current supported toolsets no longer emit any of the four warnings for these sources, so this removes the pragmas instead of adding push/pop plumbing around them.

The resulting change is only 12 deleted lines.

### Validation

- The official MSVC off-by-default warning list does not include C5046, C4268, C4499 or C4630.
- A rich generated-header translation unit compiles under `/Wall` with both v143 and v145 without any of the four warnings.
- The v145 named-module target rebuilds under `/Wall` without any of the four warnings.
- Normal v143 and v145 `/W4 /WX` builds and tests remain clean.
- The original consumer repro reports C5046 again after including `<winrt/base.h>`.

MSVC warning-level reference: https://learn.microsoft.com/cpp/build/reference/compiler-option-warning-level

### DIFF
--- a/cppwinrt/code_writers.h
+++ b/cppwinrt/code_writers.h
@@ -117,6 +117,33 @@ namespace cppwinrt
         return { w, write_close_file_guard };
     }
 
+    static void write_macro_warning_pop(writer& w)
+    {
+        auto format = R"(#ifdef _MSC_VER
+#pragma warning(pop)
+#endif // _MSC_VER
+)";
+
+        w.write(format);
+    }
+
+    // base_macros.h disables a handful of compiler warnings that C++/WinRT's own
+    // declarations would otherwise trigger. Those disables have to stay in effect
+    // for the rest of the file that includes it, so the enclosing scope is opened
+    // here rather than inside base_macros.h itself. Without the matching pop the
+    // disables would remain active for everything the consumer writes after the
+    // include, silently turning the warnings off for their code as well.
+    [[nodiscard]] static finish_with wrap_macro_warning_scope(writer& w)
+    {
+        auto format = R"(#ifdef _MSC_VER
+#pragma warning(push)
+#endif // _MSC_VER
+)";
+
+        w.write(format);
+        return { w, write_macro_warning_pop };
+    }
+
     [[nodiscard]] static finish_with wrap_lean_and_mean(writer& w, bool is_lean_and_mean = true)
     {
         if (is_lean_and_mean)
--- a/cppwinrt/file_writers.h
+++ b/cppwinrt/file_writers.h
@@ -17,6 +17,7 @@ namespace cppwinrt
                 w.write(strings::base_detect_numerics);
                 w.write(strings::base_include_numerics);
             }
+            auto wrap_macro_warnings = wrap_macro_warning_scope(w);
             w.write_root_include("base_macros");
             w.write(strings::base_source_location);
             w.write(strings::base_types);
--- a/strings/base_macros.h
+++ b/strings/base_macros.h
@@ -23,6 +23,11 @@
 #define WINRT_IMPL_SHIM(...) (*(abi_t<__VA_ARGS__>**)&static_cast<__VA_ARGS__ const&>(static_cast<D const&>(*this)))
 
 #ifdef _MSC_VER
+// These disables deliberately apply to the remainder of the file that includes
+// this header, because they cover declarations made throughout C++/WinRT. The
+// including file opens a #pragma warning(push) beforehand and pops it at the
+// end, which keeps them from escaping into consumer code.
+
 // Note: this is a workaround for a false-positive warning produced by the Visual C++ 15.9 compiler.
 #pragma warning(disable : 5046)
 


sylveon:
If these apply only to base.h, then why not move them to base.h?

YexuanXiao:
I doubt whether C++/WinRT still needs to keep these pragmas, as support for VS2017 and VS2019 may have long since ceased to exist.

DanielJump:
@sylveon @YexuanXiao You are both right to question preserving these pragmas. I tested removing them instead.

The v143 regular header build no longer emits C5046 or C4268, and the v145 named-module build no longer emits C4499 or C4630. Both build with `/W4 /WX`, and their tests pass. I replaced the push/pop implementation with deletion of all four obsolete disables, so the PR is now only 12 deleted lines and fixes the leak at its source.

DanielJump:
Correction to my previous validation note: `/W4 /WX` alone would not have answered whether these were off-by-default warnings. I had not explicitly checked that before pushing the rewrite.

I have now checked the official MSVC off-by-default list and rebuilt under `/Wall`. None of C5046, C4268, C4499 or C4630 is listed as off by default. A rich generated-header TU is clean for all four on v143 and v145 under `/Wall`, and the v145 named-module target is also clean for all four under `/Wall`. The deletion still holds, but the stronger validation should have happened first.

YexuanXiao:
I suggest changing it to:

```cpp
#if defined(_MSC_VER) && defined(WINRT_IMPL_BUILD_MODULE)
// C++ module warnings by /W4
#pragma warning(disable : 4499)
#pragma warning(disable : 4630)
#endif // _MSC_VER
```

because all module files define WINRT_IMPL_BUILD_MODULE before including base_macro.h and header mode does not need them. C++ modules do not leak macros, so they do not need to be pushed/popped. As for the other two warnings, if they no longer occur in VS2022, then perhaps it's better to simply remove them, since C++/WinRT is no longer tested against VS2017 and VS2019.

YexuanXiao:
I tested base.h and base_macro.h using https://godbolt.org/z/jajMTveT5, and VS17.0 no longer produces the erroneous 4499 and 4630 warnings with /W4 and /WX enabled.

DanielJump:
Thanks for checking VS 17.0 as well. That matches the v143 and v145 results here: neither C4499 nor C4630 is emitted for the module path anymore. Since the supported header and module toolsets are clean for all four warnings, I will leave the PR as the simpler deletion rather than retain module-only suppressions.

PR #1626, opened 2026-09-11
Fix the issue where #include "pch.h" is always being generated

YexuanXiao:
This MSBuild property is problematic. Based on my testing, it is always true, so it always generates `#include "pch.h"`. The solution comes from https://github.com/microsoft/microsoft-ui-xaml/blob/914147a04d5439e88d551f23869cfaeae6eb474b/src/XamlCompiler/Targets/Microsoft.UI.Xaml.Markup.Compiler.interop.targets#L352 .

@DefaultRyan @Scottj1s


sylveon:
This sounds like it's only a problem if you disable precompiled headers but don't delete the pch.h file

YexuanXiao:
> This sounds like it's only a problem if you disable precompiled headers but don't delete the pch.h file

This is confusing. Since WinUI has provided a better approach, cppwinrt should adopt it.

YexuanXiao:
Anyone who wants to disable the pch will run into this issue: after disabling the pch, they investigate which files are still including it, and then discover that cppwinrt still requires it. This makes them confusing, and I'm one of those people. I believe others have encountered the same confusion, which is why WinUI implemented this change.

YexuanXiao:
> This sounds like it's only a problem if you disable precompiled headers but don't delete the pch.h file

I retested it, and after deleting pch.h, it still includes pch.h. In my opinion, this is clearly a bug.

Issue #1602, opened 2026-06-27
Bug: Why CppWinRTModuleInclude not auto include the dependencies of include modules


wherewhere:
### Version

3.0.260520.1

### Summary

When we use `CppWinRTModuleInclude`, we expect that we can write the module which we need so that we do not need to generate everything. But when we add it, we will find that it is useless because it indeed only generates the module we write but not consider the dependencies which the module referenced. So that we need to write all dependencies manually. But the dependencies of dependencies of dependencies will be almost all namespace to include. We are almost unable to find out all the dependencies.

### Reproducible example

Add

```xml
<CppWinRTModuleInclude>
  Windows.Foundation
</CppWinRTModuleInclude>
```

Then try to fix the dependencies. Do not say use `CppWinRTModuleExclude`, we need use all the member under `Windows.Foundation` and do not want to know what they dependence.

### Expected behavior

C++/WinRT should add all used module.

### Actual behavior

C++/WinRT only add `Windows.Foundation`.

### Additional comments

I don't think it is a feature. Or the C++/WinRT should include all `Windows.*` modules just like C#/WinRT so that when we add third part `CppWinRTModuleInclude` it can correctly reference the `Windows.*` modules.

YexuanXiao:
I requested a feature on [DevCom](https://developercommunity.visualstudio.com/t/C20-modules:-CompileAsCppModule-items/11114019) to support on-demand compilation of modules. If the C++ target of MSBuild improves in this regard, then C++/WinRT will not need to further support it. If you are interested in this issue, make sure to vote for it.

PR #1627, opened 2026-09-12
Fix exclude-only C++/WinRT module filtering

hoshiizumiya:
When `CppWinRTModuleExclude` is specified without `CppWinRTModuleInclude`, C++/WinRT currently generates no namespace `.ixx` files at all, even though the documented contract says that an unspecified `CppWinRTModuleInclude` means “all namespaces” and `CppWinRTModuleExclude` should only remove matching namespace prefixes from that candidate set. For example, enabling `CppWinRTBuildModule` and setting only `<CppWinRTModuleExclude>Microsoft.UI</CppWinRTModuleExclude>` should generate modules for all eligible namespaces except `Microsoft.UI*`, but instead the module set becomes empty. The cause is that `build_filters` constructs `module_filter` whenever either include or exclude is non-empty, and the underlying `winmd::reader::filter` interprets an empty include set as “include nothing” rather than “no include restriction”; once `module_filter` is non-empty, every namespace must pass `module_filter.includes(...)`, so an exclude-only configuration rejects all namespaces. This behavior conflicts with the documented `(all)` default for `CppWinRTModuleInclude`; the module-specific filter should therefore treat an empty include set as unrestricted and apply only the exclude filter.


hoshiizumiya:
Remind that this won't fix #1602 
This is related to #1602 in that both involve module include/exclude filtering, but it is a separate issue. #1602 concerns whether an explicit `CppWinRTModuleInclude` should automatically include the transitive dependencies of the selected namespaces. This issue occurs before dependency resolution is relevant: when `CppWinRTModuleInclude` is not specified at all, its documented default is `(all)`, so an exclude-only configuration should mean “all eligible namespaces minus the excluded prefixes.” Instead, the empty include set is currently treated as an empty candidate set once `CppWinRTModuleExclude` causes a module filter to be constructed, resulting in no namespace modules being generated.

PR #1628, opened 2026-09-16
Make natvis find winmds from the WinMD Reference

### LINKED ISSUE #2809: Natvis for C++/WinRT doesn't work
### Describe the bug

Natvis for C++/WinRT doesn't work because the *.winmd files aren't deployed alonside the process *.exe. This affects both SDK and user authored types.

### Steps to reproduce the bug

1. Create a solution from `Blank App, Packaged (WinUI 3 in Desktop)` template.
2. Set a breakpoint in `App::OnLaunched()`
3. Debug and pause at breakpoint.
4. Try to inspect the `LaunchActivatedEventArgs` argument.

### Expected behavior

The debug visualizer shows object properties based on metadata.

### Screenshots

Screenshot 2022-08-02 121502

Notice the output message. To enable natvis diagnostics see [documentation](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/natvis).

### NuGet package version

1.1.3

### Packaging type

Packaged (MSIX)

### Windows version

Windows 11 version 21H2 (22000)

### IDE

Visual Studio 2022

### Additional context

A workaround is to manually copy the *.winmd files from `Microsoft.WindowsAppSDK` to the executable folder.

YexuanXiao:
Fix https://github.com/microsoft/WindowsAppSDK/issues/2809.

Currently, the VSIX extension only looks for winmd files from the exe directory and the system directory, making the visualization useless. In particular, metadata from the WindowsAppSDK was also not taken into consideration.

This patch passes the information of all winmd references to the binary through MSBuild, which is then parsed by the VSIX plugin to obtain the paths of the referenced winmd files. All of this is automatic.

All winmds are loaded lazily. When looking up a type, the extension first collects the paths of all winmd files, and then looks for the appropriate file based on the namespace of the type. Once a namespace has been looked up, it is recorded to avoid repeating a futile load the next time.


YexuanXiao:
Moved to draft because I found that some environments may not display the visualization of types in WAS.

